PR #1437, opened 2021-02-04
buildelement: switch order of timed_activity and sandbox.batch

abderrahim:
We want the actual sandbox operations to be timed.


cs-shadow (approved):
Good spot, thanks!

I've rebased your branch on latest master, and will hit merge once the CI passes here.

PR #1434, opened 2021-02-04
Redesign/redocument configuration surface for interaction with remotes

gtristan:
This implements everything as described on [the mailing list](https://lists.apache.org/thread.html/r3eb4ffaad44d8308ad26b3b5466723a11b6183cdb564dbf7f7497106%40%3Cdev.buildstream.apache.org%3E)

PR #1438, opened 2021-02-05
Tristan/remote cli options

gtristan:
This merge request implements `--remote` options consistently in the CLI, so that remotes can be fully specified for every command which supports remotes.


gtristan:
This merge request depends on https://github.com/apache/buildstream/pull/1434 being merged first.

cs-shadow:
Hi @gtristan, I haven't had a chance to look at the rest but here are my thoughts from the UI/UX perspective:

tl;dr I think it's preferrable to use singular option names in both cases - i.e. whether multiple remotes are allowed or just one.

---

When only one value is expected, we always use the singular option name. So, that's sorted. For the remainder of this comment, I'll assume that we're dealing with an option that needs to accept multiple values.

Click offers two ways of handling multiple values for options:

1. [multiple=True](https://click.palletsprojects.com/en/7.x/options/?highlight=multiple#multiple-options), which looks like `--option value1 --option value2 ...`

2. [nargs](https://click.palletsprojects.com/en/7.x/options/?highlight=multiple#multi-value-options), which looks like `--option value1 value2 ...`

In my opinion singular option names (like  `--remote`) make more sense for (1) and plural option names (like `--remotes`) for (2). This is because `--remotes R1 R2 R3` sounds natural while `--remotes R1 --remotes R2` sounds rather awkward. Similarly, `--remote R1 --remote R2` sounds better than `--remote R1 R2 R3`

Currently, we are using `nargs` exclusively for postional arguments and `multiple` excusively for options. As such, my suggestion would be to use singular names whenever dealing with options, regardless of how many values they accepts.

In this pull request, seems like you're already doing so, like `--artifact-remote` and `--source-remote`.

Lastly, `--ignore-project-artifact-remotes` also seems good to me. This is because it is a flag and cannot be repeated. And the plural name correctly indicates that it will operate on all remotes and not a specific one.

---

PS: Using `multiple=True` for options is somewhat forced on us because options are not allowed to have nargs < 0.

gtristan:
> tl;dr I think it's preferrable to use singular option names in both cases - i.e. whether multiple remotes are allowed or just one.
> [...]
> In this pull request, seems like you're already doing so, like `--artifact-remote` and `--source-remote`.

Agreed, and indeed the patch is implemented already as you suggest.

e.g.:

```python
@click.option(
    "--remote",
    "remotes",
    type=RemoteSpecType(RemoteSpecPurpose.PULL),
    multiple=True,
    help="A remote for downloading sources",
)
```
Here the CLI option is `--remote` while the variable name passed to the function is a list named `remotes`.

When observing `--help` output, the value type `REMOTE` is derived from the newly added `RemoteSpecType` (click ParamType derivative), it will show something like:

> --remote   REMOTE   Some command specific bla bla to feed to the blabbernaught

Regarding documentation, I opted to add a single general section to the `Using -> Commands` section explaining how to fully specify remotes, and explaining that all `--remote` like options can be specified multiple times (this section is then linked to from the user configuration section which discusses cache servers, which is where the main user facing cache server configuration documentation resides).

src/buildstream/_frontend/app.py
```diff
@@ -711,7 +711,7 @@ def _handle_failure(self, element, task, failure):
                             _Scope.BUILD,
                             self.shell_prompt,
                             isolate=True,
-                            usebuildtree="always",
+                            usebuildtree=True,
```
sstriker, line 714: How does this change relate?

gtristan: This is just a side effect of changing the various `Stream()` APIs (using the new type annotations for the changed APIs) and reviewing all of the call sites for these changed `Stream()` APIs.

There is no such thing as `"always"` as an answer to the question *"Shall I use a build tree for this shell"*, this does not technically fix any bug because `"always"` is a truthy value, but the signature takes a `bool`, so we now give it a `bool`.

PR #1439, opened 2021-02-09
buildelement: switch order of timed_activity and sandbox.batch

cs-shadow:
We want the actual sandbox operations to be timed.

---

Replaces #1437

All credits to @ abderrahim, just resubmitting the patch this from a branch in this repository so that the CI actually runs.

### DIFF
--- a/src/buildstream/buildelement.py
+++ b/src/buildstream/buildelement.py
@@ -288,7 +288,7 @@ class BuildElement(Element):
         root_list = self.__layout.get("/", None)
         if root_list:
             element_list = [element for element, _ in root_list]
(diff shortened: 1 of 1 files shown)


cs-shadow:
Interesting, apparently Action runs are tied to commit ids (and not pull requests). And it looks like Actions runs trigerred on the same commit in my branch is also updating the action statuses on #1437.

As such, I'll close this replacement and merge the original PR. Sorry for the noise :)

---

If nothing else, I think I've found a workaround to trigger action runs when triggered from a fork - we could write an action that responds to new pull requests by pushing them to a temporary branch in the main repository. We should probably not need to have to do that though.

PR #1448, opened 2021-02-25
Make public data immutable

gtristan:
This is the first step in the [sealing the sandbox](https://lists.apache.org/thread.html/r3eb2dce3561cb46ac80c859b14aafb2471aaf3e319a39f70475fe22a%40%3Cdev.buildstream.apache.org%3E) initiative.


gtristan:
Before merging this, we need to prepare updates to the adjacent plugin repos and ensure they are ported to work with this change.

nanonyme:
So will this prevent having manifest generators that are the basis of CVE checking system used by freedesktop-sdk and gnome-build-meta? Ping @jjardon

gtristan:
> So will this prevent having manifest generators that are the basis of CVE checking system used by freedesktop-sdk and gnome-build-meta? Ping @jjardon

The manifest generators will have to be reimplemented, I did discuss this at length with @jjardon some time ago and he agreed that we need not push for this reimplementation to block BuildStream 2.

Frankly, these plugins have been broken and doing illegal things from the start, e.g., they are making assumptions and calling into plugin specific APIs based on what plugin they *think* they have discovered on a dependency based on calling `.get_kind()` (which exists for logging and error reporting purposes only), as reported a very long time ago: https://gitlab.com/BuildStream/bst-plugins-experimental/-/issues/2

Further, these plugins are the first to break sandboxing guarantees by having host python writing directly into the sandbox, setting a trend that is proving very difficult to reverse, as [outlined here](https://lists.apache.org/thread.html/re3055975198fa5115b7ce3f533e58518443976b6ae2b13c3be15c881%40%3Cdev.buildstream.apache.org%3E).

The recommended way to implement manifest generation is:
* Introduce `Source` level APIs which allow plugins to report information in a standardized way
  * The `url` and `ref` fields are entirely *private* details of `Source` plugins, which only might or might not have such YAML configurations, some `Source` plugins may have multiple URLs, and `local` source plugins do not have any URLs. The manifest generation assumes there is such a concept of a "url", as such, the `Source` API needs to be extended in such a way that a plugin may *report* what the `url`, or list of `urls` is, same thing goes for the `refs`.
* This new `Source` level API will allow for the possibility of implementing a command like `bst source show`, which will allow the frontend to interrogate sources and print them in a custom formattable, machine readable way to stdout, similar to `bst show` (such a command has been requested various times, nobody has gotten around to proposing an implementation yet).
* The manifest generation can then be trivially implemented with a shell script (or python or other script) that invokes BuildStream and extracts and formats the output as desired.

abderrahim:
How about instead of removing `set_public_data` from the public API, let's keep it with the new semantics (of only being allowed in configure phase). Looking at bst-plugins-experimental, all uses of `set_public_data()` don't need to be in `assemble()` and can be moved to either `configure()` or `configure_dependencies()`. The only exception (that generates public data based on the contents of the sandbox) is `dpkg_build`.

gtristan:
This has been abandoned.

PR #1449, opened 2021-02-25
_context.py: Ensure paths in user configuration are absolute

abderrahim:
Relative paths don't make sense in user configuration. The exception is
workspacedir where `.` is useful as it will be combined with the name
specified on the command line.

Backport of f2c3f8eb0db188caac6991632c4453c39b42e84c (part of [!1566](https://gitlab.com/BuildStream/buildstream/-/merge_requests/1566))

### DIFF
--- a/buildstream/_context.py
+++ b/buildstream/_context.py
@@ -181,6 +181,12 @@ class Context():
             path = os.path.normpath(path)
             setattr(self, directory, path)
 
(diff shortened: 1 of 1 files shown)


gtristan:
Sorry for the delay, we've been investigating why the CI is not running on pull requests originating from forks and still need to sort that out.

PR #1450, opened 2021-03-02
Make artifact and source cache query explicit

juergbi:
Artifact and source cache query can be fairly expensive as it checks the presence of all artifact and source blobs, respectively. This will be especially true with the upcoming optional support for remote caching (reducing blob downloads with remote execution to a minimum) in PR #1431.

This branch makes cache queries more explicit in the code and puts the `Stream` class in control in preparation for parallelization of cache queries.

This is a simplified replacement for PR #1430, not yet moving the cache queries to scheduled jobs, thus avoiding a performance hit in certain use cases.

### DIFF
--- a/src/buildstream/_artifact.py
+++ b/src/buildstream/_artifact.py
@@ -540,7 +540,7 @@ class Artifact:
 
         return dependency_refs
 
-    # cached():
+    # query_cache():
     #
     # Check whether the artifact corresponding to the stored cache key is
     # available. This also checks whether all required parts of the artifact
@@ -550,11 +550,7 @@ class Artifact:
     # Returns:
     #     (bool): Whether artifact is in local cache
     #
-    def cached(self):
-
-        if self._cached is not None:
-            return self._cached
-
+    def query_cache(self):
         context = self._context
 
         artifact = self._load_proto()
@@ -587,6 +583,18 @@ class Artifact:
         self._cached = True
         return True
 
+    # cached()
+    #
+    # Return whether the artifact is available in the local cache. This must
+    # be called after `query_cache()` or `set_cached()`.
+    #
+    # Returns:
+    #     (bool): Whether artifact is in local cache
+    #
+    def cached(self):
+        assert self._cached is not None
+        return self._cached
+
     # cached_logs()
     #
     # Check if the artifact is cached with log files.
@@ -599,15 +607,6 @@ class Artifact:
         # Log files are currently considered an essential part of an artifact.
         # If the artifact is cached, its log files are available as well.
         return self._element._cached()
-
-    # reset_cached()
-    #
-    # Allow the Artifact to query the filesystem to determine whether it
-    # is cached or not.
-    #
-    def reset_cached(self):
-        self._proto = None
-        self._cached = None
 
     # set_cached()
     #
--- a/src/buildstream/_artifact.py
+++ b/src/buildstream/_artifact.py
@@ -591,9 +591,12 @@ class Artifact:
     # Returns:
     #     (bool): Whether artifact is in local cache
     #
-    def cached(self):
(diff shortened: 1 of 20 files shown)


src/buildstream/_artifact.py
```diff
@@ -550,11 +550,7 @@ def get_dependency_artifact_names(self):
     # Returns:
     #     (bool): Whether artifact is in local cache
     #
-    def cached(self):
-
-        if self._cached is not None:
-            return self._cached
-
+    def query_cache(self):
```
gtristan, line 553: This is rather orthogonal to the patch, but I think we need to cleanup how this function behaves differently depending on global values decided in the `Context`.

I don't feel like it is safe to have such a deep function call behaving differently *whole sale* based on user configuration, and think rather we should be passing these values in as arguments when checking what *"domains"* of an artifact are locally cached (i.e. is only the ref cached, are the files available locally, etc).

* `Context.require_artifact_files`

  This is only ever set to `False` by `Stream` in the case that remote execution is enabled

* `Context.require_artifact_directories`

  This is unconditionally `True` all the time, maybe this historically meant something, but we forgot to remove this boolean once things changed such that it is always true (at least grep seems to tell me nothing ever sets this value to `False`)

* `Element._artifact_files_required()`

  Even here I feel like it may happen that an `Element` which does evaluate to `False` for this function (i.e. it *"requires artifact files"*) could justifiably decide to ask the `Artifact` whether the files are cached or not

We can resolve this separately from the branch, but as I am reading it this jumps out at me and strikes me as problematic so just raising it here.

src/buildstream/_artifactelement.py
```diff
@@ -131,9 +131,9 @@ def clear_artifact_name_cache(cls):
     # Once we've finished pulling an artifact, we assume the
     # state of the pulled artifact.
     #
-    def _pull_done(self):
-        super()._pull_done()
+    def _load_artifact_done(self):
```
gtristan, line 132: I think we need to streamline how `ArtifactElement` instances are initialized with this.

This branch changes a bit the semantics of how an artifact is loaded onto an element, but does not appear to update `Element.__initialize_from_artifact()`, perhaps we can have `Element.__initialize_from_artifact()` directly call `Element._load_artifact()` or such.

Another point, is that loading the artifact from the local cache is obviously a hard requirement when loading any `ArtifactElement`, regardless of whether `Stream.query_cache()` was called on the element.

I think this branch does not exactly change the current initialization (and deferring the artifact loading is planned in a later branch), in any case, it should be clear somehow (perhaps in the `Stream.query_cache()` API comments) that this will end up being a `noop` for `ArtifactElements` which are anyway synchronously loaded at instantiation time.

src/buildstream/_artifact.py
juergbi: PR #1431 completely drops the two variables and the method you mention above, see commit 87e93c1fc6a45c83823a541719f24e1aa8f92f56.

gtristan: Sounds great !

src/buildstream/_artifactelement.py
juergbi: Using the `_load_artifact()` code path for artifact element early loading seems sensible. I've pushed a separate commit for this.

In `Stream.query_cache()` I've added an explicit no-op case if the artifact has already been loaded.

gtristan: Thanks :)

My concern was mostly to ensure that the state of an artifact loaded early on, or an artifact loaded as a result of pulling it, would be as close as possible.

gtristan: Looking at your patch, it seems the code has changed more than I expected it might have... anyway, looks good to me :)

PR #1454, opened 2021-03-08
Remove requirement of running private global fixtures

gtristan:
Some of the special global fixtures in the private
`buildstream.testing._fixtures` module are required to run even
in tests declared in external plugin repositories, this is especially
true ever since the scheduler was made to be threaded, and the loaded
netrc data is no longer local to the process which uses it.

Since then, we've added 0360bc1feca1d5429cdb7fbc083727d242499733
which explicitly resets more state in the DownloadableFileSource, tests
in external plugin repos which derive this source will fail without having
this state reset between BuildStream invocations.

This patch changes things such that the internal node state and internal
DownloadableFileSource state is reset unconditionally between BuildStream
invocations, so that it is not necessary for external repos to concern
themselves with cherry picking these functions in any conftest.py they
might use.

The test `tests/sourcecache/fetch.py::test_fetch_fallback` needed to be fixed
in this case, as it is using under the hood nonsense to test, which relies on
state being persisted where this cannot and should not be guaranteed.

### DIFF
--- a/src/buildstream/_stream.py
+++ b/src/buildstream/_stream.py
@@ -51,7 +51,8 @@ from ._remotespec import RemoteSpec
 from ._state import State
 from .types import _KeyStrength, _PipelineSelection, _Scope, _HostMount
 from .plugin import Plugin
-from . import utils, _yaml, _site, _pipeline
+from . import utils, node, _yaml, _site, _pipeline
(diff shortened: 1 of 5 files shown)


BenjaminSchubert (approved):
Thanks!

PR #1457, opened 2021-03-17
element.py: Compute cache key also for rdep without buildable callback

### LINKED ISSUE #1452: BUG when building in non-strict mode
Trying to build freedesktop-sdk in non-strict mode, I got the following message.

```
[--:--:--] BUG     [????????] freedesktop-sdk.bst:components/sqlite.bst: Build

      File "/home/abderrahim/.virtualenvs/bst-master/lib/python3.9/site-packages/buildstream/_scheduler/jobs/job.py", line 442, in child_action
      File "/home/abderrahim/.virtualenvs/bst-master/lib/python3.9/site-packages/buildstream/_scheduler/jobs/elementjob.py", line 93, in child_process
      File "/home/abderrahim/.virtualenvs/bst-master/lib/python3.9/site-packages/buildstream/_scheduler/queues/buildqueue.py", line 58, in _assemble_element
        return element._assemble()
      File "/home/abderrahim/.virtualenvs/bst-master/lib/python3.9/site-packages/buildstream/element.py", line 1776, in _assemble
        return self._cache_artifact(sandbox, collect)
      File "/home/abderrahim/.virtualenvs/bst-master/lib/python3.9/site-packages/buildstream/element.py", line 1821, in _cache_artifact
        assert self.__cache_key is not None
    AssertionError
```

It isn't reproducible in that rerunning the pipeline succeeds (or actually fails in a subsequent element; this is the second time it happened).

This is using bc81cc6360f4777493d3bdcc3fa382bd5f0f358a (branch abderrahim/bst2)

juergbi:
The reverse dependency may not be in the build queue yet and thus, the buildable callback may be `None`. However, it's still necessary to update the cache key in non-strict mode.

Fixes #1452.

Fixes: 8d9bf024 ("element.py: Stop computing cache keys when...")

### DIFF
--- a/src/buildstream/element.py
+++ b/src/buildstream/element.py
@@ -2412,10 +2412,12 @@ class Element(Plugin):
                     rdep.__build_deps_uncached -= 1
                     assert not rdep.__build_deps_uncached < 0
 
-                    if rdep.__buildable_callback is not None and rdep._buildable():
+                    if rdep._buildable():
                         rdep.__update_cache_key_non_strict()
-                        rdep.__buildable_callback(rdep)
-                        rdep.__buildable_callback = None
+
+                        if rdep.__buildable_callback is not None:
+                            rdep.__buildable_callback(rdep)
+                            rdep.__buildable_callback = None
 
     # _walk_artifact_files()
     #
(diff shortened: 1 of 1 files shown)


juergbi:
As the issue is essentially a race condition, writing a reliable regression test for the test suite may no be straight forward, unfortunately. I was able to reproduce the issue manually with a clean freedesktop-sdk build, though, and I've verified the fix manually the same way.

BenjaminSchubert (approved):

PR #1462, opened 2021-03-19
tar.py: fix permissions in tarballs

### LINKED ISSUE #1459: buildstream stages files from tar files with wrong permissions
Trying to build qemu ([tarball](https://download.qemu.org/qemu-5.2.0.tar.xz)), I noticed that buildstream stages some files as executable even though they shouldn't be. For instance
```python
>>> import tarfile
>>> tar = tarfile.open('qemu-5.2.0.tar.xz')
>>> tar.getmember('qemu-5.2.0/pc-bios/hppa-firmware.img').get_info()
{'name': 'qemu-5.2.0/pc-bios/hppa-firmware.img', 'mode': 436, 'uid': 1000, 'gid': 1000, 'size': 785696, 'mtime': 1607446784, 'chksum': 7844, 'type': b'0', 'linkname': '', 'uname': 'fluxion', 'gname': 'fluxion', 'devmajor': 0, 'devminor': 0}
```
where 436 is 664 in octal. However in the sandbox
```bash
$ ls -l pc-bios/hppa-firmware.img 
-rwxr-xr-x 1 abderrahim abderrahim 785696 Nov 11  2011 pc-bios/hppa-firmware.img
```
that is 755.

I see the tar plugin does some [funny stuff with the permissions](https://github.com/apache/buildstream/blob/master/src/buildstream/plugins/sources/tar.py#L67), the error certainly lies in there.

abderrahim:
the code was confusing & and | and thus making everything executable.

Fixes #1459

### DIFF
--- a/src/buildstream/plugins/sources/tar.py
+++ b/src/buildstream/plugins/sources/tar.py
@@ -77,7 +77,7 @@ class ReadableTarInfo(tarfile.TarInfo):
         # Respect umask instead of the file mode stored in the archive.
         # The only bit used from the embedded mode is the executable bit for files.
         umask = utils.get_umask()
-        if self.isdir() or bool(self.__permission | 0o100):
+        if self.isdir() or bool(self.__permission & 0o100):
             return 0o777 & ~umask
         else:
             return 0o666 & ~umask


gtristan:
Good catch!

gtristan (approved):

PR #1463, opened 2021-03-20
plugin.py: Avoid import errors on certain platforms

gtristan:
In some cases (on Debian with python 3.9 it has been seen) when multiple
threads enter the Plugin.blocking_activity() context manager simultaneously,
we get ImportErrors from the multiprocessing submodules complaining that
we are importing symbols from partially initialized submodules (hinting at
possible circular imports).

Ensuring that these submodules have been initialized up front circumvents
these edge case stack trace bugs from occurring.

### DIFF
--- a/src/buildstream/plugin.py
+++ b/src/buildstream/plugin.py
@@ -110,7 +110,21 @@ Class Reference
 """
 
 import itertools
+
+#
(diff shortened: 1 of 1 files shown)


BenjaminSchubert (approved):
That is probably going to need some pylint exclusion :)

PR #1464, opened 2021-03-21
Relicensing under the Apache License 2.0

gtristan:
This branch changes the license from the LGPL 2.1 to the Apache License 2.0.

This was done following the rules outlined in: https://infra.apache.org/apply-license.html


gtristan:
NOTE: This should not be merged until all contributors relevant to the master branch have signed the ICLA, or the corporate CLA in the case that developers were employed to make contributions on their behalf.

nanonyme:
Will this same be done for bst-1 branch assuming it still needs releases?

nanonyme:
I would suggest as *first* step on this road update https://github.com/apache/buildstream/blob/master/CONTRIBUTING.rst. It currently says *nothing* about ICLA's.

gtristan:
> I would suggest as _first_ step on this road update https://github.com/apache/buildstream/blob/master/CONTRIBUTING.rst. It currently says _nothing_ about ICLA's.

I've raised this before and my understanding from @sstriker is that we don't need ICLAs to accept contributions from fly by contributors via pull requests on github. At least this should be true once the ASF license is in place (after merging this PR).

We *do* require the ICLAs in order to relicense code which was contributed under the LGPL.

nanonyme:
@gtristan great. Can you please handle the bst-plugins-experimental release then at some point? That said, that doesn't prevent merging this change. It prevents tagging after this change.

gtristan:
> @gtristan great. Can you please handle the bst-plugins-experimental release then at some point? That said, that doesn't prevent merging this change. It prevents tagging after this change.

Note that `bst-plugins-experimental` is not slated for a relicensing, so afaics is completely unrelated to this topic.

nanonyme:
You are now removing API (Python bytecode mangling) that latest bst-plugins-experimental release depends on. How can it possibly not be related?

gtristan:
> You are now removing API (Python bytecode mangling) that latest bst-plugins-experimental release depends on. How can it possibly not be related?

Ah I see what you mean, but I thought there was another issue we recently closed specifically about that :)

This PR is specifically about updating the source code headers and COPYING file and such license related details.

nanonyme:
Sorry, I mixed up threads and didn't realize context switch. Indeed this PR is not directly related although lack of that tag is blocking tags with Apache license.

PR #1467, opened 2021-03-21
_artifact.py: don't consider an artifact cached if metadata is missing

### LINKED ISSUE #1466: Error pushing artifact (blob not found)
I'm seeing this error during push
```
[--:--:--] START   [964e4af3] core-deps/dnsmasq.bst: Push
[--:--:--] STATUS  [964e4af3] core-deps/dnsmasq.bst: Pushing data from artifact 964e4af3 -> http://127.0.0.1:11001/
[00:00:00] FAILURE [964e4af3] core-deps/dnsmasq.bst: Failed to push artifact blobs: Failed to upload blob cd832935dcdd40d5a9a2d929b9c9dc3ab5a755d7c9f16b747242df0e7b4c6432: 5
```

I believe the 5 above means [NOT FOUND](https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto#L78), and indeed the blob in question doesn't exist in buildstream's cache. I don't know which blob is this, but I guess it's something that bst doesn't check and just assumes it's always there if the artifact is there.

abderrahim:
Fixes #1466

### DIFF
--- a/src/buildstream/_artifact.py
+++ b/src/buildstream/_artifact.py
@@ -574,7 +574,7 @@ class Artifact:
 
         # Check whether public data and logs are available
         logfile_digests = [logfile.digest for logfile in artifact.logs]
-        digests = [artifact.public_data] + logfile_digests
+        digests = [artifact.low_diversity_meta, artifact.high_diversity_meta, artifact.public_data] + logfile_digests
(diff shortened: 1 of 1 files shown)


juergbi (approved):
This looks good to me.

Issue #1483, opened 2021-06-21
Failed to build with Python 3.10


tim77:
buildstream fails to build with Python 3.10.0a5.

```python
Traceback (most recent call last):
  File "/builddir/build/BUILD/buildstream-1.6.1/doc/bst2html.py", line 38, in <module>
    from buildstream import _yaml
  File "/builddir/build/BUILD/buildstream-1.6.1/buildstream/__init__.py", line 33, in <module>
    from .source import Source, SourceError, Consistency, SourceFetcher
  File "/builddir/build/BUILD/buildstream-1.6.1/buildstream/source.py", line 150, in <module>
    from collections import Mapping
ImportError: cannot import name 'Mapping' from 'collections' (/usr/lib64/python3.10/collections/__init__.py)
Traceback (most recent call last):
  File "/builddir/build/BUILD/buildstream-1.6.1/doc/bst2html.py", line 38, in <module>
    from buildstream import _yaml
  File "/builddir/build/BUILD/buildstream-1.6.1/buildstream/__init__.py", line 33, in <module>
    from .source import Source, SourceError, Consistency, SourceFetcher
  File "/builddir/build/BUILD/buildstream-1.6.1/buildstream/source.py", line 150, in <module>
    from collections import Mapping
ImportError: cannot import name 'Mapping' from 'collections' (/usr/lib64/python3.10/collections/__init__.py)
Traceback (most recent call last):
  File "/builddir/build/BUILD/buildstream-1.6.1/doc/bst2html.py", line 38, in <module>
    from buildstream import _yaml
  File "/builddir/build/BUILD/buildstream-1.6.1/buildstream/__init__.py", line 33, in <module>
    from .source import Source, SourceError, Consistency, SourceFetcher
  File "/builddir/build/BUILD/buildstream-1.6.1/buildstream/source.py", line 150, in <module>
    from collections import Mapping
ImportError: cannot import name 'Mapping' from 'collections' (/usr/lib64/python3.10/collections/__init__.py)
Traceback (most recent call last):
  File "/builddir/build/BUILD/buildstream-1.6.1/doc/bst2html.py", line 38, in <module>
    from buildstream import _yaml
  File "/builddir/build/BUILD/buildstream-1.6.1/buildstream/__init__.py", line 33, in <module>
    from .source import Source, SourceError, Consistency, SourceFetcher
  File "/builddir/build/BUILD/buildstream-1.6.1/buildstream/source.py", line 150, in <module>
    from collections import Mapping
[...]

PR #1487, opened 2021-07-08
WIP: Use collections.abc instead collections

jjardon:
collections is deprecated since version 3.3, will be removed in version 3.10

See https://docs.python.org/3/library/collections.html#module-collections

Part of #1483

### DIFF
--- a/contrib/update_committers.py
+++ b/contrib/update_committers.py
@@ -9,7 +9,7 @@ import json
 import urllib.request
 import urllib.parse
 from jinja2 import Environment, FileSystemLoader
-from collections import OrderedDict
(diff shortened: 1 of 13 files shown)


nanonyme:
This looks reasonable. The only Python 3 versions this would break are long EOL. (pre-3.3)

nanonyme:
Tests are failing. Reviewed and it looks like this is wrong. Only these were moved https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes

jjardon:
Yeah, seems all this has been fixed already in master. Closing

PR #1489, opened 2021-08-02
requirements/requirements.in: It depends on grpc >= 1.30

jjardon:
with older version you will get errors like this accessing the CAS:
```
[--:--:--][][] WARNING Failed to initialize remote https://freedesktop-sdk-cache.codethink.co.uk:11001: Received http2 header with status: 404
```

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,4 +1,16 @@
 =================
+buildstream 1.6.2
+=================
+
+  o Fix some issues with a previous fix for #532
+
+  o Ported to github CI infrastructure
+
+  o Ensure paths specified in user configuration are absolute
+
+  o Import some symbols from collections.abc, required for python 3.10
+
+=================
 buildstream 1.6.1
 =================
 
--- a/requirements/requirements.in
+++ b/requirements/requirements.in
@@ -1,5 +1,5 @@
 Click
-grpcio >= 1.10
+grpcio >= 1.30
 jinja2 >= 2.10
 pluginbase
 protobuf >= 3.5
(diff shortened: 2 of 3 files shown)


juergbi:
Is grpcio 1.37 the minimum version you've determined or is this simply a known working version? In CI even for BuildStream master we still have a minimum of 1.10 declared and we're testing with 1.34 in CI. And for the C++ library we even still test with gRPC 1.16 (Debian 10), gRPC 1.25 (CentOS 7.7 and Ubuntu 18.04) and gRPC 1.26 (Fedora 32/33), as far as I can tell.

gRPC 1.37 is fairly recent. Can this be an issue on older distros or is this a non-issue as everyone is pulling it from PyPI and there are also no indirect dependency issues?

In any case, any potential issues with old grpcio versions likely also apply to BuildStream master. And thus, this should first be fixed in master before a backport to bst-1.

jjardon:
To be fair I'm not aware of the exact version that fix the problem but I know 1.37 work correctly. Also I know is definitely a problem in fedora 34, which has grpc 1.26

What do you test in the CI exactly? We are using buildbox-casd as remote cache.

Maybe is because you require 1.34 in master? I will change this PR to relax the requirement to that version

jjardon:
PR for master here (1.34 is already in requirements.txt): https://github.com/apache/buildstream/pull/1490

nanonyme:
@juergbi there's a requirement bump induced by RAAPI-based remote cache.

nanonyme:
The Fedora version (1.26.0) at least is known not to work correctly with newer caches.

jjardon:
As we do not know exactly the first version that works, I have further relaxed the requirement to 1.30, which is available in more distros (like debian)

gtristan:
Rebased this in #1497 and merging there

PR #1490, opened 2021-08-03
requirements/requirements.in: It depends on grpc >= 1.34

jjardon:
Make explicit what is already in requirements.txt

with older version you will get errors like this accessing the CAS:
[--:--:--][][] WARNING Failed to initialize remote https://freedesktop-sdk-cache.codethink.co.uk:11001: Received http2 header with status: 404

### DIFF
--- a/requirements/requirements.in
+++ b/requirements/requirements.in
@@ -1,5 +1,5 @@
 Click >= 7.0
-grpcio >= 1.10
+grpcio >= 1.34
 Jinja2 >= 2.10
 pluginbase
 protobuf >= 3.6


gtristan:
To really update grpc, we need to commit the results of `make -C requirements/` to update the `.txt` files used by tox environments in CI.

However, I happen to know this will not pass, partly because of new linting errors and partly because of https://github.com/pallets/click/issues/2037

I've been working on this and should have a patch soon to update the requirements and pass tests soon...

gtristan:
This current just updates the minimum version but doesn't refresh the versions we test with, considering these items to be orthogonal, I will go ahead and just merge this right away.

Separately I am working on #1494 to hopefully fix the CI in master.

PR #1491, opened 2021-08-07
Handled unaliased and mis-aliased sources

### LINKED ISSUE #1119: BuildStream fails to detect missing source alias if already cached
[See original issue on GitLab](https://gitlab.com/BuildStream/buildstream/-/issues/1119)
In GitLab by [[Gitlab user @palvarez89]](https://gitlab.com/palvarez89) on Sep 4, 2019, 16:19

## Summary

[//]: # (Summarize the bug encountered concisely)
If you use in your elements an alias, that isn't defined in your `project.conf`, and you happen to have that source cached, BuildStream will just use the cached source. This behaviour is confusing, as I had some builds working locally and only detected this error when sending my work to the CI.

## Steps to reproduce

[//]: # (How one can reproduce the issue - this is very important)
```bash
# 1 Clone freedesktop sdk
git clone https://gitlab.com/freedesktop-sdk/freedesktop-sdk.git
cd freedesktop-sdk

# 2 Force build glib element, to cache glib sources
vim element/components/glib.bst # Add 'echo 1' into the integration commands
bst build components/glib.bst

# 3 Remove 'gnome:' alias
sed -i -e 's/gnome.*//' include/aliases.yml

# 5 Force rebuild of glib, to reproduce the bug
vim element/components/glib.bst # Add 'echo 2' into the integration commands
bst build components/glib.bst

# 6 Remove cached glib source
find ~/.cache/buildstream/sources/ -name gnome_glib -exec rm -rf "{}" \;

# 7 Attempt to force rebuild glib
vim element/components/glib.bst # Add 'echo 3' into the integration commands
bst build components/glib.bst
```

## What is the current bug behavior?

[//]: # (What actually happens)
After step 5, BuildStream builds `glib` successfully.

[//]: # (What you should see instead)
Step 5 should fail, as it doesn't have the `gnome:` alias defined.

In some cases, you might find yourself copying elements from one project to
another, after building some of these elements from the former. The new
project will not have the aliases needed for those elements, but the
build will succeed. You will only notice these failures when building
in a clean environment (e.g. CI)

## Relevant logs and/or screenshots

[//]: # (Paste any relevant logs - please use code blocks ``` to format console output, logs, and code as it's hard to read otherwise. You can also add a Snippet and link it here. Check the markdown giude on Gitlab for further tips)

## Possible fixes

[//]: # (If you can, link to the line of code that might be responsible for the problem)

## Other relevant information

* BuildStream version affected: /milestone %"BuildStream_v1.4"

----
[//]: # (To review information about possible relevant labels for this issue please view the list of labels: https://gitlab.com/BuildStream/buildstream/labels)

gtristan:
This branch:

* Adds load time errors when a URL refers to a non-existing alias
* Adds load time (fatal-able) warning when using a URL which fails to use an alias

Fixes #1119

### DIFF
--- a/src/buildstream/_context.py
+++ b/src/buildstream/_context.py
@@ -17,7 +17,7 @@
 #  Authors:
 #        Tristan Van Berkom <tristan.vanberkom@codethink.co.uk>
 
-from typing import TYPE_CHECKING, List, Dict, Set, Tuple, Optional, Iterable
+from typing import TYPE_CHECKING, List, Dict, Set, Optional, Iterable
 
 import os
 import shutil
--- a/src/buildstream/source.py
+++ b/src/buildstream/source.py
@@ -164,7 +164,7 @@ from typing import Iterable, Iterator, Optional, Tuple, TYPE_CHECKING
 from . import _yaml, utils
 from .node import MappingNode
 from .plugin import Plugin
-from .types import SourceRef, Union
+from .types import SourceRef, Union, CoreWarnings
 from ._exceptions import BstError, ImplError, PluginError
 from .exceptions import ErrorDomain
 from ._loader.metasource import MetaSource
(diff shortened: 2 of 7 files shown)


gtristan:
Setting WIP for now, need to add tests and pass CI

BenjaminSchubert (approved):
I like it, thanks!

gtristan:
Thanks for reviewing!

One of the checks failed due to the ostree race condition, merging.

PR #1492, opened 2021-08-08
Support for python 3.9

gtristan:
This branch mostly just formally adds some testing support for python 3.9


gtristan:
Oops wrong target branch

PR #1497, opened 2021-08-23
Update grpc minimum version to 1.30

gtristan:
This rebases @jjardon's patch against `bst-1`, replacing pull request #1489


gtristan:
Some tests are still spuriously failing, partly due to using `pytest-xdist`, unrelated to this patch, merging this regardless.

PR #1498, opened 2021-08-24
Update pylint version once more on `bst-1` branch

gtristan:
Now that upstream issue https://github.com/PyCQA/pylint/issues/4837 is fixed.


nanonyme:
Are you sure this is right target branch?

gtristan:
> Are you sure this is right target branch?

Oops thanks for pointing this out.

I actually opened #1499 when I realized I had opened this PR against the wrong branch, and I couldn't find the "change base" feature (I just now discovered you can re-target the PR with a drop down menu when editing the PR)... it looks like I forgot to close this one.

PR #1502, opened 2021-08-25
Fix click deprecation warning for get_terminal_size

nanonyme:
shutil replacement is available since Python 3.3. The click variant will be removed in soon to be released 8.1 which will break bst1.

### DIFF
--- a/buildstream/_frontend/status.py
+++ b/buildstream/_frontend/status.py
@@ -19,6 +19,8 @@
 import os
 import sys
 import curses
+import shutil
+
 import click
 
 # Import a widget internal for formatting time codes
@@ -72,7 +74,7 @@ class Status():
                                      success_profile, error_profile,
                                      stream)
 
-        self._term_width, _ = click.get_terminal_size()
+        self._term_width, _ = shutil.get_terminal_size()
         self._alloc_lines = 0
         self._alloc_columns = None
         self._line_length = 0


gtristan:
Can you rebase this and see if the CI is now working from forks ?

gtristan (approved):

nanonyme:
@gtristan looks like it's now just pending approval for workflow to run.

gtristan:
Looks like CI is working and the lint check failed for the import ordering check

nanonyme:
Force pushed re-ordering and added newline to make it more clear where standard library imports end. I don't know why tests are failing on Ubuntu 18.04. Looks like since I'm still considered first-time contributor, another workflow approval is required.

nanonyme:
@gtristan are we using threading with py.test? It looks like bst1 tests use child threads and timeout constructs through signals which are imcompatible with each other. There was just very exotic stack trace on Debian 10 as a result of this.

gtristan:
Yes, that’ll be `pytest-xdist` which sometimes causes crashes.

thanks for pointing out it is caused by timeouts using sigalarm! Maybe we can finally fix that :)

gtristan:
Do we need this patch on master also ?

nanonyme:
Apparently yes. The same problem is in master https://github.com/apache/buildstream/blob/a935429a8089b12e46a0257372efedd4fa13dbb7/src/buildstream/_frontend/status.py#L67 and https://github.com/apache/buildstream/blob/a935429a8089b12e46a0257372efedd4fa13dbb7/src/buildstream/_frontend/status.py#L219

nanonyme:
Oh, oops. it's in two different places also with bst-1. I'll create a new PR.

PR #1512, opened 2021-09-03
.github: do not test in fedora 32; It's EOL for a while now

jjardon:


juergbi (changes requested):
The image in `tests-template` also needs to be updated to Fedora 33 for jobs that don't override `image`.

gtristan:
Closing as this is included in #1532

PR #1519, opened 2021-09-27
Add `project-root` and `toplevel-root` variables

### LINKED ISSUE #1047: Downloadable file plugins should be able to handle local sources within the project
[See original issue on GitLab](https://gitlab.com/BuildStream/buildstream/-/issues/1047)
In GitLab by [[Gitlab user @jennis]](https://gitlab.com/jennis) on Jun 17, 2019, 16:37

## Summary

I don't think it's uncommon/unreasonable to expect that users may want to use a local file (e.g. a tarball) which they have locally, within their project. The downloadable file source plugins accept a `url` to the file (see [here](https://buildstream.gitlab.io/buildstream/sources/tar.html)) and if we want to use a local file, we can specify a file url.

However, if I want to distribute my project, with the file, the absolute path within the file url will likely cause problems. Therefore I propose that we allow the downloadable source plugin to accept a `path` node, which, like the local source plugin is the relative path to the file from the project's root.

We should take care to ensure that the user specifies either `url` or `path`, and not both.

gtristan:
This addresses the long lasting problem of using sources which usually download from remote URIs with local project-relative user provided data.

This adds two similar variables with slightly different use cases:
* `project-root`: This is to use for source alias and mirror value expansions, and is suitable for things like tarballs which you have committed in your BuildStream project
  * When your project is accessed via a junction, things will *just work*
* `toplevel-root`: This is to use for source alias and mirror value expansions, and is suitable for things like tarballs or git repositories which must be placed somewhere adjacent to your project before launching a build
  * When your project is accessed via a junction, the sources will be appropriately looked up in a location that is relative to the toplevel project being built

Once merged, this will close #1047.

### DIFF
--- a/doc/source/format_declaring.rst
+++ b/doc/source/format_declaring.rst
@@ -652,6 +652,61 @@ read-only variables are also dynamically declared by BuildStream:
 
   The name of project where BuildStream is being used.
 
+* ``project-root`` & ``project-root-uri``
+
+  The directory where the project is located on the host.
+
+  This variable is only available when declaring
+  :ref:`source alias values <project_source_aliases>` or
+  :ref:`source mirror values <project_essentials_mirrors>` and allows
+  access to files in a project on the build host.
+
+  * The ``project-root`` variable is a regular absolute path
+  * The ``project-root-uri`` variable is a properly quoted ``file://`` URI
+
+  .. tip::
+
+     Use this variable to declare :ref:`source alias values <project_source_aliases>`
+     to refer to files which you store as a part of your project, e.g. tarballs
+     which you have committed to you BuildStream project.
+
+  .. attention::
+
+     This feature has been provided for convenience when putting together a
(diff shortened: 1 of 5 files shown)


doc/source/format_declaring.rst
```diff
@@ -652,6 +652,39 @@ read-only variables are also dynamically declared by BuildStream:
...
+  specification of ``file://`` URIs to access files in a project
+  relative location on the build host.
+
+  .. tip::
+
+     Use this variable to declare :ref:`source alias values <project_source_aliases>`
+     to refer to files which you store as a part of your project, e.g. tarballs
+     which you have committed to you BuildStream project.
+
+* ``project-root``
```
juergbi:
```suggestion
* ``toplevel-root``
```

gtristan: Oops fixed.

PR #1524, opened 2021-10-10
Revert to expecting '.bst' on the command line to disambiguate elements and artifacts

### LINKED ISSUE #1411: Commands which accept artifact names and element targets are ambiguous
[See original issue on GitLab](https://gitlab.com/BuildStream/buildstream/-/issues/1411)
In GitLab by [[Gitlab user @tristanvb]](https://gitlab.com/tristanvb) on Nov 19, 2020, 07:06

We need something more explicit for commands which support both artifact names and element names, and then additionally support globbing of the provided artifact names and element names.

## Background

Glob patterns were originally added when adding the artifact commands, and at the same time, it was decided that it would be convenient to also support element names in artifact commands.

In the case that the user is interested in the artifact which a project element currently produces (which is a likely case), then supporting element names in artifact commands provides a short cut so that the user need not manually discover the cache key and derive the artifact name.

Originally, there was a hidden *"trick"* being done under the hood, which cause only glob patterns ending in *".bst"* to be matched against elements - this is sneaky and ugly, and provides a worse experience for commands which accept elements but do not accept artifacts, like `bst show`, which also support glob patterns, but would report no results if the user did not know about specifying this hacky *".bst"* required suffix to the glob pattern.

When fixing the above, we discovered that we now have commands which ambiguously match the same glob pattern to both element targets and artifact names, so we caused this to be an error.

## Original comment reference

The following discussion from !2102 should be addressed:

- [ ] [[Gitlab user @juergbi]](https://gitlab.com/juergbi) started a [discussion](https://gitlab.com/BuildStream/buildstream/-/merge_requests/2102#note_448763866): 

    > > globs are not discriminated against for having `.bst`
    > >
    > > * For a command which only supports elements, then `**` will match all elements in the project as one would expect
    > > * For a command which supports both elements and artifacts, the pattern `**` will now match all elements **and** all artifacts
    > 
    > I'm worried about patterns that can match both elements and artifacts. It would mean that the same pattern leads to different results for different commands (commands only accepting elements and commands accepting both). And I can't think of a reasonable use case for wanting to match both in a single pattern. Especially given that one is specific to the current project state and the other potentially matches artifacts from random projects.
    > 
    > I think it's important to fix this before 2.0 but we don't have to block this MR on that discussion.
    > 
    > Could it make sense to error out when a single glob matches both elements and artifacts until this is discussed and fixed properly? This would make sure nobody starts relying on this odd behavior.

gtristan:
Summary of changes:

  * _stream.py: Treat targets with `.bst` suffixes as elements and other
    targets as artifacts, and improve error reporting around this.

    Also use a new machine readable error string to denote when we fail
    to create a directory when creating a workspace.

  * _loader/loader.py: Remove handling of LoadErrorReason.LOADING_DIRECTORY

    This additional handling was only to suggest that maybe the user
    meant to specify "<directory_name>.bst" in the case a directory is
    encountered, but now we will bail out earlier if an element target
    is specified without a ".bst" suffix anyway.

    The only way to reach this error is to load a directory which itself
    already has a ".bst" suffix, in which case the additional suggestion
    is no longer useful.

  * tests/format/elementnames.py: Expect different error for target elements

  * tests/frontend/artifact_show.py: Removed test that is no longer relevant,
    it is now impossible to glob for both elements and artifacts with the
    same glob expression.

  * tests/frontend/show.py: Removed some parameters of the glob test which
    were expecting to get element results without specifying a ".bst" suffix

  * tests/frontend/workspace.py: Specify a newly added machine readable reason
    for the error of failing to create a directory while opening a workspace

  * tests/internals/loader.py: Trigger the LoadErrorReason.LOADING_DIRECTORY
    error by creating a directory named "element.bst", so that it gets by
    the initial element name suffix checks.

Fixes #1411

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -15,6 +15,10 @@ CLI
   o The `--pull` option is removed from the `bst shell` and `bst artifact checkout`
     commands, which will now unconditionally try to pull required artifacts.
 
+  o BREAKING CHANGE: Now the artifact commands only accept element names if specified
+    with a `.bst` suffix (including wildcard expressions), and otherwise assumes the
+    command means to specify an artifact name.
+
 
 ==================
 buildstream 1.93.6
--- a/src/buildstream/_frontend/cli.py
+++ b/src/buildstream/_frontend/cli.py
@@ -1190,20 +1190,28 @@ def artifact():
 
     \b
     - artifact refs: triples of the form <project name>/<element name>/<cache key>
-    - element paths
+    - element names
 
-    When elements are given, the artifact corresponding to the element is used.
+    When elements are given, the artifact is looked up by observing the element
+    and it's current cache key.
 
     The commands also support shell-style wildcard expansion: `?` matches a
-    single character, and `*` matches zero or more. The patterns are matched
-    against artifact refs by default. If the pattern ends with `.bst` then
-    it matches element paths instead. Some example arguments are:
+    single character, `*` matches zero or more characters but does not match the `/`
+    path separator, and `**` matches zero or more characters including `/` path separators.
+
+    If the wildcard expression ends with `.bst`, then it will be used to search
+    element names found in the project, otherwise, it will be used to search artifacts
+    that are present in the local artifact cache.
+
+    Some example arguments are:
 
     \b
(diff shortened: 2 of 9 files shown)


gtristan:
~~Some documentation still needs to be amended before we can consider merging this.~~

This is done now

PR #1526, opened 2021-10-11
Remove bst-artifact-server

### LINKED ISSUE #1468: Remove `bst-artifact-server`
Now that there exists remote asset servers which can be used as artifact caches and source caches, as noted in #1323, we should drop `bst-artifact-server`.

gtristan:
This fixes #1468


juergbi:
At the very least for testing we need to keep the server around. It makes sense to remove it from the list of entry points to install to make it clear that we recommend better servers for production use. However, for testing, this trivial implementation is sensible, in my opinion.

gtristan:
> At the very least for testing we need to keep the server around. It makes sense to remove it from the list of entry points to install to make it clear that we recommend better servers for production use. However, for testing, this trivial implementation is sensible, in my opinion.

Perhaps we should move it to the tests directory, or better yet, to `buildstream.testing`...

gtristan:
> [...]
> Perhaps we should move it to the tests directory, or better yet, to `buildstream.testing`...

Scratch that, I think it makes better sense to keep it all together with it's family in the `_cas/` subdir.

I'll finish this up by amending documentation and explaining how to use `bb-storage` and `bb-remote-asset`

PR #1528, opened 2021-10-25
Fix integration cache cleanup

juergbi:
bst-plugins-experimental CI is currently failing due to running out of disk space on the shared GitLab runners. This is caused by broken cleanup of integration cache directories. This branch fixes that issue.

### DIFF
--- a/src/buildstream/testing/integration.py
+++ b/src/buildstream/testing/integration.py
@@ -103,11 +103,5 @@ def integration_cache(request):
 
     # Clean up the artifacts after each test session - we only want to
     # cache sources between tests
-    try:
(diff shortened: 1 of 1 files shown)


juergbi:
The bst-plugins-experimental MR https://gitlab.com/BuildStream/bst-plugins-experimental/-/merge_requests/148 depends on this.

gtristan:
Looks good to me, at least we reuse the sources which is the important bit (avoiding costly re-downloading of external sources when running integration tests on my laptop connected to my cell phone data network).

Nuking the artifacts and "source cache" in CAS between modules probably improves the tests overall anyway.

gtristan (approved):

PR #1529, opened 2021-10-27
Update mypy to 0.910

juergbi:
mypy 0.730 is not compatible with Python 3.9, resulting in the following
bogus error message:

    src/buildstream/_cas/cascache.py:417: error: syntax error in type comment


gtristan (approved):
Thanks !

PR #1530, opened 2021-10-28
_stream.py: Fix double pull scheduling on the same element in `bst sh…

gtristan:
…ell`

When running `bst shell --build` we need to add the toplevel element to
the dependency list for the `pull` invocation, but this is not needed
when running without the `--build` option.


juergbi (approved):
I was confused at first why a build shell would pull the artifact of the top-level element. However, I see that this is needed for buildtree support, of course.

PR #1531, opened 2021-10-29
requirements: Limit grpcio to >= 1.34 and < 1.39

gtristan:
We are experiencing segfaults with grpcio > 1.39, consistently with
python3.9 and spuriously with other versions of python too.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -84,6 +84,7 @@ disable=#####################################
         # Messages that are of no use to us #
         #####################################
         ,
+	consider-using-f-string,
         fixme,
         missing-docstring,
         no-self-use,
--- a/requirements/cov-requirements.txt
+++ b/requirements/cov-requirements.txt
@@ -1,12 +1,12 @@
 coverage==4.4
 pytest-cov==2.10.1
-pytest==6.2.4
+pytest==6.2.5
 Cython==0.29.24
 ## The following requirements were added by pip freeze:
 attrs==21.2.0
 iniconfig==1.1.1
 packaging==21.0
-pluggy==0.13.1
+pluggy==1.0.0
 py==1.10.0
-pyparsing==2.4.7
+pyparsing==3.0.3
 toml==0.10.2
--- a/requirements/dev-requirements.txt
+++ b/requirements/dev-requirements.txt
@@ -1,14 +1,14 @@
 pexpect==4.8.0
-pylint==2.10.2
+pylint==2.11.1
 # Pytest 6.0.0 doesn't play well with pylint
-pytest==6.2.4
+pytest==6.2.5
 pytest-datafiles==2.0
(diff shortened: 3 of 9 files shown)


gtristan:
I've narrowed it down to which versions are broken, seems to be `grpcio >= 1.39`

This last test should be checking `1.38.1` https://gitlab.com/BuildStream/buildstream-docker-images/-/jobs/1728633087 which is the latest pre 1.39 version (lets wait for it to be successful, so far I've only tested up to 1.38.0 but I suspect it will be the same as 1.38.1)

gtristan:
oh crap, more linter changes...

nanonyme:
Interesting. bst1 works just fine with grpcio 1.39.1 which is now in Fedora 35. Looks like buildstream-docker-images is out of date and running on old version of Fedora.

juergbi:
> Interesting. bst1 works just fine with grpcio 1.39.1 which is now in Fedora 35. Looks like buildstream-docker-images is out of date and running on old version of Fedora.

The bst2 test suite crashes even on Fedora 35 with grpcio 1.39 installed by tox: https://gitlab.com/BuildStream/buildstream-docker-images/-/jobs/1732134224

nanonyme:
I don't trust grpcio installed with tox much. Is it possible to test with distro grpcio in F35? We made sure it has new enough and working grpcio since it's the most typical production platform for BuildStream (due to gnome-build-meta project)

gtristan:
> I don't trust grpcio installed with tox much. Is it possible to test with distro grpcio in F35? We made sure it has new enough and working grpcio since it's the most typical production platform for BuildStream (due to gnome-build-meta project)

Well, the point is to test with all of the major versions that are not EOL, according to https://fedoraproject.org/wiki/End_of_life it seems that 33 and 34 are not EOL yet (Note that this branch itself has commented out most of the CI in order to test this patch more efficiently). The point of course being to run CI in the buildstream repo for many different active distro versions.

We can of course *add* F35 but this doesn't fix breakage on existing supported platforms.

gtristan:
> > Interesting. bst1 works just fine with grpcio 1.39.1 which is now in Fedora 35. Looks like buildstream-docker-images is out of date and running on old version of Fedora.
> 
> The bst2 test suite crashes even on Fedora 35 with grpcio 1.39 installed by tox: https://gitlab.com/BuildStream/buildstream-docker-images/-/jobs/1732134224

So, to clarify, the newer python from F35 combined with grpcio 1.39 breaks equally.

I think it's fairly safe to deduce that this is a grpcio issue that is triggered with newer versions of python (and that bst2 exercises grpcio in it's tests a lot more than bst1 does, exposing this issue).

At this point we've found a range of grpcio versions which works, if we had more info on this we could report it upstream to grpcio and hope for a fix in a future `> 1.41.1` version.

PR #1537, opened 2021-11-10
Override project defined source mirrors, and be selective about what mirrors to use

### LINKED ISSUE #1484: Override aliases and mirrors in junctioned projects
This would be an essential feature to allow projects to store all the source code in their own mirrors, it is important for the parent/main project to store all of the source code and ensure that it can be built in perpetuity.

gtristan:
This fixes #1484, and implements the [proposal](https://lists.apache.org/api/source.lua?id=1w3s72q4yf23rx0yc57tw0lf94q3g5bj)

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -12,6 +12,9 @@ Format
     contain invalid characters, as was already documented. Now the warning
     is replaced by an unconditional error.
 
+  o User configuration now has the ability to override source mirrors defined
+    by project configuration.
+
 CLI
 ---
   o BREAKING CHANGE: Removed `--deps plan` from all commands
--- a/NEWS
+++ b/NEWS
@@ -15,6 +15,9 @@ Format
   o User configuration now has the ability to override source mirrors defined
     by project configuration.
 
+  o User configuration now has fetch source and track source configurations to
+    control what mirrors/URIs can be accessed to obtain sources.
+
 CLI
 ---
   o BREAKING CHANGE: Removed `--deps plan` from all commands
--- a/doc/source/format_project.rst
+++ b/doc/source/format_project.rst
@@ -287,9 +287,10 @@ list of URIs, e.g.
 The order that the mirrors (and the URIs therein) are consulted is in the order
 they are defined when fetching, and in reverse-order when tracking.
 
-A default mirror to consult first can be defined via
-:ref:`user config <config_default_mirror>`, or the command-line argument
-:ref:`--default-mirror <invoking_bst>`.
+The mirrors can be overridden on a per project basis using
+:ref:`user configuration <config_mirrors>`. One can also specify which mirror should
+be used first in the :ref:`user configuration <config_default_mirror>`, or using
+the  :ref:`--default-mirror <invoking_bst>` command-line argument.
 
 
 .. _project_plugins:
--- a/doc/source/using_config.rst
+++ b/doc/source/using_config.rst
@@ -1060,13 +1060,45 @@ modifying some low level component.
    the ``--strict`` and ``--no-strict`` command line options.
 
 
+.. _config_mirrors:
+
+Mirrors
+~~~~~~~
+Project defined :ref:`mirrors <project_essentials_mirrors>`, can be overridden
+with user configuration. This is helpful when you need to mirror all of the source
+code used by subprojects and ensure that your project can be built in perpetuity.
+
+**Example**
+
+.. code:: yaml
+
+   projects:
+     project-name:
+       mirrors:
+       - name: middle-earth
+         aliases:
+           foo:
+           - http://www.middle-earth.com/foo/1
+           - http://www.middle-earth.com/foo/2
+           bar:
+           - http://www.middle-earth.com/bar/1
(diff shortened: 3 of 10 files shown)


jjardon:
@gtristan Would it be possible to backport this to bst-1? We need this feature to use our own mirrors in GNOME

gtristan:
This branch has been updated to provide full test coverage.

gtristan:
> @gtristan Would it be possible to backport this to bst-1? We need this feature to use our own mirrors in GNOME

It seems that this would be technically possible yes. It would be a bit more fragile in `bst-1` due to project name uniqueness not being strictly enforced (mostly the addressing scheme works only with junction names, although we already have the `projects` user configuration section in `bst-1`) - although in practice this would not realistically cause issues.

gtristan:
This actually passes it's tests, however is failing in CI due to #1368 (well known spurious failure) and now #1549 which is causing all our workflows to fail.

gtristan:
As it was questioned in #1542, should we:

* Remove the strange behavior of traversing active mirrors/URLs backwards when tracking ?
* Make the default track source policy be to ignore mirrors and track only from the originally defined aliases ?
  * This would IMO address the concern that we *want* to track upstream by default even though we want to *fetch* from our own mirrors, in order to ensure we get new refs from upstream even when our downstream mirroring solutions might be lagging behind or broken

jjardon:
> As it was questioned in #1542, should we:
> 
>     * Remove the strange behavior of traversing active mirrors/URLs backwards when tracking ?
> 
>     * Make the default track source policy be to ignore mirrors and track only from the originally defined aliases ?
>       
>       * This would IMO address the concern that we _want_ to track upstream by default even though we want to _fetch_ from our own mirrors, in order to ensure we get new refs from upstream even when our downstream mirroring solutions might be lagging behind or broken

I can see we want to get upstream to get the latest, but then we have another problem if we use a tag that is not in our mirror as next build fetch will fail (if we restrict your project to only use mirrors, for example)

So, I think default should be always be (for both track and fetch):
- 1. Try to use mirror
- 2. Try to use upstream

but that order should be configurable separately for track and fetch, so you can select the behaviour you want for your project

BenjaminSchubert (approved):
I did not do a thorough review, but tests and documentation seem good.

gtristan:
> 
> I can see we want to get upstream to get the latest,

Right, that is the rationale behind the original special case to reverse the URLs when tracking - which is fairly obsolete by this branch.

> but then we have another problem if we use a tag that is not in our mirror as next build fetch will fail (if we restrict your project to only use mirrors, for example)

Which I think is desirable, we want to have an error as early as possible when our mirrors are not working right ?

> So, I think default should be always be (for both track and fetch):
> 
>     * 1. Try to use mirror
> 
>     * 2. Try to use upstream

Agree for `fetch`, disagree for `track`, I think by default tracking should *only* track upstreams.

Other than that, I don't think that an order is very necessary.

It's important to note that once a remote is *reached* and successfully tracked, then we will *not* rollover to the next mirror to see if we get anything more recent (and there is no concept of "recent", only what happens to be at the tip of a tracking branch).

This is to say that if you track your mirror first and successfully reach it, and it does **not** have the latest upstream yet (because the mirror is not mirroring correctly), then we would silently ignore the upstream.

> but that order should be configurable separately for track and fetch, so you can select the behaviour you want for your project

I'm skeptical about retaining configurability of the reverse order - with the current approach the user can dictate an order of mirrors to use in `buildstream.conf` anyway, by way of explicitly overriding project declared mirrors, but I doubt that is useful in practice.

gtristan:
Updated branch so that:
* We remove the obsolete hard code which reverses URIs when tracking
* Make the default track policy "aliases", so that (by default) we track only from original upstreams instead of mirrors which could be out of date

juergbi:
> Updated branch so that:
> 
>     * We remove the obsolete hard code which reverses URIs when tracking
> 
>     * Make the default track policy "aliases", so that (by default) we track only from original upstreams instead of mirrors which could be out of date

This seems reasonable to me and the branch looks good to me - once the failing tests have been fixed.

It might be possible to reduce the load on upstream/alias servers (at least for git) if we fetch from a mirror before we fetch from upstream for actual tracking. That way BuildStream would download very little from upstream, unless the mirror is out-of-date, without any risk of getting stale or wrong information from a mirror when tracking. However, I don't know whether this would be an important optimization. I certainly don't consider this a blocker for this PR.

gtristan:
> It might be possible to reduce the load on upstream/alias servers (at least for git) if we fetch from a mirror before we fetch from upstream for actual tracking. That way BuildStream would download very little from upstream, unless the mirror is out-of-date, without any risk of getting stale or wrong information from a mirror when tracking. However, I don't know whether this would be an important optimization. I certainly don't consider this a blocker for this PR.

That is the current (unchanged) behavior. The reversal which I removed was a special case for running `bst track` only, this PR doesn't change the default behavior when all URI sources are allowed, which is to place mirrors before the default upstream.

PR #1543, opened 2021-12-09
remove things no longer needed after dropping bst-artifact-server

abderrahim:


src/buildstream/_cas/casserver.py
```diff
@@ -48,7 +43,6 @@
 # Not enough that we'd like to duplicate code, but enough that we want
 # to make it very obvious what we're using, so in this case we import
 # the specific methods we'll be using.
```
juergbi: I think we can also drop this whole note as we no longer intend to split this module off.

abderrahim: Dropped.

Issue #1475, opened 2021-05-06
Drop deprecated functions in buildstream._yaml


nanonyme:
There's tons of warnings being spewed
```
round_trip_dump will be removed, use
  
    yaml=YAML()
    yaml.dump(...)
  
  instead
    _yaml.roundtrip_dump(config, filename)
```
in various bst-plugins-experimental tests because BuildStream is using deprecated API

PR #1546, opened 2021-12-13
_yaml.pyx: port roundtripping code to the new ruamel.yaml API

### LINKED ISSUE #1495: _yaml.pyx deletes comments when dumping using recent versions of ruamel.yaml
With recent versions of `ruamel.yaml`, our test case `tests/internals/yaml.py::test_roundtrip_dump` is failing, and deleting comments from the test file.

This is [filed upstream](https://sourceforge.net/p/ruamel-yaml/tickets/390/), but is not reproducible with vanilla ruamel, this only happens with our `_yaml.pyx` customizations.

### Version 0.17.6

Since this version of `ruamel.yaml`, our loading/dumping routine results in deleting comments from the `roundtrip-test.yaml` sample file.

The `roundtrip-test.yaml` file is changed in the following way:

```diff
24,26d23
< 
< # We roundtrip integers and floats as strings, to prevent truncation.
< 
32,35d28
< 
< # We also roundtrip booleans in various forms as strings to prevent
< # normalisation to 'True' and 'False'
< 
51d43
< 
53,55d44
< 
< # That is all
< 
```

### Version 0.17.5

In this version of `ruamel.yaml`, the test is broken more severely with these versions, instead of losing comments we simply get the following crash (which is an instance of `ruamel.yaml` [issue 385](https://sourceforge.net/p/ruamel-yaml/tickets/385/)):

```
datafiles = local('/home/tristan/work/buildstream/.tox/py39-nocover/tmp/test_roundtrip_dump_True_0'), fromdisk = True

    @pytest.mark.datafiles(os.path.join(DATA_DIR))
    @pytest.mark.parametrize("fromdisk", [(True), (False)])
    def test_roundtrip_dump(datafiles, fromdisk):
        filename = os.path.join(datafiles.dirname, datafiles.basename, "roundtrip-test.yaml")
        with open(filename, "r") as fh:
            rt_raw = fh.read()
        if fromdisk:
>           rt_loaded = _yaml.roundtrip_load(filename)

tests/internals/yaml.py:415: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
src/buildstream/_yaml.pyx:427: in buildstream._yaml.roundtrip_load
    contents = roundtrip_load_data(data, filename=filename)
src/buildstream/_yaml.pyx:461: in buildstream._yaml.roundtrip_load_data
    contents = yaml.load(contents, yaml.RoundTripLoader, preserve_quotes=True)
.tox/py39-nocover/lib/python3.9/site-packages/ruamel/yaml/main.py:1067: in load
    return loader._constructor.get_single_data()
.tox/py39-nocover/lib/python3.9/site-packages/ruamel/yaml/constructor.py:120: in get_single_data
    node = self.composer.get_single_node()
.tox/py39-nocover/lib/python3.9/site-packages/ruamel/yaml/composer.py:76: in get_single_node
    document = self.compose_document()
.tox/py39-nocover/lib/python3.9/site-packages/ruamel/yaml/composer.py:99: in compose_document
    node = self.compose_node(None, None)
.tox/py39-nocover/lib/python3.9/site-packages/ruamel/yaml/composer.py:142: in compose_node
    node = self.compose_mapping_node(anchor)
.tox/py39-nocover/lib/python3.9/site-packages/ruamel/yaml/composer.py:222: in compose_mapping_node
    item_value = self.compose_node(node, item_key)
.tox/py39-nocover/lib/python3.9/site-packages/ruamel/yaml/composer.py:142: in compose_node
    node = self.compose_mapping_node(anchor)
.tox/py39-nocover/lib/python3.9/site-packages/ruamel/yaml/composer.py:222: in compose_mapping_node
    item_value = self.compose_node(node, item_key)
.tox/py39-nocover/lib/python3.9/site-packages/ruamel/yaml/composer.py:142: in compose_node
    node = self.compose_mapping_node(anchor)
.tox/py39-nocover/lib/python3.9/site-packages/ruamel/yaml/composer.py:222: in compose_mapping_node
    item_value = self.compose_node(node, item_key)
.tox/py39-nocover/lib/python3.9/site-packages/ruamel/yaml/composer.py:112: in compose_node
    if self.parser.check_event(AliasEvent):
.tox/py39-nocover/lib/python3.9/site-packages/ruamel/yaml/parser.py:145: in check_event
    self.current_event = self.state()
.tox/py39-nocover/lib/python3.9/site-packages/ruamel/yaml/parser.py:636: in parse_block_mapping_value
    return self.parse_block_node_or_indentless_sequence()
.tox/py39-nocover/lib/python3.9/site-packages/ruamel/yaml/parser.py:350: in parse_block_node_or_indentless_sequence
    return self.parse_node(block=True, indentless_sequence=True)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <ruamel.yaml.loader.RoundTripLoader object at 0x7ff65bd85040>, block = True, indentless_sequence = True

    def parse_node(self, block=False, indentless_sequence=False):
        # type: (bool, bool) -> Any
        if self.scanner.check_token(AliasToken):
            token = self.scanner.get_token()
            event = AliasEvent(token.value, token.start_mark, token.end_mark)  # type: Any
            self.state = self.states.pop()
            return event
    
        anchor = None
        tag = None
        start_mark = end_mark = tag_mark = None
        if self.scanner.check_token(AnchorToken):
            token = self.scanner.get_token()
            self.move_token_comment(token)
            start_mark = token.start_mark
            end_mark = token.end_mark
            anchor = token.value
            if self.scanner.check_token(TagToken):
                token = self.scanner.get_token()
                tag_mark = token.start_mark
                end_mark = token.end_mark
                tag = token.value
        elif self.scanner.check_token(TagToken):
            token = self.scanner.get_token()
            start_mark = tag_mark = token.start_mark
            end_mark = token.end_mark
            tag = token.value
            if self.scanner.check_token(AnchorToken):
                token = self.scanner.get_token()
                start_mark = tag_mark = token.start_mark
                end_mark = token.end_mark
                anchor = token.value
        if tag is not None:
            handle, suffix = tag
            if handle is not None:
                if handle not in self.tag_handles:
                    raise ParserError(
                        'while parsing a node',
                        start_mark,
                        _F('found undefined tag handle {handle!r}', handle=handle),
                        tag_mark,
                    )
                tag = self.transform_tag(handle, suffix)
            else:
                tag = suffix
        # if tag == '!':
        #     raise ParserError("while parsing a node", start_mark,
        #             "found non-specific tag '!'", tag_mark,
        #      "Please check 'http://pyyaml.org/wiki/YAMLNonSpecificTag'
        #     and share your opinion.")
        if start_mark is None:
            start_mark = end_mark = self.scanner.peek_token().start_mark
        event = None
        implicit = tag is None or tag == '!'
        if indentless_sequence and self.scanner.check_token(BlockEntryToken):
            comment = None
            pt = self.scanner.peek_token()
>           if self.loader and self.loader.comment_handling is None:
E           AttributeError: 'RoundTripLoader' object has no attribute 'comment_handling'

.tox/py39-nocover/lib/python3.9/site-packages/ruamel/yaml/parser.py:413: AttributeError
```
### Version 0.17.4

This is the last version of `ruamel.yaml` which behaves how we expect it to.

abderrahim:
Also drop roundtrip_load_data that was unused

Fixes #1495 (and related to #1475)


abderrahim:
The problem I'm noticing with this is that if I run the tests in parallel, they often fail with errors like "I/O operation on closed file" or "expected DocumentStartEvent and got MappingStartEvent".

This may be caused by multiple threads trying to use the same object. Do we have something in buildstream that uses thread local storage?

gtristan:
> The problem I'm noticing with this is that if I run the tests in parallel, they often fail with errors like "I/O operation on closed file" or "expected DocumentStartEvent and got MappingStartEvent".
> 
> This may be caused by multiple threads trying to use the same object. Do we have something in buildstream that uses thread local storage?

Yes, the `Messenger` uses thread local storage for per-task logging file handles and such.

abderrahim:
I've made it so a new instance is created on each call (like it was previously). I can explore thread local storage if you think it's worth it.

gtristan:
> I've made it so a new instance is created on each call (like it was previously). I can explore thread local storage if you think it's worth it.

I don't see that change here, in master `Messenger` has `_MessengerLocal` which is TLS for the messenger.

Also, can we check if this change fixes `tests/internals/yaml.py::test_roundtrip_dump` which we currently have on XFAIL ?

gtristan:
> The problem I'm noticing with this is that if I run the tests in parallel, they often fail with errors like "I/O operation on closed file" or "expected DocumentStartEvent and got MappingStartEvent".
> 
> This may be caused by multiple threads trying to use the same object. Do we have something in buildstream that uses thread local storage?

I ran the full tests with `-n 12` and didn't see any of these errors... I'll run it again and report back if I see any of these...

Also, after running the full tests, I'm still seeing:

```
  /home/tristan/work/buildstream/.tox/py39-nocover/lib/python3.9/site-packages/buildstream/element.py:1684: PendingDeprecationWarning: 
  round_trip_dump will be removed, use
  
    yaml=YAML()
    yaml.dump(...)
  
  instead
    env_dump = yaml.round_trip_dump(self.get_environment(), default_flow_style=False, allow_unicode=True)
```

Although this looks like much less deprecation warnings than before the patch.

abderrahim:
Maybe I wasn't clear enough:
- The implementation in master uses one instance per call (via the current implementation of the deprecated API in ruamel.yaml)
- My first port used a single instance for everything and failed the tests (I assume because the API is not thread safe)
- The latest version of the PR goes back to using one instance per call
- I guess the optimal solution would be to use thread local storage and use one instance per thread (but I didn't implement it).

Note that I only said this is related to #1475, not that it fixes it. It fixes #1495. The remaining deprecation notices are due to usage of ruamel.yaml from outside of `_yaml.pyx`, see https://github.com/apache/buildstream/blob/master/src/buildstream/_frontend/widget.py#L378. The problem with those is that there is no replacement for the APIs that return a string, we would need to use a StringIO or something to pass to ruamel, but if we're doing that anyway maybe it's a good idea to rewrite the code to use a single StringIO for everything.

abderrahim:
Maybe I wasn't clear enough:
- The implementation in master uses one instance per call (via the current implementation of the deprecated API in ruamel.yaml)
- My first port used a single instance for everything and failed the tests (I assume because the API is not thread safe)
- The latest version of the PR goes back to using one instance per call
- I guess the optimal solution would be to use thread local storage and use one instance per thread (but I didn't implement it).

Note that I only said this is related to #1475, not that it fixes it. It fixes #1495. The remaining deprecation notices are due to usage of ruamel.yaml from outside of `_yaml.pyx`, see https://github.com/apache/buildstream/blob/master/src/buildstream/_frontend/widget.py#L378. The problem with those is that there is no replacement for the APIs that return a string, we would need to use a StringIO or something to pass to ruamel, but if we're doing that anyway maybe it's a good idea to rewrite the code to use a single StringIO for everything.

gtristan:
> Maybe I wasn't clear enough:
> 
>     * The implementation in master uses one instance per call (via the current implementation of the deprecated API in ruamel.yaml)
>     * My first port used a single instance for everything and failed the tests (I assume because the API is not thread safe)
>     * The latest version of the PR goes back to using one instance per call
>     * I guess the optimal solution would be to use thread local storage and use one instance per thread (but I didn't implement it).
> 

Ah, you mean one instance _of a YAML handle_ per call... yeah this sounds perfectly good to me.

> Note that I only said this is related to #1475, not that it fixes it. It fixes #1495. The remaining deprecation notices are due to usage of ruamel.yaml from outside of `_yaml.pyx`, see https://github.com/apache/buildstream/blob/master/src/buildstream/_frontend/widget.py#L378. The problem with those is that there is no replacement for the APIs that return a string, we would need to use a StringIO or something to pass to ruamel, but if we're doing that anyway maybe it's a good idea to rewrite the code to use a single StringIO for everything.

Got it.

Ok so... IIUC we can pretty much merge this and we'll have made some progress, and there are no CI errors because of the new approach this patch takes.

The other parts which use `round_trip_dump()` we could have a private `utils` function which fiddles with a `StringIO` to cover those parts, which could be fine to do separately.

abderrahim:
> Ok so... IIUC we can pretty much merge this and we'll have made some progress, and there are no CI errors because of the new approach this patch takes.

That's correct.

> The other parts which use round_trip_dump() we could have a private utils function which fiddles with a StringIO to cover those parts, which could be fine to do separately.

Yup.

gtristan (approved):
Let's go ahead and merge this !

PR #1547, opened 2021-12-16
Pure python cython

abderrahim:
A proof of concept for #1412


src/buildstream/_types.py
```diff
@@ -72,11 +73,11 @@ def set_values(mcs, kls, data):
 
         type.__setattr__(kls, "_value_to_entry", value_to_entry)
 
-    def __repr__(self):
-        return "<fastenum '{}'>".format(self.__name__)
+    def __repr__(cls):
```
nanonyme, line 76: Looks weird. This looks like it assumes it's a class method but there's no classmethod decorator. Is this change because it's a metaclass?

nanonyme: Same thing for other changes self->cls below. Is there something weird going on here that you don't have to use `self.__class__.__name__`?

PR #1552, opened 2022-01-07
bst-1: .github: Test in current supported fedora34 (32 and 33 are EOL)

jjardon:


jjardon:
Let's do this as part of https://github.com/apache/buildstream/pull/1553

jjardon:
> Let's do this as part of #1553

#1553 is currently failing for fedora35, let's merge this in the meantime

PR #1553, opened 2022-01-07
bst-1: Test in fedora 35

jjardon:


juergbi (changes requested):
The tests are skipped on Fedora 35 because Python 3.10 is not in the configured list of environments.

We need to add py310 to `tox.ini` and `.github/common.env`. If tests pass with this change, this looks good to me.

juergbi:
`python3.10` also needs to be added to `buildstream/plugins/sources/pip.py`. I don't know why we can't simply use `python3` there but that's better addressed separately (and first in master).

juergbi (approved):

PR #1554, opened 2022-01-07
Drop support for Python 3.6

abderrahim:


nanonyme:
Linter breaks because sys is now unnecessarily imported. Fedora has the crashing buildbox-casd. But yes, so the real impact is this actively drops support from CentOS 7.7 and Ubuntu 18.04. While they are both maintained for far future still, Python 3.6 has no future with regards to Python ecosystem anywhere near that far (it already went EOL this Winter) and it is going to be dramatically hard to maintain this project in the future without allowing installation of packages that require Python > 3.6.

juergbi:
For master I think we definitely want to drop support for Python 3.6, dropping CentOS 7 and Ubuntu 18.04 from CI. Note that the current Red Hat Software Collections available for RHEL 7 officially provides Python 3.8 and this can also be installed on CentOS (`rh-python38`), however, I don't think we need to test this in CI.

For BuildStream 1.6 we should probably retain Python 3.6 support. CI images would have to be frozen as I'd like to drop CentOS 7 and Ubuntu 18.04 from buildstream-docker-images. For an unlikely BuildStream 1.8 release, we could consider dropping Python 3.6 support.

juergbi (changes requested):
This also requires updates to `NEWS`, `doc/source/main_install.rst`, `tox.ini` and `.github/common.env`.

.github/compose/ci.docker-compose.yml
```diff
@@ -34,14 +34,6 @@ services:
     <<: *tests-template
     image: registry.gitlab.com/buildstream/buildstream-docker-images/testsuite-debian:10-${CI_IMAGE_VERSION:-latest}
 
-  ubuntu-18.04:
-    <<: *tests-template
-    image: registry.gitlab.com/buildstream/buildstream-docker-images/testsuite-ubuntu:18.04-${CI_IMAGE_VERSION:-latest}
```
jjardon, line 39: while dropping support for python2.6 maybe be ok, I think we still need to check bst-1 runs fine in ubuntu-18.04 and centos-7 (ubuntu has python2.7 for example)

abderrahim: Do you mean 3.7? It has indeed both 3.7 and 3.8. But ubuntu 18.04 is pretty old, and we should probably use 20.04 at this point.

jjardon: Yeah sorry I meant 3.7

jjardon: CCing @ssssam as he might have an opinion on this

jjardon: Ah wait, this is for master, nevermind

juergbi (approved):

PR #1560, opened 2022-01-13
Fix shutdown of test server

juergbi:
buildbox-casd processes used by test servers were not terminated and kept running after the end of the tests (or crashed because the test deleted the directories used by buildbox-casd).

### DIFF
--- a/tests/testutils/artifactshare.py
+++ b/tests/testutils/artifactshare.py
@@ -1,7 +1,6 @@
 import os
 import shutil
 import signal
-import sys
 from collections import namedtuple
 from contextlib import ExitStack, contextmanager
 from concurrent import futures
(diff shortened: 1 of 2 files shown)


tests/testutils/artifactshare.py
```diff
@@ -67,7 +57,17 @@ def run(self, q):
...
+
+            server.stop(0)
+
+        # Save collected coverage data
+        try:
+            from pytest_cov.embed import cleanup
+        except ImportError:
+            pass
+        else:
+            cleanup()
```
BenjaminSchubert, line 70:
```suggestion
        with suppress(ImportError):
            from pytest_cov.embed import cleanup
           cleanup()
```
would that be more readable?

BenjaminSchubert (approved):
Seems good, thanks

tests/testutils/artifactshare.py
juergbi: It would be a bit more concise, however, pylint raises an 'import-error', which can be suppressed but reduces the readability advantage, in my opinion. Also, ignoring potential import errors inside the `cleanup()` function is not ideal, even though this is unlikely to be an issue. I'll leave it as is for now.

PR #1562, opened 2022-01-14
use ruamel-yaml < 0.17 

abderrahim:
somewhere in the 0.17 series, ruamel.yaml broke round-tripping for the old API (See #1495)

### DIFF
--- a/requirements/cov-requirements.txt
+++ b/requirements/cov-requirements.txt
@@ -7,5 +7,5 @@ packaging==21.3
 pluggy==1.0.0
 py==1.11.0
 pyparsing==3.0.7
-pytest==7.0.0
+pytest==7.0.1
 tomli==2.0.1
--- a/requirements/dev-requirements.txt
+++ b/requirements/dev-requirements.txt
@@ -1,6 +1,6 @@
 pep8==1.7.1
 pylint==2.12.2
-pytest==7.0.0
+pytest==7.0.1
 pytest-datafiles==2.0
 pytest-env==0.6.2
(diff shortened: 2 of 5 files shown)


gtristan:
This needs a rebase after applying #1564.

In fact, I overlooked that #1564 was applied in error, that's my fault and probably needs to be fixed in some way.

Updating the dependencies in this way leads to an inability to install on `ubuntu-18.04` and `centos-7.7.1908`.

abderrahim:
> This needs a rebase after applying #1564.
> 
> In fact, I overlooked that #1564 was applied in error, that's my fault and probably needs to be fixed in some way.
> 
> Updating the dependencies in this way leads to an inability to install on `ubuntu-18.04` and `centos-7.7.1908`.

Yeah, I noticed that. But I think it's better done in a different PR. So the problem is that those distros use EOL version of python (3.6), so if I update the requirements using a new version of python I end up with new versions of dependencies that don't support python 3.6.

I also noticed that in `requirements/cov-requirements.in`, there are different versions depending on python_version which means that `requirements/cov-requirements.txt` (which CI uses) will contain either depending on the version of python used to generate it.

I'm trying to regenerate requirements using python 3.6, let's see what it gives.

abderrahim:
Let's see if #1586 fixes CI.

gtristan:
Can we get a rebase of this ? The CI failure issue should be fixed by merging #1591.

nanonyme:
@gtristan this should now be good to merge, right?

PR #1563, opened 2022-01-14
sandbox/_sandboxbwrap.py: don't try to delete symlinks as directories

abderrahim:
This prevents breaking projects that have workarounds for it


gtristan (approved):

abderrahim:
@gtristan Looks like my first approach was wrong, I re-did it.

PR #1565, opened 2022-01-15
workspace.py: Move bzr specific test to the bzr source

BenjaminSchubert:

PR #1566, opened 2022-01-15
workspace.py: Remove assumption on specific kinds to testing multiple

BenjaminSchubert:
This makes the test more explicit in testing that it can actually open
multiple different projects at the same time, and removes the need of
actually using multiple kinds

### DIFF
--- a/tests/frontend/workspace.py
+++ b/tests/frontend/workspace.py
@@ -29,6 +29,7 @@
 import os
 import stat
 import shutil
+import tempfile
 
 import pytest
 
@@ -73,8 +74,19 @@ class WorkspaceCreator:
 
         # Create our repo object of the given source type with
         # the bin files, and then collect the initial ref.
-        repo = create_repo(kind, str(self.tmpdir))
-        ref = repo.create(self.bin_files_path)
+        # And ensure we store it in a suffix-specific directory, to avoid clashes
+        # if using multiple times the same kind element here.
+        repo = create_repo(kind, str(self.tmpdir), "repo-for-{}".format(element_name))
+
+        with tempfile.TemporaryDirectory() as tempdir:
+            dst_repo = os.path.join(tempdir, "repo")
+            shutil.copytree(self.bin_files_path, dst_repo)
+            # Touch a file with the element name in, to allow validating that this
+            # is the correct repo
+            # pylint: disable=consider-using-with
+            open(os.path.join(dst_repo, element_name), "a", encoding="utf-8").close()
+
+            ref = repo.create(os.path.join(tempdir, "repo"))
 
         # Write out our test target
         element = {"kind": "import", "sources": [repo.source_config(ref=ref)]}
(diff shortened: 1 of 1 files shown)


tests/frontend/workspace.py
```diff
@@ -175,20 +187,13 @@ def test_open_bzr_customize(cli, tmpdir, datafiles):
...
 def test_open_multi(cli, tmpdir, datafiles):
-
     workspace_object = WorkspaceCreator(cli, tmpdir, datafiles)
     workspaces = workspace_object.open_workspaces(repo_kinds)
 
     for (elname, workspace), kind in zip(workspaces, repo_kinds):
         assert kind in elname
         workspace_lsdir = os.listdir(workspace)
-        if kind == "git":
-            assert ".git" in workspace_lsdir
```
juergbi, line 168: Do we lose the check that a workspace on a git source has a `.git` directory? Or is such a check irrelevant or redundant (it might make sense to check this in a plugin-specific test)?

BenjaminSchubert: For git, we have https://github.com/apache/buildstream/blob/master/tests/sources/git.py#L64-L86 that ensures we can successfully run `git`.

For `bzr`, we have https://github.com/apache/buildstream/pull/1565/files that also tangentially ensures we have a .bzr directory.

If you think it's better to have explicit tests, I am happy to add some in the source-specific entries.

juergbi: Tests that run `git`/`bzr` on the workspace directory should indeed be sufficient.

juergbi (approved):

gtristan:
We added tests to ensure that by default we don’t stage the git/bzr history (when we made that behavior default), we should keep these assertions in some form somewhere.

BenjaminSchubert:
> We added tests to ensure that by default we don’t stage the git/bzr history (when we made that behavior default), we should keep these assertions in some form somewhere.

This PR doesn't change this. It merely removes a fuzzy validation that the workspace was opened correctly. We have tests in each of those sources that make sure of this.

gtristan:
> > We added tests to ensure that by default we don’t stage the git/bzr history (when we made that behavior default), we should keep these assertions in some form somewhere.
> 
> This PR doesn't change this. It merely removes a fuzzy validation that the workspace was opened correctly. We have tests in each of those sources that make sure of this.

Technically this testing should be in the domain of the sources themselves, but do remember that some sources (probably `git` and `bzr` included) use *different* codepaths for opening a workspace than they do for merely staging sources (see `Source.init_workspace()`).

If this assertion exists for opening workspaces on `git` / `bzr` *somewhere* then lets immediately just remove these.

BenjaminSchubert:
@gtristan absolutely:

For git, we have https://github.com/apache/buildstream/blob/master/tests/sources/git.py#L64-L86 that ensures we can successfully run `git`.

For `bzr`, we have https://github.com/apache/buildstream/pull/1565/files that also tangentially ensures we have a .bzr directory.

So merging this for now :) thanks!

PR #1567, opened 2022-01-17
fuse/hardlinks: Do not follow symlinks for chown

doraskayo:
This PR fixes the following issue detected by `gnulib`:
```
...
checking for fchownat... yes
checking whether fchownat works with AT_SYMLINK_NOFOLLOW... no
```

See commit messages for more information.

### DIFF
--- a/buildstream/_fuse/hardlinks.py
+++ b/buildstream/_fuse/hardlinks.py
@@ -117,8 +117,8 @@ class SafeHardlinkOps(Operations):
         full_path = self._full_path(path)
 
         # Ensure copies on chown
-        self._ensure_copy(full_path)
-        return os.chown(full_path, uid, gid)
+        self._ensure_copy(full_path, follow_symlinks=False)
+        return os.chown(full_path, uid, gid, follow_symlinks=False)
 
     def getattr(self, path, fh=None):
         full_path = self._full_path(path)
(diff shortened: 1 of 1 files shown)


nanonyme:
Is this not a problem for chmod?

nanonyme (approved):

juergbi:
> Is this not a problem for chmod?

I indeed expect the same logic to be needed for `chmod` and `chown`.

Actually, as I understand it, the low level (kernel) FUSE operations are all inode-based (to the extent possible) and libfuse maps inodes to paths to make it easier to implement (certain types of) filesystems. I.e., I would expect that we should never follow symlinks.

doraskayo:
@nanonyme, @juergbi, you're most likely correct regarding `chmod`, but I think it's not an issue in reality because the `AT_SYMLINK_NOFOLLOW` flag is not currently respected at the kernel side either (see [fchmodat(2)](https://linux.die.net/man/2/fchmodat)). It may become an issue if it's ever implemented in the kernel, however.

I mostly didn't have time to confirm and test it, so I considered it out of scope for now.

nanonyme:
Okay, then this sounds good to merge. It would be nice to have actual test suite that would fail if the FUSE implementation is noncompliant for future purposes.

PR #1568, opened 2022-01-17
fuse: Pass through the underlying filesystem's inode numbers

doraskayo:
This PR fixes the following issue detected by `gnulib`:
```
...
checking whether this system has an arbitrary file name length limit... yes
checking for d_ino member in directory struct... no
```
See commit messages for more information.

### DIFF
--- a/buildstream/_fuse/hardlinks.py
+++ b/buildstream/_fuse/hardlinks.py
@@ -126,11 +126,11 @@ class SafeHardlinkOps(Operations):
     def readdir(self, path, fh):
         full_path = self._full_path(path)
 
-        dirents = ['.', '..']
+        dir_entries = ['.', '..']
         if os.path.isdir(full_path):
-            dirents.extend(os.listdir(full_path))
-        for r in dirents:
-            yield r
+            dir_entries.extend(os.listdir(full_path))
+        for entry in dir_entries:
+            yield entry
 
     def readlink(self, path):
         pathname = os.readlink(self._full_path(path))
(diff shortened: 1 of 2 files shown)


nanonyme (commented):
It makes me wonder if this can have impact on reproducibility on different machines where you have different inode numbers for same content.

nanonyme (approved):

juergbi:
A potential issue I see is that `ensure_copy()` can cause the inode number of an existing file to change when it shouldn't. I don't know whether there are any applications where this would cause an issue. It's likely that filling in the inode number is more important than the potential inconsistency.

OSTree's `rofiles-fuse` also passes through the inode number, an indication that this is not a significant issue.
https://github.com/ostreedev/ostree/blob/main/src/rofiles-fuse/main.c#L137

doraskayo:
@juergbi, you're right. `ensure_copy()` creates a copy of a file when the file is hardlinked, and the new file is assigned a new inode number by the undetlying filesystem. Since this PR passes through the underlying filesystem's inode numbers, this new inode number would be reflected to callers of stat(2).

Without this PR, libfuse's internally-generated inode was consistent for each unique path and as long as the file wasn't unlinked. However, it also meant that different hardlinks of the same file had different inode numbers, which is also somewhat inconsistent, but in a different way.

I honestly don't have a better idea on how to avoid both inconsistencies, or even only the former inconsistency, while also providing a consistent inode number in readdir(2).

Edit: now that I think about it, one can't actually avoid both inconsistencies by definition. It would require returning the same inode number for multiple files which aren't hardlinked in some cases. This means inode numbers wouldn't be unique anymore, which is a much worse option.

nanonyme:
Related to https://github.com/apache/buildstream/issues/324 "In some situations this can cause issues. E.g., if an unintentionally hardlinked file is modified in place. The SaferHardlinks FUSE layer protects against this during integration. However, the information is also lost when creating images for deployment."

PR #1570, opened 2022-01-18
Fix `MappingNode._composite()`

gtristan:
This branch fixes `Node._composite()` so that mutable data in the *source* nodes are no longer given away to the composite *target*, leaving the *source* vulnerable to later mutations when the *target* is mutated. Instead clones of source values are provided when they are composited onto a target node.

### DIFF
--- a/src/buildstream/node.pyx
+++ b/src/buildstream/node.pyx
@@ -446,7 +446,7 @@ cdef class ScalarNode(Node):
                                        self.get_provenance(),
                                        target_value.get_provenance()))
 
-        target.value[key] = self
+        target.value[key] = self.clone()
 
     cdef bint _is_composite_list(self) except *:
         return False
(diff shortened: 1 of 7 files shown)


BenjaminSchubert (approved):

nanonyme:
Is this really sufficient case to trigger the bug? Don't you need to have multiple elements, some of which are not parallel and some which are? I think @abderrahim was at least saying the issue is some variable is pre-expanded that shouldn't be.

gtristan:
> Is this really sufficient case to trigger the bug? Don't you need to have multiple elements, some of which are not parallel and some which are? I think @abderrahim was at least saying the issue is some variable is pre-expanded that shouldn't be.

Not yet, no.

gtristan:
> > Is this really sufficient case to trigger the bug? Don't you need to have multiple elements, some of which are not parallel and some which are? I think @abderrahim was at least saying the issue is some variable is pre-expanded that shouldn't be.
> 
> Not yet, no.

This has now been updated to properly trigger the bug.

gtristan:
And this merge request has now been updated to include both the test case and the fix for the underlying composition issue.

@BenjaminSchubert, can you take a look at this ?

BenjaminSchubert (approved):

gtristan:
We’ve got some kind of github bug, the CI actions appear to be randomly getting cancelled :(

PR #1573, opened 2022-01-20
element.py: fix plugin defaults being modified by instances

### LINKED ISSUE #1360: `notparallel` variable is broken
[See original issue on GitLab](https://gitlab.com/BuildStream/buildstream/-/issues/1360)
In GitLab by [[Gitlab user @willsalmon]](https://gitlab.com/willsalmon) on Jul 10, 2020, 10:39

## Summary

The `notparallel` variable dose not seem to force `MAKEFLAGS: -j1` in the environment

This means that projects that have bugs associated with `make -j2` can not be built, eg slang.

Please see the failure: https://gitlab.com/celduin/bsps/bst-boardsupport/-/jobs/625698066
The variable being incorrectly set on line 21 in the log: https://gitlab.com/celduin/bsps/bst-boardsupport/-/jobs/625698066/artifacts/file/artifact/freedesktop-sdk/components-slang/1ad3d1df-build.3146.log
The element https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/blob/master/elements/components/slang.bst showing the:
```
variables:
  builddir: ''
  notparallel: true
```

The problem is further compounded by the fact that bst will not let you set `MAKEFLAGS` in the environment section as it is `protected`

abderrahim:
Closes #1360

### DIFF
--- a/src/buildstream/element.py
+++ b/src/buildstream/element.py
@@ -2946,7 +2946,7 @@ class Element(Plugin):
         else:
             environment = project.base_environment.clone()
 
-        default_env._composite(environment)
+        default_env.clone()._composite(environment)
         element_env._composite(environment)
         environment._assert_fully_composited()
 


gtristan (commented):
I think we established that `MappingNode._composite()` is buggy inasmuch as compositing a source node onto another node somehow mutates the source node.

We need to fix that rather than work around it here, who knows what other side effects we might experience from this broken composition function otherwise.

EDIT: Noticed Ben’s comment on the issue and replied there.

gtristan:
Closing this in favor of #1570 which fixes the issues.

PR #1575, opened 2022-01-24
Heal from corruption where cache_size is empty file

### LINKED ISSUE #1540: [bst-1] "Error instantiating artifact cache: Size '' parsed from '/home/fedora/.cache/buildstream/artifacts/cache_size' was not an integer"
I had to shut down server while it was in the middle of running a bst (1.6.3) command. On restart, bst fails with

"Error instantiating artifact cache: Size '' parsed from '/home/fedora/.cache/buildstream/artifacts/cache_size' was not an integer"

I daresay the cache is somehow corrupted, but it should not be possible to get into a state where bst can't recover

nanonyme:
Related to #1540

### DIFF
--- a/buildstream/_artifactcache/artifactcache.py
+++ b/buildstream/_artifactcache/artifactcache.py
@@ -448,6 +448,10 @@ class ArtifactCache():
                 self._cache_size = stored_size
             else:
                 self.compute_cache_size()
+                # Computing cache doesn't actually write the value.
+                # Write cache size explicitly here since otherwise
+                # in some cases it's not stored on disk.
+                self.set_cache_size(self._cache_size)
 
         return self._cache_size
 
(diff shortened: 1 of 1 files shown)


juergbi (approved):

nanonyme:
Results in output like
```
[--:--:--][][] WARNING Failure resolving cache size

    Size '' parsed from '/var/home/nanonyme/.cache/buildstream/artifacts/cache_size' was not an integer
```
after which BuildStream continues as expected.

nanonyme:
I converted into draft since I realized it emits the warning on every run.

nanonyme:
@juergbi please review later again there's a reasonably major change in the flow i added to actually fix the cache.

gtristan (approved):
This looks safe to me.

Instead of bailing out on an invalid read of the cache size, we will end up recalculating the cache size and issuing a warning message.

PR #1576, opened 2022-01-25
_sandboxbwrap.py: try to support architectures that don't match host

abderrahim:
This can be used by setting up the host to run binaries from different
architectures via binfmt and qemu-user.

### DIFF
--- a/buildstream/sandbox/_sandboxbwrap.py
+++ b/buildstream/sandbox/_sandboxbwrap.py
@@ -49,6 +49,14 @@ class SandboxBwrap(Sandbox):
         '/dev/zero'
     ]
 
+    ARCHITECTURES = {
+        'amd64': 'x86_64',
+        'arm64': 'aarch64',
+        'i386': 'i686',
+        'armhf': 'armv7l',
+        'ppc64el': 'ppc64le',
+    }
+
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.user_ns_available = kwargs['user_ns_available']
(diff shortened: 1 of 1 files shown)


buildstream/sandbox/_sandboxbwrap.py
```diff
@@ -58,11 +66,25 @@ def __init__(self, *args, **kwargs):
...
         config = self._get_config()
 
-        # We can't do builds for another host or architecture except 32 bit on 64 bit
+        # We can't do builds for another host OS
         if config.build_os != host_os:
             raise SandboxError("Configured and host OS don't match.")
 
         if config.build_arch != host_arch:
+            try:
+                archtest = utils.get_host_tool('arch-test')
```
nanonyme, line 75: My only concern with this is availability of arch-test. I'm sort of wondering what would be the problem of just removing all this sanitation and only keeping linux32 parts. Build will fail anyway if CPU arches don't match. Why does BuildStream need to check? It's quite complex to do so.

nanonyme: Maybe this is fine. We can in our build images replace this with shim as well if desired

gtristan:
I was first concerned with the additional optional dependency appearing like an API surface of sorts, i.e. "If you run BuildStream in an environment with a binary called `arch-test`, then it will be consulted for yada yada", but given that we already use `linux32` this seems to be less of a concern.

That said, by applying this patch, are we setting up our users with unrealistic expectations when transitioning to BuildStream 2 ?

Native building of foreign binary formats is not supported by BuildStream as far as I'm aware, and hopefully, eventually remote execution clusters could be used for that kind of thing.

If we move ahead with this, we should not have the expectation that `buildbox-run` is going to behave in the same way.

abderrahim:
I'm going to propose a similar thing for buildbox-run (possibly a little more sophisticated, by looking at what other container tools like flatpak and docker/podman do).

That said, it is easier with bst2: We can use remote execution, by running a buildbox docker image of the architecture we want to build for. There is also a workaround we can use with bst2: we can build buildbox-run for the non-native architecture. It'll then think it is running on the correct architecture.  We can't do that with bst1 since it means compiling python for the non-native architecture, which will make bst very slow.

gtristan:
Alright, I've spoken with @juergbi about this and while it is a bit uncomfortable to do this in bst-1 *before* addressing this in master, it is agreed that it will be desirable to have a solution for this in master & buildbox-run.

So let's go ahead with this and roll out a new `1.6.4`

PR #1583, opened 2022-02-04
.github: Explicitly set job timeouts to 4 hours

gtristan:
Given that the runners have a tendency to be fairly crappy, what
takes about 20min to run on your typical laptop ends up taking over
an hour in CI.

Lets not take any chances and bump the timeout to 4 hours across the board.


gtristan:
Closing this as it's a failed attempt.

The only solution to this problem that I can see is to have CI runners that aren't rubbish.

PR #1586, opened 2022-02-09
Regenerate requirements using python 3.6

abderrahim:


abderrahim:
Sure enough, this breaks for new versions of python.

abderrahim:
I don't know what to make of this: generating the requirements using 3.6 or 3.7 makes newer versions fail. Generating with anything but 3.6 makes 3.6 fail.

nanonyme:
I think the right thing to do is to have different versions for different versions of Python. I have no idea idea how to generate that but there's py_version in requirements file format which can be used to declare that version x is for Python older than some version and version y is for Python newer than the version.

PR #1587, opened 2022-02-09
Backport ostree changes from bst-plugins-experimental

nanonyme:
This drops dependency on PyGObject

Origin https://gitlab.com/BuildStream/bst-plugins-experimental/-/blob/master/src/bst_plugins_experimental/sources/ostree.py


gtristan (approved):
Thanks for the backport ! This will be good to include in the next stable...

PR #1590, opened 2022-02-11
Fix coverage

abderrahim:


tests/sourcecache/staging.py
```diff
@@ -160,7 +160,7 @@ def test_staged_source_build(tmpdir, datafiles, cli):
     files = []
     for _, _, filename in os.walk(source_dir):
         files.extend(filename)
-    assert files == []
+    assert not files
```
nanonyme: Consider `assert not files, files` so we get files into output.

tox.ini
```diff
@@ -2,7 +2,7 @@
 # Tox global configuration
 #
 [tox]
-envlist = py37,py{38,39,310}-nocover
+envlist = py37-nocover,py{38,39,310}
```
juergbi: Based on the Debian 10 CI run of this branch and https://github.com/apache/buildstream/pull/1420#issuecomment-773094285, coverage should also work with Python 3.7. Have you encountered issues with coverage in Python 3.7?

juergbi (changes requested):
`.github/common.env` needs to be updated to actually use the tox environments with coverage in CI.

Other than that, this looks good to me, assuming CI with coverage enabled will pass.

tox.ini
abderrahim: I don't remember, I'll try to change `common.env` and see

juergbi:
This was fixed by #1817.

Issue #1588, opened 2022-02-11
1.6: CI failing due to Python 3.6


sam-thursfield-nutanix:
As of 11-02-2022, CI pipelines are failing for distros with Python 3.6 available:

  * Ubuntu 18.04: https://github.com/apache/buildstream/runs/5120980091?check_suite_focus=true
  * CentOS 7.7: https://github.com/apache/buildstream/runs/5120980121?check_suite_focus=true
 
The issue is: BuildStream Git tree has a single, frozen set of requirements which are used by all Python versions.

This was updated in https://github.com/apache/buildstream/pull/1564/ - which was accidentally merged despite the CI failures - breaking the branch.

The root cause is: Python 3.6 is old and some deps have dropped support already, such that there is no single frozen requirements.txt that can work with Python 3.6, 3.7, 3.8, 3.9 and 3.10 at the same time. (Proven here: https://github.com/apache/buildstream/pull/1586)

This doesn't affect *users* as `pip3 install buildstream` does not honour the frozen requirements.txt anyway. It only affects the test suite.

Also related: https://github.com/apache/buildstream/pull/1562

Options going forwards:

  * Drop Python 3.6 support - this makes sense for 'master', but we cannot do a 1.6.4. "bugfix" release that drops Py 3.6 support
  * Specify a different  frozen set of requirements for each Py version - see https://github.com/apache/buildstream/pull/1586#issuecomment-1034193819

juergbi:
I just noticed that besides being EOL upstream, Python 3.6 is no longer supported in RHEL 7 either. Python 3.8 is supported in RHEL 7. See https://access.redhat.com/support/policy/updates/rhscl-rhel7. Python 3.8 is also available in Ubuntu 18.04 (universe repository).

With Python 3.6 effectively being unmaintained upstream and downstream and no longer supported by a growing number of Python packages, I consider it acceptable to drop official Python 3.6 support even in the 1.6.x branch. With this I don't mean that we start using Python 3.7-only constructs in the 1.6.x branch, the branch would be expected to continue working with Python 3.6. However, I'd be in favor of dropping Python 3.6 from CI, which eliminates the issue with frozen `requirements.txt`. If we accidentally break Python 3.6 support, we should still accept (reasonable) patches to fix it.

nanonyme:
[...]

PR #1591, opened 2022-02-12
Stop running CI for distro versions with Python 3.6

nanonyme:
This does not drop Python 3.6 support and fixes for Python 3.6 compatibility are accepted.
Part of https://github.com/apache/buildstream/issues/1588

### DIFF
--- a/.github/compose/ci.docker-compose.yml
+++ b/.github/compose/ci.docker-compose.yml
@@ -34,14 +34,6 @@ services:
     <<: *tests-template
     image: registry.gitlab.com/buildstream/buildstream-docker-images/testsuite-debian:10-${CI_IMAGE_VERSION:-latest}
 
-  ubuntu-18.04:
-    <<: *tests-template
-    image: registry.gitlab.com/buildstream/buildstream-docker-images/testsuite-ubuntu:18.04-${CI_IMAGE_VERSION:-latest}
-
-  centos-7.7.1908:
-    <<: *tests-template
-    image: registry.gitlab.com/buildstream/buildstream-docker-images/testsuite-centos:7.7.1908-${CI_IMAGE_VERSION:-latest}
-
   docs:
     <<: *tests-template
     command: tox -e docs
--- a/.github/run-ci.sh
+++ b/.github/run-ci.sh
@@ -69,8 +69,6 @@ if [ -z "${test_names}" ]; then
     runTest "debian-10"
     runTest "fedora-34"
     runTest "fedora-35"
-    runTest "ubuntu-18.04"
-    runTest "centos-7.7.1908"
 else
     for test_name in "${test_names}"; do
 	runTest "${test_name}"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,8 +28,6 @@ jobs:
           - debian-10
           - fedora-34
           - fedora-35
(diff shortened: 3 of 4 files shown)


juergbi (approved):

.github/compose/ci.docker-compose.yml
```diff
@@ -34,14 +34,6 @@ services:
     <<: *tests-template
     image: registry.gitlab.com/buildstream/buildstream-docker-images/testsuite-debian:10-${CI_IMAGE_VERSION:-latest}
 
-  ubuntu-18.04:
-    <<: *tests-template
-    image: registry.gitlab.com/buildstream/buildstream-docker-images/testsuite-ubuntu:18.04-${CI_IMAGE_VERSION:-latest}
-
```
jjardon, line 40: Would not be better to still check bst-1 runs ok in 18.04 with python 3.7? Same for centos-7.7? (totally ok with remove this in master but this is the bst-1 stable branch) CC: @ssssam

nanonyme: @jjardon on CentOS 7.7 that would be 3.8 as that is the stack provided, not 3.7. That could in theory be done but requires significant amount of development work in the images and possibly in CI. Is there real need to keep supporting Ubuntu 18.04? Even flatpak-builder dropped support for that.

juergbi: The most important aspect, in my opinion, is that we keep testing Python 3.7 in CI. We can't but also shouldn't have to test all distro/Python combination. I think it's preferable to test distros with their default Python 3 version as that's more likely what people actually use.

Python 3.7 is already covered by Debian 10, which is expected to get LTS support until June 2024, which is after the Python 3.7 EOL date (June 2023). While there is a Python 3.7 package for Ubuntu 18.04, it's not the default version and it's in the 'universe' archive, which means that it's not guaranteed to be supported by Canonical, as far as I can tell. We now have a Ubuntu 20.04 (Python 3.8) Docker image for master, which we could also use for 1.x.

CentOS 7.7 (1908) is no longer supported. We could upgrade to the still supported CentOS 7.9 (2009) and install Python 3.8 from the software collection. However, I'd like to drop CentOS 7 from the Docker images for BuildStream master to lower maintenance cost and image build time. This means that we'd have to maintain a separate branch of bst-docker-images for 1.x. In my opinion, it would make more sense to upgrade to CentOS Stream 8 with Python 3.8 or 3.9, which would be useful for both 1.x and master.

jjardon: > The most important aspect, in my opinion, is that we keep testing Python 3.7 in CI

Yeah agree with that, and as that's covered with Debian 10 I think I'm ok with removing ubuntu-18.04

For CentOS, taking in account current usage of bst-1, I think is still worth to try to keep CentOS 7.x tests running, but I'd like to hear @ssssam opinions here

juergbi: @ssssam mentioned that they will be using CentOS 7 but will anyway upgrade to a more recent version of Python, which means that the current outdated and broken CentOS 7 test is not useful to them either. I suggest we go ahead and merge this now but try to re-add a new CentOS 7 test (with Python 3.8) in a follow-up.

jjardon: wfm!

PR #1592, opened 2022-02-15
update requirements (and fix new linter errors)

abderrahim:


src/buildstream/element.py
```diff
@@ -866,7 +866,7 @@ def _dependencies(self, scope, *, recurse=True, visited=None):
         # containing element that have been visited for the `_Scope.BUILD` case
         # and the second one relating to the `_Scope.RUN` case.
         if not recurse:
-            result: Set[Element] = set()
+            result: Set["Element"] = set()
```
nanonyme, line 817: This looks odd. Why was the old one not allowed?

abderrahim: The error I got is 
```
E0601: Using variable 'Element' before assignment (used-before-assignment)
```

This also means that this line is probably not covered by our tests.

This should be possible now that we require python 3.7 by using `from __future__ import annotations` but I'd rather not do it since it needs the future import even in 3.10.

gtristan (approved):
Let's go ahead and rebase and merge this

PR #1594, opened 2022-02-21
tests: Test that user specified plugins shadow core plugins

### LINKED ISSUE #1555: Ensure core plugins are shadowed by explicitly loaded plugins
At this time the core BuildStream codebase has no way of adding new core plugins as plugin names share the same namespace.

It was decided in [the january 5th meeting](https://lists.apache.org/thread/d7fgbdnfx62b4v1mz8g8qq56ohm3kjmz) that the most elegant way to address this is to simply ensure that explicitly loaded plugins *shadow* plugin names from the core - if the core ever exposes a new plugin which collides with a user's plugin name - then it will not break the project, and the user can readjust the project if they want to leverage the new plugin.

NOTE: It may be that the codebase currently behaves this way, if that is the case, we at least need to add a regression test to ensure this correct behavior.

NOTE2: This also raises the question about how to handle name clashing when loading multiple plugins from different origins, for this I have opened #1556 which we can easily implement at any time (no need to block 2.0 for this enhancement).

gtristan:
Fixes #1555

### DIFF
--- a/tests/plugins/shadow.py
+++ b/tests/plugins/shadow.py
@@ -1,0 +1,47 @@
+# Pylint doesn't play well with fixtures and dependency injection from pytest
+# pylint: disable=redefined-outer-name
+
+#
+# This test case ensures that user provided plugins appropriately shadow
+# core defined plugins, which is a behavior that is required in order to
+# make it safe for BuildStream to add more plugins in the future without
+# stomping on plugin namespace.
+#
+
+import os
+import pytest
+
+from buildstream.testing import cli  # pylint: disable=unused-import
+from buildstream import _yaml
+
+
+DATA_DIR = os.path.join(os.path.dirname(os.path.realpath(__file__)), "shadow")
+
+
+def update_project(project_path, updated_configuration):
+    project_conf_path = os.path.join(project_path, "project.conf")
+    project_conf = _yaml.roundtrip_load(project_conf_path)
+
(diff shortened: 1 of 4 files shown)


gtristan:
Ran tests locally and they pass, current github runners are not working properly, timing out after only 30min...

nanonyme:
@gtristan we seriously need to debug what's wrong. The runners work just fine with bst-1 branch but not master.

nanonyme:
And it really looks like bst/entire build hangs during the tests, it's not just that there's a timeout. Output simply ends at 58% tests done.

gtristan:
> @gtristan we seriously need to debug what's wrong. The runners work just fine with bst-1 branch but not master.

The runners are timing out after 30 minutes afaics.

While bst1 takes only 15 or 20 minutes, bst2 takes around 45 minutes even locally on my laptop.

It seems the tests which involve spinning up a dummy artifact remote cache locally take a ridiculous amount of time especially (up to 45 seconds for a single test), while we *should* be able to improve the timing of the tests, I think it’s not really acceptable to live with cheap runners that simply bail out after only 30 minutes.

nanonyme:
I removed hardcoded 30 minute timeout and it seems that it varies when it fails https://github.com/apache/buildstream/pull/1593. But what is the alternative? They either need to get fixed or project needs some other CI infra than what Apache can provide

PR #1595, opened 2022-02-22
Remove SandboxFlags from public API surface

### LINKED ISSUE #265: SandboxFlags API allows too much control, remove some
[See original issue on GitLab](https://gitlab.com/BuildStream/buildstream/-/issues/265)
In GitLab by [[Gitlab user @tristanvb]](https://gitlab.com/tristanvb) on Feb 22, 2018, 12:24

Currently the `SandboxFlags` argument for sandboxes allows plugins too much control on how to use a sandbox.

With the exception of `SandboxFlags.ROOT_READ_ONLY`, the remaining flags are only for internal use, so that we can control the behavior of a sandbox in a `bst shell` environment.

As a result, we should remove the remaining options from here while retaining `ROOT_READ_ONLY`, technically this is an API break; but it can only break for plugins who are doing something illegal in the first place.

gtristan:
The SandboxFlags expose internal details that are not suitable for
consumption by plugin authors.

Change summary:

  * sandbox/sandbox.py: Remove SandboxFlags as argument in public API
    surfaces, and expose a boolean `root_read_only` argument instead,
    while providing an internal `Sandbox._run_with_flags()` API for
    internal use.

  * __init__.py, sandbox/__init__.py:
    Updated to remove SandboxFlags from public API

  * buildelement.py, scriptelement.py, plugins/elements/autotools.py,
    plugins/elements/compose.py, plugins/elements/manual.py:
    Stop using SandboxFlags, use `root_read_only` parameter instead.

  * element.py: Adapt to API changes, use internal `_run_with_flags()` only
    for special cases such as running shells.

  * sandbox/_sandboxbuildboxrun.py, sandbox/_sandboxdummy.py,
    sandbox/_sandboxreapi.py: Adapt to API changes

Fixes #265

### DIFF
--- a/src/buildstream/__init__.py
+++ b/src/buildstream/__init__.py
@@ -28,7 +28,7 @@ if "_BST_COMPLETION" not in os.environ:
     del get_versions
 
     from .utils import UtilError, ProgramNotFoundError
-    from .sandbox import Sandbox, SandboxFlags, SandboxCommandError
+    from .sandbox import Sandbox, SandboxCommandError
     from .storage import Directory
     from .types import CoreWarnings, OverlapAction
     from .node import MappingNode, Node, ProvenanceInformation, ScalarNode, SequenceNode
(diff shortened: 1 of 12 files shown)


gtristan:
Ran CI locally, including linting, format checking, docs and integration tests.

Bypassing CI again while CI remains broken.

PR #1597, opened 2022-02-22
element.py: Include weak cache key in cache key calculation.

### LINKED ISSUE #1270: bst checkout fails to find the correct artifact in non strict mode
[See original issue on GitLab](https://gitlab.com/BuildStream/buildstream/-/issues/1270)
In GitLab by [[Gitlab user @abderrahimk]](https://gitlab.com/abderrahimk) on Mar 23, 2020, 06:12

## Steps to reproduce

Not sure it is reproducible, but here is what happens with me. I'm using gnome-build-meta commit 6f538b0529b288b0af257906f0f2b11a34a8e5fb, and have the following in my config.

```
projects:
  freedesktop-sdk:
    strict: False
  gnome:
    strict: False
```

```
> bst build vm/image.bst
> bst checkout vm/image.bst # => error
> bst --strict build vm/image.bst
> bst checkout vm/image.bst # => error
> bst --strict checkout vm/image.bst # => works
```

## Relevant logs and/or screenshots

[bst-show.log](/uploads/c3b14a0f32ec81498418508dedbd6ad3/bst-show.log)
[bst--strict-show.log](/uploads/eadff80450e4e2ee6377aab2e3c98a04/bst--strict-show.log)

## Other relevant information

* BuildStream version affected: /milestone %BuildStream_v1.4

----

gtristan:
This should eliminate a class of issues related to strict/non-strict
discrepency.

This also updates the cache key test.

Fixes #1270


gtristan:
This PR replaces #1421 as discussed [here](https://github.com/apache/buildstream/pull/1421#issuecomment-773094780), it is insufficient to consider only the `Element.BST_STRICT_REBUILD` flag, since dependencies can themselves be considered strict individually.

juergbi (approved):
The change makes sense to me. That said, it's been a while since deeply thinking about cache key calculation.

A regression test for #1270 would be great but I don't know how easy it would be to construct a test case for this.

PR #1604, opened 2022-02-24
make: Make it easier to override parameters to make

nanonyme:
Backport of https://gitlab.com/BuildStream/bst-plugins-experimental/-/merge_requests/121

### DIFF
--- a/buildstream/plugins/elements/make.yaml
+++ b/buildstream/plugins/elements/make.yaml
@@ -1,8 +1,14 @@
 # make default configurations
 
 variables:
-  make: make PREFIX="%{prefix}"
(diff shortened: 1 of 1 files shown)


gtristan (commented):
This looks like it will break cache keys, which we don't do in minor point releases in `bst-1`, and seeing as `bst-1` is essentially in maintenance mode, I'm doubtful we'll ever want to see a `1.8.x` feature adding release of `bst-1`.

nanonyme:
Why would this break cache keys? It doesn't affect the string outputs.

gtristan:
> Why would this break cache keys? It doesn't affect the string outputs.

Then it wouldn’t break cache keys :)

PR #1607, opened 2022-03-03
source.py: Split up staging functions into separate codepaths.

gtristan:
Currently we pass a file path string or Directory object to these methods
depending on whether BST_STAGE_VIRTUAL_DIRECTORY is set, which is not very
pretty for plugin developers especially if they are using type annotations
and need to handle `Union[str, Directory]` in these methods.

Instead, created Source.stage_directory() and Source.init_workspace_directory()
to handle the Directory variants of these methods separately.

Also updated `local` and `workspace` plugins to use the new methods.

### DIFF
--- a/src/buildstream/plugins/sources/local.py
+++ b/src/buildstream/plugins/sources/local.py
@@ -90,7 +90,7 @@ class LocalSource(Source):
         # Nothing to do here for a local source
         pass  # pragma: nocover
 
-    def stage(self, directory):
+    def stage_directory(self, directory):
         #
         # We've already prepared the CAS while resolving the cache key which
         # will happen before staging.
@@ -102,7 +102,7 @@ class LocalSource(Source):
         with self._cache_directory(digest=self.__digest) as cached_directory:
             directory.import_files(cached_directory)
 
-    def init_workspace(self, directory):
+    def init_workspace_directory(self, directory):
         #
         # FIXME: We should be able to stage the workspace from the content
         #        cached in CAS instead of reimporting from the filesystem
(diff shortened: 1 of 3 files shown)


abderrahim:
I have been thinking of this too (in the context of #1317). What I was thinking to do is to always pass a Directory to `stage()` (I didn't look into `init_workspace()`, so I can't say anything). The plugins would need to change indeed, but I think it's not that hard.

```
with self.tempdir() as directory:
    ...
    vdir.import_files(directory)
```

gtristan:
> I have been thinking of this too (in the context of #1317). What I was thinking to do is to always pass a Directory to `stage()` (I didn't look into `init_workspace()`, so I can't say anything). The plugins would need to change indeed, but I think it's not that hard.
> 
> ```
> with self.tempdir() as directory:
>     ...
>     vdir.import_files(directory)
> ```

It is desirable for plugins to stage directly into cas using the Directory API and avoid an additional disk copy, so I can see the desire to make the API more rigid and have more direct messaging to plugin authors that they ultimately have to get the data into a cas Directory, however I don't think it's a realistic goal for the majority of source plugins, and so I would much prefer to favor a simpler and more convenient API for the majority of source plugins.

Regarding `init_workspace()`, it's main purpose is to prepare a special case staging ground for a workspace - plugins like `git` and `bzr` which tend to carry non-deterministic data in them because they carry project history (i.e. the `.git/` directory) will normally omit such non-deterministic data at normal `stage()` time, but will include the `.git/` directory for the convenience of a workspace.

In the end what we're really talking about is whether or not to provide a more natural/convenient API for most plugins rather then forcing them to do the `with self.tempdir()` trick themselves, I think that since the majority of plugins will have to do it, we aught to do it for them and offer a more convenient API.

gtristan:
There hasn't been any more input for a month, I still think this approach makes more sense than forcing source plugin authors to work with the `Directory` interface unconditionally, and this patch does improve the interfaces and is a breaking change which we should land before declaring API stability.

Let's merge this.

PR #1608, opened 2022-03-03
ostree: Remove mirror parameter

nanonyme:
It doesn't produce any value and the resultant repository will
not work correctly if ostree repo originally used collection id


nanonyme:
Corresponding MR for bst master in https://gitlab.com/BuildStream/bst-plugins-experimental/-/merge_requests/160

abderrahim (approved):
LGTM

PR #1609, opened 2022-03-04
Directory API refactoring

### LINKED ISSUE #1294: Review public `Directory` API
[See original issue on GitLab](https://gitlab.com/BuildStream/buildstream/-/issues/1294)
In GitLab by [[Gitlab user @juergbi]](https://gitlab.com/juergbi) on Apr 27, 2020, 07:58

We should make sure we're happy to support the public `Directory` API long term before releasing 2.0. Aspects to consider:
* Do we want to keep the `*paths: str` approach (variable argument list for individual path components)? It can be convenient to avoid the need for string joining/splitting and eliminates the issue of OS-specific path separators. However, it's different from prevalent filesystem APIs, which may confuse plugin developers. And there are also cases where the API is less convenient (e.g., using paths specified in `.bst` files). If we decide to switch to a traditional path-based API, I'd always use `/` as path separator to match REAPI and avoid issues building e.g. Linux artifacts from Windows (using remote execution or Docker).
* `import_files()`: Do we need the full complexity of the current API? `filter_callback`, `report_written`, `FileListResult`
* `import_single_file()`: Do we want to support specifying a destination path? Maybe rename the method to `import_file()`
* Are `list_modified_paths()` and `mark_unmodified()` the best approach for what `compose` needs?
* Is `set_deterministic_user()` needed in the public API?
* `Directory.stat()` returns a `os.stat_result`. Would it be better to define a custom, extensible data type instead? https://gitlab.com/BuildStream/buildstream/-/merge_requests/1878#note_331809335
* Get rid of the undocumented dictionary returned from the `CasBasedDirectory.walk()` API
  * This is used in the frontend in order to implement `bst artifact list-contents`
  * To avoid redundancy and duplication, it would be suitable to use the same objects which `Directory.stat()` returns in this `walk()` API, perhaps a custom object with a bit more context than `os.stat_result` would be beneficial here, as we can add more context about the file, such as symbolic link targets.
  * Doing this part should result in better updated documentation of what is going on in the `_frontend/cli.py` and `_frontend/widget.py` portions which deal with formatting this data for the user.
* Consider removing the `mark_unmodified()` API, as this sets the deterministic mtime only for the local filebased directory, but this is only relevant for builds and we only ever stage into CAS now
  * Note: deterministic mtime is now handled by *not* specifying any mtime in cas, and expecting a sandbox backend which stages files from CAS, like `buildbox` to use the deterministic mtime in place of any explicitly set mtime.

gtristan:
This branch includes a series of changes to cleanup the Directory API.

Fixes #1294

### DIFF
--- a/src/buildstream/storage/_casbaseddirectory.py
+++ b/src/buildstream/storage/_casbaseddirectory.py
@@ -31,13 +31,13 @@ import stat
 import tarfile as tarfilelib
 from contextlib import contextmanager
 from io import StringIO
+from typing import Optional
 from google.protobuf import timestamp_pb2
 
 from .. import utils
 from ..types import FastEnum
 from .._protos.build.bazel.remote.execution.v2 import remote_execution_pb2
 from .directory import Directory, DirectoryError, FileMode, FileStat
-from ._filebaseddirectory import FileBasedDirectory
 from ..utils import FileListResult, BST_ARBITRARY_TIMESTAMP
 
 
@@ -522,16 +522,23 @@ class CasBasedDirectory(Directory):
 
         result = FileListResult()
 
-        if isinstance(external_pathspec, FileBasedDirectory):
-            external_pathspec = external_pathspec._get_underlying_directory()
-
+        # See if we can get a source directory to copy from
+        source_directory: Optional[str] = None
         if isinstance(external_pathspec, str):
+            source_directory = external_pathspec
+        elif isinstance(external_pathspec, Directory):
+            try:
+                source_directory = external_pathspec._get_underlying_directory()
+            except DirectoryError:
+                pass
+
+        if source_directory:
             # Import files from local filesystem by first importing complete
             # directory into CAS (using buildbox-casd) and then importing its
             # content into this CasBasedDirectory using CAS-to-CAS import
             # to write the report, handle possible conflicts (if the target
             # directory is not empty) and apply the optional filter.
-            digest = self.cas_cache.import_directory(external_pathspec, properties=properties)
+            digest = self.cas_cache.import_directory(source_directory, properties=properties)
             external_pathspec = CasBasedDirectory(self.cas_cache, digest=digest)
 
         assert isinstance(external_pathspec, CasBasedDirectory)
--- a/src/buildstream/storage/_casbaseddirectory.py
+++ b/src/buildstream/storage/_casbaseddirectory.py
@@ -167,55 +167,11 @@ class CasBasedDirectory(Directory):
     #############################################################
     #              Implementation of Public API                 #
     #############################################################
-    def descend(self, *path: str, create: bool = False, follow_symlinks: bool = False) -> "CasBasedDirectory":
-        # Note: At the moment, creating a directory by descending does
-        # not update this object in the CAS cache. However, performing
-        # an import_files() into a subdirectory of any depth obtained by
-        # descending from this object *will* cause this directory to be
-        # updated and stored.
-        current_dir = self
-        paths = list(path)
 
-        for element in paths:
-            # Skip empty path segments
-            if not element:
-                continue
-
-            self.__validate_path_component(element)
-
-            entry = current_dir.__index.get(element)
-
-            if entry:
-                if entry.type == FileType.DIRECTORY:
-                    current_dir = entry.get_directory(current_dir)
-                elif follow_symlinks and entry.type == FileType.SYMLINK:
-                    assert entry.target is not None
-                    linklocation = entry.target
-                    newpaths = linklocation.split(os.path.sep)
-                    if os.path.isabs(linklocation):
-                        current_dir = current_dir.__find_root().descend(*newpaths, follow_symlinks=True)
-                    else:
-                        current_dir = current_dir.descend(*newpaths, follow_symlinks=True)
-                else:
-                    error = "Cannot descend into {}, which is a '{}' in the directory {}"
-                    raise DirectoryError(
-                        error.format(element, current_dir.__index[element].type, current_dir), reason="not-a-directory"
-                    )
-            else:
-                if element == ".":
-                    continue
-                if element == "..":
-                    if current_dir.__parent is not None:
-                        current_dir = current_dir.__parent
-                    # In POSIX /.. == / so just stay at the root dir
-                    continue
-                if create:
-                    current_dir = current_dir.__add_directory(element)
-                else:
-                    error = "'{}' not found in {}"
-                    raise DirectoryError(error.format(element, str(current_dir)), reason="directory-not-found")
-
-        return current_dir
+    def descend(self, path: str, *, create: bool = False, follow_symlinks: bool = False) -> "CasBasedDirectory":
+        self._validate_path(path)
+        paths = path.split("/")
+        return self.__descend(paths, create=create, follow_symlinks=follow_symlinks)
 
     def import_single_file(self, external_pathspec: str) -> FileListResult:
         result = FileListResult()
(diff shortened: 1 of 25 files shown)


abderrahim:
I find it a little hard to review, it would be nice to split into smaller PRs for review. The following I think are uncontroversial and can easily be merged in a first batch:

- storage: Rename VirtualDirectoryError -> DirectoryError
- storage/directory.py: Remove _mark_changed() internal API.
- storage/directory.py: Removed mark_unmodified/list_modified_paths
- storage/directory.py: Make set_deterministic_user() an internal function
- storage/directory.py: Added full mypy annotations to API surface
- storage/directory.py: Added declaration for abstract internal method _get_underlying_directory()
- storage: Raise only DirectoryError from Directory backends
- storage/directory.py: Remove "report_written" from Directory.import_files()
- storage/_casbaseddirectory.py: Stop importing FileBasedDirectory
- storage/directory.py: Make get_size() an internal method


I have a couple of comments on these:
- storage/directory.py: Remove `can_link` parameter from Directory.import_files()
  Even though it is currently unused, there is potential to have a similar option that uses  the `move_files` hint  (by passing `--capture-allow-file-moves` to `buildbox-casd`). Can be done later though.
- storage/directory.py: Make Directory.export_files() an internal function
  If we want to go for https://github.com/apache/buildstream/pull/1607#issuecomment-1059019265, we'll need to keep this public. Again, this can be made public later if needed (even if it's not at 2.0).

gtristan:
> I find it a little hard to review, it would be nice to split into smaller PRs for review.

I'd rather keep this in one PR as it's all related to the Directory API cleanup, splitting this up also generates extra considerations when updating moving parts, note that I already have https://gitlab.com/BuildStream/bst-plugins-experimental/-/merge_requests/166 lined up for this.

I had considered trying to reduce the number of commits (which would be quite difficult short of squashing it all into one commit) but I would like to keep them all in history, every commit here passes it's own tests and tells a full story of how things were changed here which is nice to keep.

[...]
> I have a couple of comments on these:
> 
>     * storage/directory.py: Remove `can_link` parameter from Directory.import_files()
>       Even though it is currently unused, there is potential to have a similar option that uses  the `move_files` hint  (by passing `--capture-allow-file-moves` to `buildbox-casd`). Can be done later though.

Sure I'll consider this unrelated, in any case it's highly unlikely that we would want to expose the abstract plugin API to such low level implementation details, given that the public API can never again be changed.

>     * storage/directory.py: Make Directory.export_files() an internal function
>       If we want to go for source.py: Split up staging functions into separate codepaths. #1607 (comment), 
>       we'll need to keep this public. Again, this can be made public later if needed (even if it's not at 2.0).

Why ? `Directory.export_files()` is only ever used by the core when *exporting* files in a virtual directory to the local filesystem, when e.g. checking out artifacts or sources, etc. `Directory.import_files()` of course needs to stay public.

src/buildstream/storage/directory.py
```diff
@@ -16,328 +17,499 @@
...
-from the backing store implementation.  Sandboxes are allowed to read
-from and write to the underlying storage, but all others must use this
-Directory class to access files and directories in the sandbox.
-
-See also: :ref:`sandboxing`.
-
+Directory - Interfacing with files
+==================================
+The Directory class is given to plugins by way of the :class:`.Sandbox`
+or in some other instances, and allows plugins to interface with files
```
juergbi: Word missing after 'or'?

src/buildstream/storage/directory.py
```diff
@@ -16,328 +17,499 @@
...
-    def __str__(self):
-        # https://github.com/PyCQA/pylint/issues/2062
-        return self.name.lower().replace("_", " ")  # pylint: disable=no-member
+    # _validate_path()
+    #
+    # Convenience function for backends to validate path input
+    #
+    def _validate_path(self, path: str) -> None:
+        if len(path) and path[0] == "/":
+            raise ValueError("Invalid path '{}'".format(path))
```
juergbi, line 509: This function considers empty paths valid. Do we really want this or is this necessary? Intuitively I'd expect `.` to be accepted for the current directory but the empty path to be rejected, which matches POSIX behavior.

src/buildstream/storage/directory.py
```diff
@@ -16,328 +17,499 @@
...
+separated with the ``/`` character regardless of the platform. Both ``.`` and ``..`` entries
+are permitted. Absolute paths are not permitted, as such it is illegal to specify a path
+with a leading ``/`` character.
+
+Directory objects represent a directory entry within the context of a *directory tree*,
+and the directory returned by
+:func:`Sandbox.get_virtual_directory() <buildstream.sandbox.Sandbox.get_virtual_directory>`
+is the root of the sandbox's *directory tree*. Attempts to escape the root of a *directory tree*
+using ``..`` entries will not result in an error, instead ``..`` entries which cross the
+root boundary will be evaluated as the root directory.
```
juergbi: Maybe mention that this matches POSIX behavior of filesystem roots?

src/buildstream/storage/directory.py
```diff
@@ -16,328 +17,499 @@
...
 
-    def descend(self, *paths: str, create: bool = False, follow_symlinks: bool = False):
+    def __iter__(self) -> Iterator[str]:
+        raise NotImplementedError()
+
+    ###################################################################
+    #                           Public API                            #
+    ###################################################################
+
+    def descend(self, path: str, *, create: bool = False, follow_symlinks: bool = False) -> "Directory":
```
juergbi: As this also accepts .. (protected against escaping), the name may be a bit off. descend is also not common terminology when working with directories, as far as I know. Could it make sense to rename it to open_directory()? This would be consistent with open_file() and would be the Directory equivalent of POSIX open(..., O_DIRECTORY).

src/buildstream/storage/directory.py
```diff
@@ -16,328 +17,499 @@
...
+
+        Raises:
+           DirectoryError: if any system error occurs.
         """
         raise NotImplementedError()
 
     # Convenience functions
-    def is_empty(self):
-        """ Return true if this directory has no files, subdirectories or links in it.
+    def is_empty(self) -> bool:
```
juergbi: The more Pythonic way would probably be to implement the inverse as `__bool__` (or `__len__` but that might not be needed).

src/buildstream/storage/_casbaseddirectory.py
```diff
@@ -114,136 +128,337 @@ def get_equivalency_properties(e: IndexEntry):
...
+                sio = StringIO(entry.target)
+                bio = BytesIO(sio.read().encode("utf8"))
+                tarfile.addfile(tarinfo, bio)
+            else:
+                raise DirectoryError("can not export file type {} to tar".format(entry.type))
+
+    def is_empty(self) -> bool:
+        return len(self.__index) == 0
+
+    def listdir(self) -> Iterator[str]:
```
juergbi: This is not declared in `Directory`. I'm wondering whether we should combine this and `list_relative_paths()` into a single method with an optional `recursive` (bool keyword-only) parameter. What do you think?

gtristan: oops listdir() is an artifact of something I've already backed out of the branch.

This is handled by way of the `Directory` being iterable instead, I'll remove this from wherever it was added.

src/buildstream/storage/directory.py
gtristan: I did `s/or/and/` here and improved the grammar of this phrase.

src/buildstream/storage/directory.py
gtristan: Ok let's do this.

src/buildstream/storage/directory.py
gtristan: Appended to this paragraph:

> This behavior matches POSIX behavior of filesystem root directories.

src/buildstream/storage/directory.py
gtristan: Using `__len__` here since that seems to make sense because the `Directory` already otherwise behaves as a collection of entries.

Also updating the docs note about iterating over entries to mention that directories behave like collections of entries, and that they are falsy when emtpy, etc.

This was tricky as I needed to track down various instances where we need to change `if directory` to `if directory is not None`

src/buildstream/storage/directory.py
gtristan: Turns out this is not particularly user friendly, not sure if we should do this.

Elements like the `import` element want to call `inputdir = inputdir.descend(self.source.strip(os.sep))`, so that if there is a "/" then the input is at the root.

Forcing the caller to specify "." here seems to introduce more special casing on the caller side than is desirable.

Let's keep it this way.

src/buildstream/storage/directory.py
gtristan: Done, let's call it `open_directory()` instead.

PR #1618, opened 2022-03-17
.github/run-ci.sh: remove extra quotes

abderrahim:
The quotes prevent passing multiple tests (or no test) on the command line

Fixes ec84e3f113e991c6824d3fa16c105cf2be0e7f50


gtristan:
I've merged the sandbox batching API changes in buildstream and bst-plugins-experimental, and updated the ref in buildstream's `tox.ini` so that tests should be passing again.

Rebasing this branch should fix the issues now.

PR #1627, opened 2022-03-22
tox.ini: add --namespace-packages to mypy

abderrahim:
It is needed by latest version of types-protobuf

Also update mypy to the latest version

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -157,9 +157,9 @@ deps =
 [testenv:mypy]
 skip_install = True
 commands =
-    mypy {posargs}
+    mypy --namespace-packages {posargs}
 deps =
-    mypy==0.910
+    mypy==0.940
     types-protobuf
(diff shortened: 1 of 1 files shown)


juergbi:
The change in types-protobuf was reverted: https://github.com/python/typeshed/pull/7522
mypy should work again with 3.19.15 even without `--namespace-packages`.

I don't know whether it still makes sense to add this option.

abderrahim:
> I don't know whether it still makes sense to add this option.
The issue mentioned that it would be the default in the next version of mypy.

What we could do instead is to depend on `types-protobuf != 3.19.14`, this way it should fix the current issue and keep working when 3.19.15 is released.

juergbi:
types-protobuf 3.19.15 has already been released on PyPI.

juergbi:
This was added as part of #1782.

PR #1632, opened 2022-03-27
tests: xfail integration tests that fail without buildbox-fuse

abderrahim:
Two of them are known (they are already marked as xfail when using buildbox-run-userchroot) but the third seems weird. It's probably a bug.

### DIFF
--- a/tests/integration/script.py
+++ b/tests/integration/script.py
@@ -6,7 +6,7 @@ import pytest
 
 from buildstream import _yaml
 from buildstream._testing import cli_integration as cli  # pylint: disable=unused-import
-from buildstream._testing._utils.site import HAVE_SANDBOX, BUILDBOX_RUN
+from buildstream._testing._utils.site import HAVE_SANDBOX, BUILDBOX_RUN, HAVE_BUILDBOX_FUSE
 
 
 pytestmark = pytest.mark.integration
@@ -95,6 +95,7 @@ def test_script_root(cli, datafiles):
 
 @pytest.mark.datafiles(DATA_DIR)
 @pytest.mark.skipif(not HAVE_SANDBOX, reason="Only available with a functioning sandbox")
+@pytest.mark.xfail(not HAVE_BUILDBOX_FUSE, reason="I don't know, seems like a bug")
 def test_script_no_root(cli, datafiles):
     project = str(datafiles)
     element_path = os.path.join(project, "elements")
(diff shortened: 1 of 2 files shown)


tests/integration/script.py
```diff
@@ -171,6 +172,7 @@ def test_script_layout(cli, datafiles):
     HAVE_SANDBOX == "buildbox-run" and BUILDBOX_RUN == "buildbox-run-userchroot",
     reason="Root directory not writable with userchroot",
 )
+@pytest.mark.xfail(not HAVE_BUILDBOX_FUSE, reason="Root directory not writable without buildbox-fuse")
```
nanonyme, line 175: Does it really make sense to xfail rather scenarios that cannot possibly work? Doesn't it just make everything slower?

gtristan: Normally we use `@pytest.mark.skipif` to conditionally skip tests which we cannot run because of lacking components in the environment.

However here it seems appropriate to conditionally `xfail` the test since this should be considered incorrect behavior of buildstream *regardless* of the absence of `buildbox-fuse`, so it's nice to have the tests complain about this.

nanonyme: Hmm, but one the other hand, this incorrect behaviour seems expected and this test is just documenting it.

juergbi: If I remember correctly, the non-writable root directory is only an expected failure with userchroot (which we already mark as xfail). If this is not working with bubblewrap without buildbox-fuse, we should at least investigate this before marking it as xfail.

abderrahim: Thinking about this, I think it's an expected failure without buildbox-fuse. I don't know much about userchroot, but I don't think it's related: if we have a the fuse layer to protect the underlying blobs from corruption we can allow their modification. If not we'll simply protect them by not allowing writing (of course, this will only protect from accidental modification and you need to run buildbox-casd as a different user to protect against malicious modification).

I think the relationship between buildbox-fuse and buildbox-run-userchroot is a correlation (if you're using userchroot you're likely not on linux so you can't use fuse) rather than a strict relation.

Did I miss something?

juergbi: Looking at the test case in more detail, this is not actually related to the writability of the root directory. It's about the possibility to modify an existing staged file (without removing it beforehand and without causing cache corruption). The userchroot xfail condition (and `reason`) makes sense for `test_script_root`, however, for the corruption tests, a different approach makes more sense:

Drop the xfail (even the existing userchroot one). Instead, expect the build of `element_name` to fail if buildbox-fuse is not used. That build should fail with a proper setup (different user for buildbox-casd). This is more precise than an xfail test marker as that way we can make sure that it fails at the right point (we don't want it to xfail if it can corrupt the canary). I haven't tested this, though.

PR #1634, opened 2022-03-29
.github/workflow: run tests using ubuntu 18.04

abderrahim:
Ubuntu 20.04 seems to have a kernel bug making tests that use buildbox-fuse
hang.


gtristan:
Ooooooooh let's watch this CI run :)

gtristan:
Very good !

It seems to have bailed out in the buildgrid test for some reason but this is a huge improvement !

PR #1643, opened 2022-05-04
requirements: Updated requirements with unrestricted grpcio

### LINKED ISSUE #1616: Depend on new version of grpcio
bst master now depends on `grpcio < 1.39` which is pretty old. In fact it is old to the point it is built with an ssl stack that is now broken. Running bst master I get the following error when initializing remotes (both freedesktop-sdk and gnome, both using certificates from Let's Encrypt)

```
E0313 17:15:25.577198275   11041 ssl_transport_security.cc:1468] Handshake failed with fatal error SSL_ERROR_SSL: error:1000007d:SSL routines:OPENSSL_internal:CERTIFICATE_VERIFY_FAILED.
```

We need to drop this versioned dependency, hopefully the newest versions work fine with python 3.10.

gtristan:
Fixes #1616


nanonyme:
@gtristan consider rebasing to pull the Fedora 36 change in.

gtristan:
It appears that this pipeline has *completely* passed on the first try, and my understanding is that #1616 is rather urgent.

It is highly possible that we will encounter spurious segfaults from newer grpcio, but these appear to be edge case and we're not seeing them right now with the new docker image, so let's merge this and fix #1616 for now and file a separate issue if the spurious segfaults come back with new grpcio.

requirements/requirements.txt
```diff
@@ -1,17 +1,15 @@
...
+grpcio==1.46.0
+Jinja2==3.1.2
 pluginbase==1.0.1
-protobuf==3.19.4
+protobuf==3.20.1
 psutil==5.9.0
 ruamel.yaml==0.17.21
 ruamel.yaml.clib==0.2.6
-setuptools==59.6.0
+setuptools==44.1.1
```
nanonyme, line 9: Whoa, I did not spot this. Could you document why setuptools is pinned now to really old version?

gtristan: Ummm, I have no idea, we are not pinning it.

This is a result of updating the requirements, so presumably one of the newer versions of *something* is requiring a restricted version of setuptools.

nanonyme: Too bad this pinning tooling is not printing out reasons for the pins. Pip tools can be made to do that.

abderrahim: This is likely due to the version of python that was used to freeze the requirements.

Issue #1644, opened 2022-05-06
Errors pushing/pulling sources to source cache servers


gtristan:
While trying to build fdsdk with bst2 using source cache servers: https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/merge_requests/8332

We find the following error while pushing sources to the source cache: https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/jobs/2412763685

```
Failure Summary
    cross-compilers/freedesktop-sdk-arm.bst:bootstrap/build/binutils-stage1.bst:
    [18:10:16][00:08:57][2d12d633][src-push:cross-compilers/freedesktop-sdk-arm.bst:bootstrap/build/binutils-stage1.bst] FAILURE FetchDirectory failed with status UNAVAILABLE: Socket closed
        Printing the last 80 lines from log file:
        /builds/freedesktop-sdk/freedesktop-sdk/cache/buildstream/logs/freedesktop-sdk/bootstrap-build-binutils-stage1/2d12d633-src-push.190.log
        ======================================================================
        [--:--:--] START   [2d12d633] cross-compilers/freedesktop-sdk-arm.bst:bootstrap/build/binutils-stage1.bst: Src-push
        [--:--:--] STATUS  cross-compilers/freedesktop-sdk-arm.bst:bootstrap/build/binutils-stage1.bst: Pushing data for source 2eef7b5f -> https://local-cas-server:1102/
        [--:--:--] STATUS  cross-compilers/freedesktop-sdk-arm.bst:bootstrap/build/binutils-stage1.bst: Pushing data for source 2eef7b5f -> https://cache.freedesktop-sdk.io:11002/
        [--:--:--] STATUS  cross-compilers/freedesktop-sdk-arm.bst:bootstrap/build/binutils-stage1.bst: Pushing source 2eef7b5f -> https://local-cas-server:1102/
        [00:08:57] FAILURE [2d12d633] cross-compilers/freedesktop-sdk-arm.bst:bootstrap/build/binutils-stage1.bst: FetchDirectory failed with status UNAVAILABLE: Socket closed
```

And also this error when pulling sources from the source cache: https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/jobs/2412763658

```
Failure Summary
    extensions/clinfo/khronos-headers.bst:
    [15:50:26][00:00:00][96f921df][   fetch:extensions/clinfo/khronos-headers.bst] FAILURE Failed to pull source d97540b8
[...]

PR #1645, opened 2022-05-06
Make source cache errors temporary

nanonyme:
Related to https://github.com/apache/buildstream/issues/1644

### DIFF
--- a/src/buildstream/_elementsourcescache.py
+++ b/src/buildstream/_elementsourcescache.py
@@ -103,7 +103,9 @@ class ElementSourcesCache(AssetCache):
 
         if errors and not source_digest:
             raise SourceCacheError(
-                "Failed to pull source {}".format(display_key), detail="\n".join(str(e) for e in errors)
+                "Failed to pull source {}".format(display_key),
+                detail="\n".join(str(e) for e in errors),
+                temporary=True,
             )
 
         # If we don't have a source proto, we can't pull source files
@@ -230,7 +232,7 @@ class ElementSourcesCache(AssetCache):
         except grpc.RpcError as e:
             if e.code() != grpc.StatusCode.RESOURCE_EXHAUSTED:
                 raise SourceCacheError(
-                    "Failed to push source blobs with status {}: {}".format(e.code().name, e.details())
+                    "Failed to push source blobs with status {}: {}".format(e.code().name, e.details()), temporary=True
                 )
             return False
 
@@ -261,7 +263,7 @@ class ElementSourcesCache(AssetCache):
         except grpc.RpcError as e:
             if e.code() != grpc.StatusCode.NOT_FOUND:
                 raise SourceCacheError(
-                    "Error checking source cache with status {}: {}".format(e.code().name, e.details())
+                    "Error checking source cache with status {}: {}".format(e.code().name, e.details()), temporary=True
                 )
 
         referenced_directories = [source_proto.files]
@@ -273,7 +275,9 @@ class ElementSourcesCache(AssetCache):
                 references_directories=referenced_directories,
             )
         except grpc.RpcError as e:
-            raise SourceCacheError("Failed to push source with status {}: {}".format(e.code().name, e.details()))
+            raise SourceCacheError(
+                "Failed to push source with status {}: {}".format(e.code().name, e.details()), temporary=True
+            )
 
         return True
 
@@ -308,7 +312,9 @@ class ElementSourcesCache(AssetCache):
             self.cas._fetch_directory(remote, source.files)
         except grpc.RpcError as e:
             if e.code() != grpc.StatusCode.NOT_FOUND:
-                raise SourceCacheError("Failed to pull source with status {}: {}".format(e.code().name, e.details()))
+                raise SourceCacheError(
+                    "Failed to pull source with status {}: {}".format(e.code().name, e.details()), temporary=True
+                )
             return False
 
         return True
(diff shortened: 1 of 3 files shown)


src/buildstream/_elementsourcescache.py
```diff
@@ -195,7 +197,7 @@ def _get_source(self, ref):
                 source_proto.ParseFromString(f.read())
                 return source_proto
         except FileNotFoundError as e:
-            raise SourceCacheError("Attempted to access unavailable source: {}".format(e)) from e
+            raise SourceCacheError("Attempted to access unavailable source: {}".format(e), temporary=True) from e
```
gtristan: If I understand correctly, here we are about to push sources (this is called from `self._push_source()`) and we call `_get_source()` to get the proto from local CAS.

I suppose this error could occur if for some reason the local source has been purged/expired from local CAS before pushing.

I don't think this error can possibly recover with retries, so `temporary` is probably wrong in this case.

src/buildstream/_sourcecache.py
```diff
@@ -128,7 +128,7 @@ def pull(self, source):
                     )
                     continue
             except CASError as e:
-                raise SourceCacheError("Failed to pull source {}: {}".format(display_key, e)) from e
+                raise SourceCacheError("Failed to pull source {}: {}".format(display_key, e), temporary=True) from e
```
gtristan, line 131: This whole `try` block confuses me.

The only functions which might potentially raise an error are `remote.init()` and `self._pull_source()`.

Neither of these appear to potentially raise `CASError`

* `BaseRemote.init()` clearly cannot raise `CASError`
* `self._pull_source` descends into the following two paths
  * `self._store_source()` (indirectly `self._store_proto()`) will just write the proto directly into the CAS directories using `utils.save_file_atomic()`
  * `AssetRemote.fetch_directory()` will do some grpc-foo and potentially raise `AssetCacheError`

I think that the `CASError` handling might be leftovers left behind from previous refactors, it would probably make more sense to:
* Remove the nonsense `CASError` handling
* Handle the `AssetCacheError` in a _more isolated `try` block_ which only handles exceptions from `AssetRemote.fetch_directory()` (i.e. from within `self._pull_source()`)
  * Raise the temporary failure from `self._pull_source()` in case `AssetRemote.fetch_directory()` raises `AssetCacheError`

src/buildstream/_elementsourcescache.py
nanonyme: Thanks, will remove

src/buildstream/_sourcecache.py
nanonyme: I'm a bit reluctant to do refactoring without really understanding codebase that well.

src/buildstream/_sourcecache.py
```diff
@@ -148,7 +148,7 @@ def pull(self, source):
                 source.info("Remote cas ({}) does not have blob {} cached".format(remote, e.blob))
                 continue
             except CASError as e:
-                raise SourceCacheError("Failed to pull source {}: {}".format(display_key, e)) from e
+                raise SourceCacheError("Failed to pull source {}: {}".format(display_key, e), temporary=True) from e
```
gtristan, line 151: This one looks correct, just adding a note here because it explains something about the previous comment related to the same function `SourceCache.pull()`.

In this case we handle `CASError` after calling `CASCache.fetch_directory()`.

In this case, `CASCache.fetch_directory` *does* raise a `CASError` (because `CasCacheError` _is a `CASError`_), however in the previous case, `AssetRemote.fetch_directory()` does *not* raise a `CASError` but raises an `AssetCacheError` instead.

This probably explains why the previous code block incorrectly expects a `CASError`.

src/buildstream/_sourcecache.py
```diff
@@ -240,7 +240,9 @@ def _pull_source(self, source_ref, remote):
 
         except grpc.RpcError as e:
             if e.code() != grpc.StatusCode.NOT_FOUND:
-                raise SourceCacheError("Failed to pull source with status {}: {}".format(e.code().name, e.details()))
```
gtristan, line 243: Here we are again in `self._pull_source()`, also related to the previous comments in `SourceCache.pull()`.

Here we are handling `grpc.RpcError` in response to `AssetRemote.fetch_directory()`, however `AssetRemote.fetch_directory()` _already handles `grpc.RpcError` and raises `AssetCacheError`_, it is not possible to catch a `grpc.RpcError` in this code block.

As mentioned in my comment in `SourceCache.pull()`, we should:
* Have a more isolated `try` block specifically around `AssetRemote.fetch_directory()`
* Raise the temporary `SourceCacheError` here, in response to `AssetRemoteError`

src/buildstream/_sourcecache.py
```diff
@@ -254,5 +256,7 @@ def _push_source(self, source_ref, remote):
 
         except grpc.RpcError as e:
             if e.code() != grpc.StatusCode.RESOURCE_EXHAUSTED:
-                raise SourceCacheError("Failed to push source with status {}: {}".format(e.code().name, e.details()))
+                raise SourceCacheError(
```
gtristan, line 259: This function `_push_source()` needs the same treatment as `_pull_source()` above:

* It is impossible to receive `grpc.RpcError` here, as this is already handled in `AssetCache.push_directory()`
* We need to handle `AssetCacheError` instead, and raise the temporary `SourceCacheError`
*

gtristan:
Thanks a lot @nanonyme for making this patch, I was going to take a look at this but you beat me to it !

I gave some review comments, apparently looking at this has uncovered some inconsistencies and problems in our current error handling in this area.

src/buildstream/_sourcecache.py
gtristan: I've setup an extension to this merge request at #1646

nanonyme:
I'm closing this. The work is continued in https://github.com/apache/buildstream/pull/1646

PR #1654, opened 2022-06-05
bst-1: .github: Run tests in fedora 36

jjardon:
And remove soon to be deprecated fedora 34

Backport of https://github.com/apache/buildstream/pull/1647

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -74,7 +74,7 @@ release = __version__
 #
 # This is also used if you do content translation via gettext catalogs.
 # Usually you set "language" from the command line for these cases.
-language = None
+language = 'en'
 
 # There are two options for replacing |today|: either, you set today to some
 # non-false value, then it is used:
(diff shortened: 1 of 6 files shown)


doc/source/conf.py
```diff
@@ -74,7 +74,7 @@
 #
 # This is also used if you do content translation via gettext catalogs.
 # Usually you set "language" from the command line for these cases.
-language = None
+language = 'en'
```
juergbi, line 77: Is this needed with more recent versions of sphinx? This change is not in master. Is there a reason for this difference?

jjardon: @juergbi Yeah, it will fail with a warning saying the language "None" is not supported

warnings are activated as errors and that's why the build will fail: maybe that is not activated in master

nanonyme: Can you create issue about this for master? It sounds like a bug that the warning is simply ignored.

jjardon: Sure, I will

juergbi (approved):

PR #1655, opened 2022-06-12
source.py: Don't ignore project.refs of subprojects

juergbi:
Use `project.refs` in subprojects if the top-level project is also configured to use `project.refs` but doesn't override the particular ref.


gtristan:
Right, I gave more thought to this patch and it makes more sense to me now.

* We can **not** write to `project.refs` in subprojects when tracking new refs (we can only track them in the toplevel, if the toplevel uses `project.refs`)
* We **can** read the `project.refs` _from subprojects_
* We **still** prioritize local `project.refs` in the toplevel if they exist when _reading_ refs for subproject elements

nanonyme:
Docs failure should be fixed by rebase.

Issue #1653, opened 2022-05-29
Error running BuildStream 1.6 with current protc


nanonyme:
When trying to install BuildStream on Fedora 36, I get error
```
TypeError: Descriptors cannot not be created directly.
If this call came from a _pb2.py file, your generated code is out of date and must be regenerated with protoc >= 3.19.0.
If you cannot immediately regenerate your protos, some other possible workarounds are:
 1. Downgrade the protobuf package to 3.20.x or lower.
 2. Set PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python (but this will use pure-Python parsing and will be much slower).
 ```

fishrockz:
I have hit this also.

dbnicholson:
Seems to me the options are:

* Have the pb2 files compiled at install time.
* Require protobuf < 4

I think the latter is the path of least resistance for now.

abderrahim:
Requiring protobuf < 4 is just a temporary measure, once distros start getting protobuf 4 we'll get a problem again (see https://github.com/apache/buildstream/issues/1623)

abderrahim:
Reading the error message carefully, it seems we can regenerate the pb2 files with the latest 3.x and they'll work with both 3.x and 4.x

gtristan:
Is this verified to be fixed by #1664 ? Can we close this ?

nanonyme:
Should be fixed, yes.

nanonyme:
If anyone else than me wants to confirm, would be great. I tested with protobuf 4.20.something and 3.19.something before change was merged.

PR #1662, opened 2022-06-28
Demand protobuf older than 4.0

nanonyme:
Fixes https://github.com/apache/buildstream/issues/1653


nanonyme:
@abderrahim suggested we should instead regenerate these with latest 3.x series protobuf.

PR #1663, opened 2022-06-28
Ensure there is no extra whitespace in ref

nanonyme:


nanonyme:
Created into bst-plugins-external as https://gitlab.com/BuildStream/bst-plugins-experimental/-/merge_requests/186

abderrahim (approved):

Issue #1667, opened 2022-06-30
Missing checkout to correct branch for installing from source 1.6.6


LawrenceHunter:
For the 'Installing' instructions the page https://docs.buildstream.build/1.6/install_linux_distro.html#installing-from-source is missing `git checkout bst-1` before `pip3 install --user -e .`

jjardon:
@gtristan @abderrahim  I think this can be closed now?

PR #1668, opened 2022-07-01
Add checkout step to linux distro install

LawrenceHunter:
Updated Installing to after cloning and cd into the buildstream directory to then git checkout to the relevant tag of the install


LawrenceHunter:
Resolves #1667

LawrenceHunter:
Wrong branch

PR #1669, opened 2022-07-01
Add checkout to version in install linux distro

### LINKED ISSUE #1667: Missing checkout to correct branch for installing from source 1.6.6

LawrenceHunter:
Add command to checkout to version tag when installing buildstream. Resolves #1667

### DIFF
--- a/doc/source/install_linux_distro.rst
+++ b/doc/source/install_linux_distro.rst
@@ -127,7 +127,7 @@ git repository and install it as a regular user::
 
   git clone https://gitlab.com/BuildStream/buildstream.git
   cd buildstream
-  git checkout <VERSION>
+  git checkout <VERSION_TAG>
   pip3 install --user -e .
 
 This will install buildstream's pure python dependencies into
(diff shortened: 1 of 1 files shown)


doc/source/install_linux_distro.rst
```diff
@@ -127,6 +127,7 @@ git repository and install it as a regular user::
 
   git clone https://gitlab.com/BuildStream/buildstream.git
   cd buildstream
+  git checkout <VERSION>
```
jjardon: Thanks @LawrenceHunter1 , unless this is replaced at build time with the latest stable tag, I think is better to change this to "git checkout bst-1"

LawrenceHunter: That works better for the latest version but it wouldn't resolve the same issue for example 1.4.2 which is also missing the branching command. This is based on my assumption when reading the docs you're reading the latest version, as the Installing from source page does not vary between which version is selected to install.

jjardon: I think you are right; but then you need to explain what "version" is (tag in the git repo), so there is no confusion

gtristan:
Looks good to me !

Thanks for your contribution.

PR #1671, opened 2022-07-03
_cas/casdprocessmanager.py: Assert minimal required version of buildbox-casd

### LINKED ISSUE #1189: Check versions of buildbox
[See original issue on GitLab](https://gitlab.com/BuildStream/buildstream/-/issues/1189)
In GitLab by [[Gitlab user @traveltissues]](https://gitlab.com/traveltissues) on Nov 1, 2019, 12:18

A number of issues have been reported with buildstream which were resolved by changing buildbox versions. This is not clearly suggested/indicated to the user. If possible buildstream should attempt to check the versions of buildbox that will be used before doing any work.

gtristan:
Here we pass along the Messenger object to the casdprocessmanager so that
we can issue a warning on startup in the case that we are unable to properly
determine the installed buildbox-casd version.

An error is raised if we successfully identify the installed version and
it is less than the required version.

Fixes #1189

### DIFF
--- a/src/buildstream/_cas/casdprocessmanager.py
+++ b/src/buildstream/_cas/casdprocessmanager.py
@@ -18,14 +18,16 @@
 import contextlib
 import threading
 import os
+import re
 import random
 import shutil
 import stat
 import subprocess
 import tempfile
 import time
-import psutil
+from subprocess import CalledProcessError
 
+import psutil
 import grpc
 
 from .._protos.build.bazel.remote.asset.v1 import remote_asset_pb2_grpc
@@ -40,6 +42,14 @@ _CASD_MAX_LOGFILES = 10
 _CASD_TIMEOUT = 300  # in seconds
 
 
+#
+# Minimum required version of buildbox-casd
+#
+_REQUIRED_CASD_MAJOR = 0
+_REQUIRED_CASD_MINOR = 0
+_REQUIRED_CASD_MICRO = 58
+
+
 # CASDProcessManager
 #
 # This manages the subprocess that runs buildbox-casd.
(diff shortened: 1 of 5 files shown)


gtristan:
It is difficult to add tests for the various failure modes in CI as this would require installing various versions of buildbox-casd, but I have tested insufficient versions, sufficient versions and undetermined ("unknown") versions locally.

src/buildstream/_cas/casdprocessmanager.py
```diff
@@ -92,6 +106,60 @@ def __init__(self, path, log_dir, log_level, cache_quota, remote_cache_spec, pro
...
+    # a warning is produced to inform the user.
+    #
+    def _check_casd_version(self, messenger):
+        #
+        # We specify a trailing "path" argument because some versions of buildbox-casd
+        # require specifying the storage path even for invoking the --version option.
+        #
+        casd_args = [utils.get_host_tool("buildbox-casd")]
+        casd_args.append("--version")
+        casd_args.append("/")
```
fishrockz, line 126: Is this slash meant to deal with the odd bug were `buildbox-casd --version` fails but  `buildbox-casd --version any_string_will_do` works ? 

have we reported this upstream to buildbox-casd? and could could have a comment to this effect as its rather odd to add the slash without knowledge of this bug? I'm sure i will have forgotten this quirk in a few weeks.

gtristan: The comment already reads:

```
        #
        # We specify a trailing "path" argument because some versions of buildbox-casd
        # require specifying the storage path even for invoking the --version option.
        #
```

And the issue was raised here: https://gitlab.com/BuildGrid/buildbox/buildbox-casd/-/issues/86

abderrahim:
Can we do something similar about buildbox-fuse?

gtristan:
> Can we do something similar about buildbox-fuse?

Some of this requires addressing more of https://gitlab.com/BuildGrid/buildbox/buildbox-casd/-/issues/86, `buildbox-run` for instance I believe does not yet expose a `--version` argument, it would be best to have the command line parsing code in `buildbox-common` automatically support a `--version` argument and update all the binaries to provide a version (this might require updating the relevant components with similar CMake foo to derive their versions from git tags, falling back to "unknown").

For `buildbox-fuse` I'm not sure it makes sense to do anything about this, as we do not directly *invoke* `buildbox-fuse` - I guess that this is more of an implementation detail of `buildbox` and we don't really have any business knowing that it exists.

That said, on the one hand it seems unimportant since `buildbox` binaries should generally be distributed/updated together, but on the other hand if this is the case, it leaves us open to problems which occur when `buildbox-fuse/buildbox-run` are updated but `buildbox-casd` has *not* been updated and so there is no new version to test for.

Probably it is best to at least:
* Have a separate version check for `buildbox-run`
* Expect that `buildbox-run` will bump it's *own* version number if an important change is introduced in the `buildbox-fuse` implementation detail

abderrahim:
> For `buildbox-fuse` I'm not sure it makes sense to do anything about this, as we do not directly _invoke_ `buildbox-fuse` - I guess that this is more of an implementation detail of `buildbox` and we don't really have any business knowing that it exists.

The thing is currently `buildbox-fuse` is an implementation detail of `buildbox-casd` and `buildbox-run` doesn't really have any business knowing that it exists. I agree that buildbox-run, as part of the buildbox project, has more business with this than buildstream.

>     * Expect that `buildbox-run` will bump it's _own_ version number if an important change is introduced in the `buildbox-fuse` implementation detail

Or possibly print the version number of buildbox-fuse if it is using it. But as I said, buildbox-fuse is currently an implementation detail in the LocalCAS protocol, and there is no real way to know that it is being used.

PR #1677, opened 2022-07-08
Clean the build directory after a successful build

juergbi:
Capturing the build directory takes time and uses disk space of the CAS cache. If the `cache-buildtrees` option is set to `auto` (the default) or `never`, the contents of the build directory can be removed after a successful build as the build tree is no longer needed, avoiding the capture costs.

### DIFF
--- a/src/buildstream/sandbox/_sandboxreapi.py
+++ b/src/buildstream/sandbox/_sandboxreapi.py
@@ -278,3 +278,11 @@ class _SandboxREAPIBatch(_SandboxBatch):
 
     def create_empty_file(self, name):
         self.script += "touch -- {}\n".format(shlex.quote(name))
+
+    def clean_directory(self, name):
+        # Do not treat error during cleanup as a fatal build error
+        self.script += "rm -rf -- {} || true\n".format(shlex.quote(name))
+        if self.first_command:
+            # Working directory may be a subdirectory of the build directory.
+            # Recreate it if necessary as output capture requires the working directory to exist.
+            self.script += "mkdir -p {} || true\n".format(shlex.quote(self.first_command.cwd))
--- a/src/buildstream/sandbox/sandbox.py
+++ b/src/buildstream/sandbox/sandbox.py
@@ -94,8 +94,6 @@ class Sandbox:
         self.__stdout = kwargs["stdout"]
         self.__stderr = kwargs["stderr"]
 
-        self._build_directory = None
-        self._build_directory_always = None
         self._vdir = None  # type: Optional[Directory]
         self._usebuildtree = False
 
@@ -541,20 +539,6 @@ class Sandbox:
     #
     def _disable_run(self):
         self.__allow_run = False
-
-    # _set_build_directory()
-    #
-    # Sets the build directory - the directory which may be preserved as
-    # buildtree in the artifact.
-    #
-    # Args:
-    #    directory (str): An absolute path within the sandbox
-    #    always (bool): True if the build directory should always be downloaded,
-    #                   False if it should be downloaded only on failure
-    #
-    def _set_build_directory(self, directory, *, always):
-        self._build_directory = directory
-        self._build_directory_always = always
 
 
 # SandboxFlags()
--- a/src/buildstream/sandbox/sandbox.py
+++ b/src/buildstream/sandbox/sandbox.py
@@ -524,6 +524,26 @@ class Sandbox:
             cwd_vdir = vdir.open_directory(cwd.lstrip(os.sep), create=True)
             cwd_vdir._create_empty_file(name)
 
+    # _clean_directory()
+    #
+    # Remove the contents of the specified directory.
+    #
+    # Args:
+    #    path (str): The path of the directory to be cleaned
+    #
+    def _clean_directory(self, path):
+        if self.__batch:
+            batch_clean = _SandboxBatchCleanDirectory(path)
+
+            current_group = self.__batch.current_group
+            current_group.append(batch_clean)
+        else:
+            vdir = self.get_virtual_directory()
+            relative_path = path.lstrip(os.sep)
+            if vdir.exists(relative_path):
+                vdir.remove(relative_path, recursive=True)
+                vdir.open_directory(relative_path, create=True)
+
     # _get_element_name()
     #
     # Get the plugin's element full name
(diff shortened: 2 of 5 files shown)


src/buildstream/sandbox/sandbox.py
```diff
@@ -526,6 +526,32 @@ def _create_empty_file(self, name):
             cwd_vdir = vdir.open_directory(cwd.lstrip(os.sep), create=True)
             cwd_vdir._create_empty_file(name)
 
+    # _clean_build_directory()
```
gtristan: I'm not particularly fond of the business logic growing here in the sandbox code, it looks like we're building on this `Sandbox._set_build_directory()` API which appears to be deadcode which we neglected to cleanup (grep tells me that there is no code observing this `_build_directory` variable anywhere).

Maybe it would make more sense to nuke these deadcode paths and some more generic private cleanup API controlled by element.py's business logic decisions.

src/buildstream/sandbox/_sandboxreapi.py
```diff
@@ -278,3 +278,10 @@ def execute_command(self, command):
 
     def create_empty_file(self, name):
         self.script += "touch -- {}\n".format(shlex.quote(name))
+
+    def clean_directory(self, name):
```
gtristan, line 282: This implementation appears to add shell commands, assuming that `rm` and `mkdir` will exist in the sandbox at runtime.

It looks like unfortunately we already went down this path with `touch` in `create_empty_file()` above.

My understanding is that *currently* we execute sandbox commands with `/bin/sh` by default, but there is nothing preventing us from working with other interpretors, or running commands in a sandbox that has no `/bin/sh` (i.e. we need only add some options to `BuildElement` to select what shell to use when invoking the sandbox).

Is having a shell a new limitation stemming from remote execution ? Shouldn't we have some way of mutating the remote sandbox content (adding/removing files/directories) without executing commands ?

src/buildstream/sandbox/sandbox.py
juergbi: Would it be acceptable to check `context.cache_buildtrees` in `BuildElement` as condition to calling a more generic method in `Sandbox`?

Or should there be a (public?) `clean_build_directory()` method in `Element`, which performs that check and calls the `Sandbox` method?

src/buildstream/sandbox/_sandboxreapi.py
juergbi: `mkdir` is already required to ensure the working directory exists a bit further up. Expecting `rm` to exist is not a big stretch. Also, the build will not fail if `rm` doesn't exist, so this optimization shouldn't break anything.

Yes, using a shell script was part of the switch to REAPI, initially only for remote execution and later also for buildbox-run. With remote execution we can only execute a single command in a sandbox but that can be `sh`. We can manipulate the input tree beforehand in any way, of course, but can't manipulate the sandbox while remote execution is already running. Issuing individual commands using separate sandboxes / REAPI executions would be terrible for performance.

Sandbox support without POSIX `sh` is certainly worth discussing. Configuration at the point of `sandbox.batch()` may make sense. However, I don't think this branch makes anything worse as it generates commands in the same place (`SandboxREAPI`) as what's already in master and doesn't go beyond very basic POSIX commands.

src/buildstream/sandbox/sandbox.py
gtristan: > Would it be acceptable to check `context.cache_buildtrees` in `BuildElement` as condition to calling a more generic method in `Sandbox`?

Looking at `BuildElement` now, it looks like we've done a good job of avoiding any internal API interactions - it would be a good thing to keep it this way (anything a build element can do, any external plugin can do).

> Or should there be a (public?) clean_build_directory() method in Element, which performs that check and calls the Sandbox method?

Is there any reason why this cleaning of the build directory should be optional ?

Perhaps a better criteria would be to clean the build directory in any case where `%{build-root}` is found outside of `%{install-root}` (this would avoid destructive behavior like in script elements where `%{build-root}` might be inside, or equal to the `%{install-root}`) ?

juergbi: If we clean the build directory even with the `cache-buildtrees: always` configuration, that option becomes useless, if I'm not missing anything.

src/buildstream/sandbox/_sandboxreapi.py
gtristan: Ok this is just unfortunate but does not worsen the situation, I suppose there is nothing we can do at this time to undo these limitations without significant work.

Probably we need some way to batch multiple generic (not strictly `/bin/sh`) program invocations without wrapping them up in a shell script, and a way to batch sandbox program executions with other directory manipulation commands in a single remote execution request, in order to have better control.

I guess the ship has sadly sailed for now and it will be a long time until we can lose these limitations, it would make sense to document these limitations too.

src/buildstream/sandbox/sandbox.py
juergbi: I've moved the business logic to a `Element.clean_buildtree()` method. Is this what you had in mind?

gtristan: > If we clean the build directory even with the `cache-buildtrees: always` configuration, that option becomes useless, if I'm not missing anything.

Maybe I'm missing something but I thought the sanbox functions in question were already only *maybe_clean_buildtree* depending on some business logic.

Indeed, looking at the updated code you appropriately note:

> This will have no effect if the build tree is required after the build.

And I *think* that the conditional statement you added to this function, makes the function safe to use for basically any element (`compose`, `import`, `filter`, etc...):

```python

        if install_root and (build_root.startswith(install_root) or install_root.startswith(build_root)):
            # Preserve the build directory if cleaning would affect the install directory
            return
```

Am I missing some detail as to why it would be necessary for a plugin implementation to explicitly invoke this, rather than to have this business logic run unconditionally after the build / before caching... directly in `Element._assemble()` ?

gtristan: After discussing on IRC, @juergbi reminds me that this function must run commands in a shell in order to be batched and avoid capturing, which sort of explains why this must be explicitly invoked by the plugin.

* It must happen at the end of a batch in order to preserve the optimization and avoid the unnecessary capture
* It must only be performed by an element which has a shell capable or running `rm`

As such, this approach appears to make sense.

We should make it obvious in the function name or docstrings that this cleanup requires a shell to work.

gtristan (approved):

PR #1679, opened 2022-07-10
setup.py, setup.cfg: Disable gRPC fork support

juergbi:
gRPC fork support is not compatible with the default polling engine of
certain gRPC versions on modern Linux systems (epollex) and it also does
not work if another thread is calling into gRPC. It also causes Python
to segfault in the test suite in some environments.

As BuildStream, after gRPC initialization, uses `fork()` only to
`exec()` subprocesses, the gRPC fork support (`atfork` handler) is not
needed. Setting `GRPC_ENABLE_FORK_SUPPORT=0` disables gRPC fork support,
fixing Python segmentation faults and error messages caused by the
`atfork` handler.

https://github.com/grpc/grpc/issues/29044

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -13,6 +13,7 @@ norecursedirs = src tests/integration/project tests/plugins/loading tests/plugin
 python_files = tests/*/*.py
 env =
     D:BST_TEST_SUITE=True
+    D:GRPC_ENABLE_FORK_SUPPORT=0
 
 markers =
     datafiles: data files for tests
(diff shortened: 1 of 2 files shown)


gtristan:
Ohhh this is super neat, I was actually looking for a way to generate the entrypoint while setting env vars but came up short, and didn't notice the hack we had in place, so this does just that.

An extension to the leading comment (which currently states that this hack is only about load time performance) might be nice, but I'm fine with it either way.

gtristan (approved):

PR #1680, opened 2022-07-10
Use `StageTree` to stage previous sources

juergbi:
This will use buildbox-fuse (via buildbox-casd) if available, improving the performance of the stage operation. With future buildbox-casd optimizations, this will also speed up the import of the modified source directory.

### DIFF
--- a/src/buildstream/_cas/cascache.py
+++ b/src/buildstream/_cas/cascache.py
@@ -444,6 +444,51 @@ class CASCache:
 
         return utils._message_digest(root_directory)
 
+    # stage_directory():
+    #
+    # A contextmanager to stage a CAS directory tree in the local filesystem.
+    #
+    # This makes the specified directory tree temporarily available for local
+    # filesystem access. This may use FUSE or hardlinking.
+    #
+    # Args:
+    #     directory_digest (Digest): The digest of a directory
+    #
+    # Yields:
+    #     (str): The local filesystem path
+    #
+    @contextlib.contextmanager
+    def stage_directory(self, directory_digest):
+        local_cas = self.get_local_cas()
+
+        request = local_cas_pb2.StageTreeRequest()
+        request.root_digest.CopyFrom(directory_digest)
+
+        done_event = threading.Event()
+
+        def request_iterator():
+            yield request
+
+            # Wait until staged tree is no longer needed
+            done_event.wait()
+
+            # A second (empty) request indicates that the staging location can be cleaned up
+            yield local_cas_pb2.StageTreeRequest()
+
+        response_stream = local_cas.StageTree(request_iterator())
+
+        try:
+            # Read primary response and yield staging location
+            response = next(response_stream)
+            yield response.path
+
+            # Staged tree is no longer needed
+            done_event.set()
+            # Wait for cleanup to complete
+            next(response_stream)
+        except StopIteration as e:
+            raise CASCacheError("Unexpected end of response stream for StageTree") from e
+
     # missing_blobs_for_directory():
     #
     # Determine which blobs of a directory tree are missing on the remote.
--- a/src/buildstream/_elementsources.py
+++ b/src/buildstream/_elementsources.py
@@ -18,7 +18,7 @@ import os
 from contextlib import contextmanager
 from typing import TYPE_CHECKING, Iterator
 
-from . import _cachekey, utils
+from . import _cachekey
 from ._exceptions import SkipJob
 from ._context import Context
 from ._protos.buildstream.v2 import source_pb2
@@ -459,7 +459,8 @@ class ElementSources:
     #   stop (Source): Only stage sources listed before this source
     #
     def _stage(self, *, stop=None):
-        vdir = CasBasedDirectory(self._context.get_cascache())
+        cas = self._context.get_cascache()
+        vdir = CasBasedDirectory(cas)
 
         for source in self._sources:
             if source == stop:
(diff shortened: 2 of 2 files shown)


abderrahim (commented):
Looks good, however I wouldn't want to merge it without adding a test without buildbox-fuse. This is now a critical feature that needs to be tested.

src/buildstream/_cas/cascache.py
```diff
@@ -443,6 +443,48 @@ def import_directory(self, path: str, properties: Optional[List[str]] = None) ->
 
         return utils._message_digest(root_directory)
 
+    # stage_directory():
+    #
+    # A contextmanager to stage a CAS directory tree in the local filesystem.
+    #
+    # This makes the specified directory tree temporarily available for local
+    # filesystem access. This may use FUSE or hardlinking.
```
gtristan, line 452: So if I understand correctly, this patch *may* use `buildbox-fuse` if available to create the local filesystem access required for a plugin to go and do it's work with previous sources staged.

What worries me here is *"This may use FUSE or hardlinking."*, is it possible that this uses hardlinking *without* FUSE ?

We should avoid a scenario where we expose writable hardlinks into the local CAS, to a source plugin which *may* in some unlikely even modify a staged file (instead of replacing/re-installing a new one).

If this is safe and does not expose writable hardlinks to the client source plugins, we should name/document this function accordingly I think.

juergbi: `StageTree` is the same method that is also used by buildbox-run-bubblewrap. I.e. this shouldn't add any risk. For safe hardlinking, buildbox-casd has to run as a different user, in which case this should be safe for both build sandboxes and sources.

That said, I don't remember whether we already verify that the setup is correct.

juergbi:
> Looks good, however I wouldn't want to merge it without adding a test without buildbox-fuse. This is now a critical feature that needs to be tested.

Right, we seem to have lost the userchroot CI job, which used hardlinking, when moving from GitLab to GitHub. I don't remember what the reason for that was. Adding that would definitely be a good idea.

juergbi:
> Right, we seem to have lost the userchroot CI job, which used hardlinking, when moving from GitLab to GitHub. I don't remember what the reason for that was. Adding that would definitely be a good idea.

I've tried re-adding the userchroot CI job. Unfortunately, an update of `pytest` (commit 1278f8b9) broke these tests because it sets temp directory permissions to `0700`, ignoring the umask, even if the temp directory is within the home directory (not in public `/tmp`). This makes it impossible for buildbox-casd to run as separate user in the same group, which is required for safe hardlinking. A workaround may be possible but I can't think of a sufficiently clean one right now.

In my opinion, this PR shouldn't be blocked by this because `StageTree` is already used by buildbox-run, i.e. if `StageTree` doesn't work safely, BuildStream is broken anyway, independent of this change.

However, as we currently only test with buildbox-fuse, it may make sense to consider all other configurations unsupported. I.e. add a buildbox-fuse check to BuildStream initialization and fail if it is not available. When we properly support environments without buildbox-fuse (with CI and safeguard in place to ensure the setup can't corrupt the cache), we can lift the buildbox-fuse requirement.

nanonyme:
Right, you could create your own fixture that uses pytest fixture but invokes chmod. But you would need to rewrite a lot of code for your own fixture to be used consistently.

nanonyme:
Right, you could create your own fixture that uses pytest fixture but invokes chmod. But you would need to rewrite a lot of code for your own fixture to be used consistently.
That said, if someone was to start rewriting things, might as well fix tmpdir not to be used since it's now deprecated https://github.com/pytest-dev/pytest/commit/70cac3d464905cb3f7cedba9ea4f61dc56613f55
Edit: okay, so it's the base temp dir that is the problem, not the tempdir itself. Then it's trickier indeed to workaround.

gtristan:
I concur that this should not be blocked because certain configurations of `buildbox` are not necessarily tested.

> However, as we currently only test with buildbox-fuse, it may make sense to consider all other configurations unsupported. I.e. add a buildbox-fuse check to BuildStream initialization and fail if it is not available. When we properly support environments without buildbox-fuse (with CI and safeguard in place to ensure the setup can't corrupt the cache), we can lift the buildbox-fuse requirement.

I don't think I like saying that *"BuildStream does not support this Buildbox setup"* as a statement from the BuildStream front.

I rather think: *"BuildStream uses Buildbox"* and from BuildStream's perspective, it is Buildbox's job to function correctly as an abstraction layer for execution on varoius platforms (I think that was the idea).

In one sense, it would be nice for BuildStream to have CI for as many variant setups as possible so that we can find and report issues to other upstream projects we depend on, but in another sense; I feel that *how* buildbox works is buildbox's problem and not BuildStream's.

src/buildstream/_cas/cascache.py
gtristan: One thing that is not obvious, is that this `StageTree` being safe for running commands inside a sandbox using buildbox-run is equally safe for us to go right ahead and run buildstream subprocesses. I would not have made that assumption.

I of course believe you that it is safe, just saying that it doesn't seem obvious that it *would* be safe :)

abderrahim:
My concern isn't whether or not one can corrupt the buildbox cache, that's indeed buildbox's problem (AFAICT, it is fool-proof but not evil-proof. For the latter, you need to run buildbox-casd as a different user).

What I'm concerned about is whether or not this is expected to work without buildbox-fuse. If a simple  `patch` or `git am` on a tree staged with hardlinks doesn't work, then we're in trouble because buildstream becomes unusable in such a setup. Currently, it just works fine without buildbox-fuse, until we want a script element root.

juergbi:
> What I'm concerned about is whether or not this is expected to work without buildbox-fuse. If a simple `patch` or `git am` on a tree staged with hardlinks doesn't work, then we're in trouble because buildstream becomes unusable in such a setup. Currently, it just works fine without buildbox-fuse, until we want a script element root.

That's a good point. I just did a manual test with both `patch` and `git am` and they work fine with hardlinked read-only files. `patch` prints the warning "File f is read-only; trying to patch anyway" but it successfully patches the file (breaking the hardlink as expected).

BTW: I'm working on a buildbox-casd patch to fall back to file copies in case buildbox-fuse is unavailable and buildbox-casd is running as the same user as BuildStream/buildbox-run, which should make cache corruption impossible (unless running with superuser privileges). A possible future improvement is to use zero-copy reflinks/clones instead, on (file)systems that support that.

PR #1691, opened 2022-07-20
Retry failed artifacts in non-strict mode when dependencies have changed

gtristan:


gtristan:
This was discussed and reviewed on slack.

Some changes since our discussion:
* The test case was enhanced such that we provide a `strong_key != strict_key` scenario for the dependency of the toplevel `target.bst` element
  * This provides a better test scenario which exercises the code change more rigorously
  * This led to discovery of other issues which were not being encountered when the failed artifact's `strong_key == strict_key`
* A new commit is added to provide a global `context.build` state, which allows the core business logic to behave differently in the case that buildstream was launched with the objective of building something
  * Without this, we end up discarding the failed build artifact in undesirable scenarios
    * When running `bst show` the failed artifact does not show up as *failed*
    * When running `bst artifact push` there is no failed artifact to push
    * When debugging *why* the build failed with `bst shell --build` there is no failed build artifact to inspect
* A new helper function `__calculate_strong_cache_key()` is added in order to streamline code which runs this calculation

juergbi (approved):

PR #1673, opened 2022-07-05
Update gitlab link to github

LawrenceHunter:
The link given in the documentation is to the old gitlab repository and as such doesn't have the latest tags etc

PR #1694, opened 2022-07-22
Some docs fixes for bst-1

gtristan:


gtristan:
Rebased changes from #1673 and merging here.

PR #1695, opened 2022-07-22
Resolve conditionals statements in the project root

gtristan:


gtristan:
This branch still has issues.

The project load process needs to parse a few things before parsing options and resolving the options (including the definition of the options themselves for example), blindly allowing option resolution at the toplevel will result bad error reporting in the case that yaml keys that are loaded *pre-option-resolution* are declared conditionally.

gtristan:
Updated patch, added an internal feature which allows the project to specify restricted keys at conditional statement resolution time.

This results in user errors like such:

```
Error loading project: project.conf [line 16 column 6]: The 'options' key cannot be specified conditionally
```

PR #1698, opened 2022-07-22
storage: Add `collect_result` parameter to `import_files()`

juergbi:
Collecting data for `FileListResult` may incur significant overhead as it always requires iterating over all files in the directory, recursively, even if the whole directory can be imported as is.


gtristan (approved):
Ah this one was a long time coming... good show :)

Issue #1455, opened 2021-03-15
BUG when trying to access an element through junction on the command line


abderrahim:
Steps to reproduce:
- have a junction `junction.bst` with `overrides`
- use `bst show` or `bst build` on an element in the junction `junction.bst:element.bst`

I get the following error:

```
    Traceback (most recent call last):
      File "/home/abderrahim/.virtualenvs/bst-master/bin/bst", line 8, in <module>
        sys.exit(cli())
      File "/home/abderrahim/.virtualenvs/bst-master/lib/python3.9/site-packages/click/core.py", line 829, in __call__
        return self.main(*args, **kwargs)
      File "/home/abderrahim/.virtualenvs/bst-master/lib/python3.9/site-packages/buildstream/_frontend/cli.py", line 258, in override_main
        original_main(self, args=args, prog_name=prog_name, complete_var=None, standalone_mode=standalone_mode, **extra)
      File "/home/abderrahim/.virtualenvs/bst-master/lib/python3.9/site-packages/click/core.py", line 782, in main
        rv = self.invoke(ctx)
      File "/home/abderrahim/.virtualenvs/bst-master/lib/python3.9/site-packages/click/core.py", line 1259, in invoke
        return _process_result(sub_ctx.command.invoke(sub_ctx))
      File "/home/abderrahim/.virtualenvs/bst-master/lib/python3.9/site-packages/click/core.py", line 1066, in invoke
        return ctx.invoke(self.callback, **ctx.params)
      File "/home/abderrahim/.virtualenvs/bst-master/lib/python3.9/site-packages/click/core.py", line 610, in invoke
        return callback(*args, **kwargs)
      File "/home/abderrahim/.virtualenvs/bst-master/lib/python3.9/site-packages/click/decorators.py", line 33, in new_func
        return f(get_current_context().obj, *args, **kwargs)
      File "/home/abderrahim/.virtualenvs/bst-master/lib/python3.9/site-packages/buildstream/_frontend/cli.py", line 607, in show
        dependencies = app.stream.load_selection(elements, selection=deps, except_targets=except_)
      File "/home/abderrahim/.virtualenvs/bst-master/lib/python3.9/site-packages/buildstream/_stream.py", line 166, in load_selection
        target_objects = self._load(
      File "/home/abderrahim/.virtualenvs/bst-master/lib/python3.9/site-packages/buildstream/_stream.py", line 1566, in _load
        elements, except_elements, artifacts = self._load_elements_from_targets(
[...]

PR #1701, opened 2022-07-25
Fix loading cross project element

gtristan:
Provides fix and test case for #1455

### DIFF
--- a/src/buildstream/_loader/loader.py
+++ b/src/buildstream/_loader/loader.py
@@ -192,6 +192,14 @@ class Loader:
         junction_path = name.split(":")
         loader = self
 
+        #
+        # In this case we are attempting to load a subproject element via the
+        # command line instead of referencing the subproject through a project
+        # element or otherwise.
+        #
+        if provenance_node is None and load_subprojects:
+            self.project.ensure_fully_loaded()
+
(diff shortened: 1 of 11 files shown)


gtristan:
This patch apparently fixes the issue in question, however it does not appear to be the best/correct fix.

We have two behaviors which should be equal:

* `bst show subproject.bst:target.bst` This is showing a stack trace

* `bst show target.bst` where the toplevel project `target.bst` depends on `subproject.bst:target.bst`, This is not showing a stack trace.

The current fix causes the first variant to not produce a stack trace, but exercises a clearly different load pattern - it would be better to not cause the core to accomadate the second loading pattern but instead cause both use cases to exercise the same loading pattern.

It is still unclear why they are not behaving the same way.

gtristan:
After further investigation I was able to come up with a patch which makes me a bit more comfortable, and focuses on `Loader.get_loader()`

* It adjusts the circular dependency handling to consider the case that a loader lookup occurred without provenance, as it does when specifying `subproject.bst:element.bst` on the command line, and was the likely source of the stack trace observed in #1455 
* When addressing a subproject element on the command line, ensure that the toplevel project is resolved at the beginning
  * This enforces a more consistent load pattern and addresses other issues and stack traces I had managed to produce with my test cases

PR #1705, opened 2022-07-26
doc/source/main_install.rst: Buildstream2 doesnt have default plugins anymore

jjardon:


gtristan:
I don’t believe BuildStream’s core `tar` plugin requires `tar` on the host, core python API is used for this.

Is there another reason to require `tar` ?

jjardon:
> I don’t believe BuildStream’s core `tar` plugin requires `tar` on the host, core python API is used for this.
> 
> Is there another reason to require `tar` ?

@gtristan Nope, I have updated the PR

nanonyme:
@jjardon @gtristan core tar plugin does need lzip though. It was erroneously removed here. It's best to have it installed so things don't surprisingly fail later. It's also important to mention it because many distributions (eg Silverblue) do not have it installed by default.

nanonyme:
See https://github.com/apache/buildstream/blob/master/src/buildstream/plugins/sources/tar.py#L101-L104

jjardon:
well spotted @nanonyme ; opened https://github.com/apache/buildstream/pull/1710

Issue #1613, opened 2022-03-10
Terminating jobs doesn't work


abderrahim:
Interrupt jobs doesn't work, I noticed this with fetch jobs before (where it seems that bst hanged but git is still downloading in the background) and today I noticed it with build jobs too. Here are the logs: notice that it says `STATUS  Build terminating` for all four build jobs, yet they continue until they end with `SUCCESS Caching artifact`.

```
User interrupted with ^C

Choose one of the following options:
  (c)ontinue  - Continue queueing jobs as much as possible
  (q)uit      - Exit after all ongoing jobs complete
  (t)erminate - Terminate any ongoing jobs and exit

Pressing ^C again will terminate jobs and exit

Choice: [continue]: t

Terminating all jobs at user request

[09:46:24][--:--:--][db2e4e4e][    main:sdk/vala-tarball.bst          ] STATUS  Build terminating
[09:46:24][--:--:--][b503e9e5][    main:sdk/gobject-introspection.bst ] STATUS  Build terminating
[09:46:24][--:--:--][a39842c0][    main:freedesktop-sdk.bst:components/libgudev.bst] STATUS  Build terminating
[09:46:24][--:--:--][ecec38b2][    main:sdk/enchant-2.bst             ] STATUS  Build terminating
[09:46:30][00:00:12][db2e4e4e][   build:sdk/vala-tarball.bst          ] SUCCESS Integrating sandbox
[09:46:30][--:--:--][db2e4e4e][   build:sdk/vala-tarball.bst          ] START   Staging sources
[09:46:30][00:00:11][ecec38b2][   build:sdk/enchant-2.bst             ] SUCCESS Integrating sandbox
[09:46:30][--:--:--][ecec38b2][   build:sdk/enchant-2.bst             ] START   Staging sources
[09:46:30][00:00:00][db2e4e4e][   build:sdk/vala-tarball.bst          ] SUCCESS Staging sources
[09:46:30][--:--:--][db2e4e4e][   build:sdk/vala-tarball.bst          ] START   Running commands

    export NOCONFIGURE=1;
    
    if [ -x ./configure ]; then true;
    elif [ -x ./autogen ]; then ./autogen;
    elif [ -x ./autogen.sh ]; then ./autogen.sh;
    elif [ -x ./bootstrap ]; then ./bootstrap;
    elif [ -x ./bootstrap.sh ]; then ./bootstrap.sh;
    else autoreconf -ivf .;
    fi
    ./configure --prefix=/usr \
[...]

PR #1706, opened 2022-07-26
merge ChildJob into Job

abderrahim:
Now that we no longer use subprocesses, we have no need for the parent/child separation of jobs.

### DIFF
--- a/src/buildstream/_scheduler/jobs/elementjob.py
+++ b/src/buildstream/_scheduler/jobs/elementjob.py
@@ -15,7 +15,7 @@
 #        Tristan Daniël Maat <tristan.maat@codethink.co.uk>
 #
 
-from .job import Job, ChildJob
+from .job import Job
 
 
 # ElementJob()
@@ -50,7 +50,7 @@ from .job import Job, ChildJob
 #
 #     complete_cb():
 #
-#     This function will be called when the child task completes
+#     This function will be called in the main thread when the child task completes
 #
 #     Args:
 #        job (Job): The job object which completed
@@ -74,16 +74,6 @@ class ElementJob(Job):
 
     def parent_complete(self, status, result):
         self._complete_cb(self, self._element, status, self._result)
-
-    def create_child_job(self, *args, **kwargs):
-        return ChildElementJob(*args, element=self._element, action_cb=self._action_cb, **kwargs)
-
-
-class ChildElementJob(ChildJob):
-    def __init__(self, *args, element, action_cb, **kwargs):
-        super().__init__(*args, **kwargs)
-        self._element = element
-        self._action_cb = action_cb
 
     def child_process(self):
 
--- a/src/buildstream/_scheduler/jobs/job.py
+++ b/src/buildstream/_scheduler/jobs/job.py
@@ -32,7 +32,7 @@ from ...types import FastEnum
 from ..._signals import TerminateException
 
 
-# Return code values shutdown of job handling child processes
+# Return code values of child tasks of a job
 #
 class _ReturnCode(FastEnum):
     OK = 0
@@ -61,25 +61,8 @@ class JobStatus(FastEnum):
 # Job()
 #
 # The Job object represents a task that will run in parallel to the main
-# process. It has some methods that are not implemented - they are meant for
+# thread. It has some methods that are not implemented - they are meant for
 # you to implement in a subclass.
-#
-# It has a close relationship with the ChildJob class, and it can be considered
-# a two part solution:
-#
-# 1. A Job instance, which will create a ChildJob instance and arrange for
-#    childjob.child_process() to be executed in another process.
-# 2. The created ChildJob instance, which does the actual work.
-#
-# This split makes it clear what data is passed to the other process and what
-# is executed in which process.
-#
-# To set up a minimal new kind of Job, e.g. YourJob:
-#
-# 1. Create a YourJob class, inheriting from Job.
-# 2. Create a YourChildJob class, inheriting from ChildJob.
-# 3. Implement YourJob.create_child_job() and YourJob.parent_complete().
-# 4. Implement YourChildJob.child_process().
 #
 # Args:
 #    scheduler (Scheduler): The scheduler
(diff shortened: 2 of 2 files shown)


src/buildstream/_scheduler/job.py
```diff
@@ -67,43 +67,57 @@ class JobStatus(FastEnum):
...
+#
+#     Returns:
+#        (object): Any abstract simple python object, including a string, int,
+#                  bool, list or dict, this must be a simple serializable object.
+#
+# Here is the calling signature of the complete_cb:
+#
+#     complete_cb():
+#
+#     This function will be called when the child task completes
```
juergbi:
```suggestion
#     This function will be called in the main thread when the child task completes
```

src/buildstream/_scheduler/job.py
```diff
@@ -322,111 +271,23 @@ async def _parent_child_completed(self, returncode):
         else:
             status = JobStatus.FAIL
 
-        self.parent_complete(status, self._result)
+        self._complete_cb(self, self._element, status, self._result)
```
juergbi: I'd suggest renaming the `_parent_child_completed()` method as well. Especially the 'parent' term no longer makes sense. Also the doc comments should be updated to refer to threads, not processes.

src/buildstream/_scheduler/job.py
```diff
@@ -27,12 +27,12 @@
...
-from ..._signals import TerminateException
+from .. import utils
+from .._utils import terminate_thread
+from .._exceptions import BstError, set_last_task_error, SkipJob
+from .._message import Message, MessageType
+from ..types import FastEnum
+from .._signals import TerminateException
 
 
 # Return code values shutdown of job handling child processes
```
juergbi: There are still a few lines mentioning processes instead of threads. Let's fix those while we're at it.

src/buildstream/_scheduler/job.py
```diff
@@ -67,43 +67,57 @@ class JobStatus(FastEnum):
 # process. It has some methods that are not implemented - they are meant for
 # you to implement in a subclass.
```
juergbi: There are no subclasses anymore.

juergbi (approved):
This looks good to me overall. `ChildJob` is indeed no longer useful. Assuming there are no plans for non-Element jobs, the second commit is a nice simplification as well.

gtristan:
In general I'm not fond of this refactor.

The reason why `ElementJob` is separate from the abstract `Job` class is because not all jobs are necessarily related to processing an element in the pipeline. Originally we had the *cache cleanup* job which was an example of this but has since been removed.

The way the scheduler manages *resources* (as in "processing", "downloading", "uploading") for which the user configuration can specify how much of these are allowed to occur in parallel is an API promise, I would say that any activity which buildstream is potentially processing and parallelizing should be handled by the scheduler and should be adhering to these resource restrictions.

While it is true that we do not *currently* have a non element processing job type, I'm not convinced that we will never have one, and it worries me to not have this design already sorted out and in place in case that other activities present themselves in the future (the danger being that implementors just bypass the scheduler and implement their own threads without conforming to the job pattern).

I should add that at the time we added the cache cleanup job, it was *surprisingly difficult* to ensure that the cache cleanup feature was integrated in cooperation with the scheduler nicely with this Job/ElementJob separation, it would be a shame to need to go through that a second time (or end up with brute force implementations of non-element job processing).

gtristan:
> In general I'm not fond of this refactor.

I had further conversation with @juergbi on this and it's better to update the issue.

Main points are:
* I'd like to keep the separation of `Job` and `ElementJob` for reasons stated above
* I agree that the simplification of `job.py` and elimination of the separate `ChildJob` object is desirable, as I believe the first patch in this branch achieves
  * I'm a little concerned about this patch, just because it appears to touch on task termination logic. BuildStream suspension and termination, especially in interactive scenarios, is really not very well covered in the test suite, so I would just appreciate that some creative interactive testing be done to avoid regressions on this front.

abderrahim:
>     * I'd like to keep the separation of `Job` and `ElementJob` for reasons stated above

The main reason I decided to merge `ElementJob` into job is the fact that it didn't do much: it just replaced the abstract method to override with a callback function, and everything else was using the callbacks instead of subclassing.

It is true that there is an `Element` specific part in it, namely the fact that an `Element` is passed to the constructor (and subsequently to the callback functions). But the `Element`-specific logging stuff is already in the parent `Job` class (exposed to subclasses via protected methods).

Would you be open to reworking this patch such that the `element` passed to the job can be `None` (and even have `None` default value), and document how to make a `Job` without an `Element`?

>     * I agree that the simplification of `job.py` and elimination of the separate `ChildJob` object is desirable, as I believe the first patch in this branch achieves

Ack.

>       * I'm a little concerned about this patch, just because it appears to touch on task termination logic. BuildStream suspension and termination, especially in interactive scenarios, is really not very well covered in the test suite, so I would just appreciate that some creative interactive testing be done to avoid regressions on this front.

It doesn't. The only change it does is a cut-and-paste of `ChildJob.terminate()` into where it was called in `Job.terminate()`, and this is not a big function or anything.

And to be honest, this was my main motivation for touching this part of the code. I'd like to try to fix #1613, and having this back and forth between two classes isn't helping me understand.

I appreciate that this is a delicate problem (as we already have #1613 and another instance where we get a BUG message when terminating), but I think this PR isn't touching that and shouldn't be blocked by testing.

gtristan:
Long time no reply here sorry.

> Would you be open to reworking this patch such that the element passed to the job can be None (and even have None default value), and document how to make a Job without an Element?

I'd really rather keep all the boiler plate in place which clearly separates the `ElementJob` as a specialized concrete class of the abstract `Job` class.

I know that the `Job` class has some unfortunate knowledge about elements, which is mostly related to logging context, but still it would be great to keep these separate.

src/buildstream/_scheduler/job.py
abderrahim: done

src/buildstream/_scheduler/job.py
abderrahim: I skipped this as after removing the element job, we're still using subclasses and overrides. Let's do the rename later

src/buildstream/_scheduler/job.py
abderrahim: I fixed those that I found.

src/buildstream/_scheduler/job.py
abderrahim: There are now that I reverted the second part.

PR #1707, opened 2022-07-31
Cache full build tree for debugging failed builds

### LINKED ISSUE #1263: Caching buildtrees only works with BuildElement (cannot enter a shell after a failed build)
[See original issue on GitLab](https://gitlab.com/BuildStream/buildstream/-/issues/1263)
In GitLab by [[Gitlab user @abderrahimk]](https://gitlab.com/abderrahimk) on Feb 15, 2020, 14:16

## Summary

When I try to enter a shell on a build failure, I get an error instead.

## Steps to reproduce

- default configuration of buildstream
- try to build an element that fails
- enter `s` at the prompt

## What is the current bug behavior?

```
Build failure on element: vm/repo.bst

Choose one of the following options:
  (c)ontinue  - Continue queueing jobs as much as possible
  (q)uit      - Exit after all ongoing jobs complete
  (t)erminate - Terminate any ongoing jobs and exit
  (r)etry     - Retry this job
  (l)og       - View the full log file
  (s)hell     - Drop into a shell in the failed build sandbox

Pressing ^C will terminate jobs and exit


Choice: [continue]: s

Dropping into an interactive shell in the failed build sandbox

Error while attempting to create interactive shell: Buildtree is not cached locally or in available remotes
```

a shell into the failed build.

## Other relevant information

* BuildStream version affected: 0591b0c8b038aa9f054b295748f3423384711d50

----

gtristan:
This change necessarily breaks cache keys and changes the artifact format.

Benefits of this change is essentially the ability to debug failed builds in non-build elements such as `script` and `compose` elements, which was agreed to be a blocker.

An additional benefit is that now one can download a failed build artifact and debug it in a shell *without* downloading any dependency artifacts.

Fixes #1263


gtristan:
This should allow `bst shell --build --use-buildtree` to work (in interactive mode or otherwise) on non-build elements like `script` elements.

This definitely breaks cache keys.

~~Need to add some tests for non-build elements.~~

Tests added for:
* `bst shell --build --use-buildtree` into a script element
* `bst shell --build --use-buildtree` into a compose element
* `bst shell --build --use-buildtree` into a compose element which failed it's integration commands

src/buildstream/_artifact.py
```diff
@@ -52,7 +52,7 @@
 #
 class Artifact:
 
-    version = 1
+    version = 2
```
wsalmonct, line 55: I had assumed that this value would match the version in src/buildstream/_protos/buildstream/v2/artifact.proto which clams to only need bumping when mandatory fields change.. are these in fact unrelated? or should the other be bumped also?

gtristan: The line:

```
  int32 version = 1;
```

Indicates that the *field number* in the artifact proto is 1, this number 1 will never change.

The comment above it indicates that it need not be incremented for an optional field, however given that buildstream requires this field *instead* of the buildtree for some operations, it's probably more prudent to just bump the field.

In any case, since we bump the `BST_ARTIFACT_VERSION` we will anyway be loading different artifacts.

PR #1710, opened 2022-08-02
doc/source/main_install.rst: Buildstream depends on lzip

jjardon:
This is a dependency of the core tar plugin

See https://github.com/apache/buildstream/pull/1705#issuecomment-1199046601

### DIFF
--- a/doc/source/main_install.rst
+++ b/doc/source/main_install.rst
@@ -25,6 +25,7 @@ BuildStream requires the following base system requirements:
 
 - python3 >= 3.7
 - pip
+- lzip
 - :ref:`buildbox-casd<install-buildbox>`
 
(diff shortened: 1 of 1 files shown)


gtristan:
I'm not really concerned about the outcome... whether we explicitly list `lzip` as a dependency or not but...

To clarify:

* BuildStream is perfectly capable of functioning without `lzip` installed, it's not required by any means
* If your BuildStream project uses a `.tar.lz` file, then the `tar` plugin will require `lzip` (this is an exceedingly rare case)
* If you try to run `BuildStream` to load an element that uses a `.tar.lz` file, it will exit early informing the user that `lzip` was not found on the host
  * So there is at least a user story for informing the user that they need `lzip` in that edge case

What do we prefer to do ?

PR #1717, opened 2022-08-12
Update pylint config and fix some trivial issues

ghost:
This enables `lint` test to pass with Pylint >= 2.14.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -80,14 +80,14 @@ confidence=
 #    pylint to also check for those things.
 #
 
-disable=#####################################
+disable=
+        #####################################
         # Messages that are of no use to us #
         #####################################
         ,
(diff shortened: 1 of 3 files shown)


ghost:
Seems this cannot work as-is with Pylint 2.13, as the `no-self-use` checks needs to be disabled there.

gtristan:
> Seems this cannot work as-is with Pylint 2.13, as the `no-self-use` checks needs to be disabled there.

Better not try to submit this change separately to the branch which updates the requirements.

ghost:
Yes, can merge into https://github.com/apache/buildstream/pull/1714 if that's acceptable

gtristan:
> Yes, can merge into #1714 if that's acceptable

Of course

PR #1719, opened 2022-08-15
Allow artifact names in bst shell

### LINKED ISSUE #1711: Support artifact names in `bst shell --build --use-buildtree`
With the addition of #1707, the dependencies of an artifact are not necessary to launch a build shell with a buildtree.

This it should be possible to launch a build shell directly on a downloaded artifact, without requiring a local copy of the project or elements.

It should be possible to debug a remotely failed build, automatically pulling the artifact a remote, with a command such as:
```
bst --pull-buildtrees \
    shell --build --use-buildtree --artifact-remote https://artifacts.org \
    project/element/329f4ff2c3855f541ff2f85d0ba5ad4a87c73610914f84483c8462b095a8007e
```

gtristan:
Now that we cache the full tree when builds fail, it becomes more interesting to have a buildshell on a failed artifact without needing to reproduce the project data.

Fixes #1711

### DIFF
--- a/tests/integration/shellbuildtrees.py
+++ b/tests/integration/shellbuildtrees.py
@@ -327,6 +327,40 @@ def test_shell_pull_cached_buildtree(share_with_buildtrees, datafiles, cli, pull
 
 
 #
+# Test behavior of shelling into a buildtree by its artifact name
+#
+@pytest.mark.datafiles(DATA_DIR)
+@pytest.mark.skipif(not HAVE_SANDBOX, reason="Only available with a functioning sandbox")
+def test_shell_pull_artifact_cached_buildtree(share_with_buildtrees, datafiles, cli):
+    project = str(datafiles)
+    artifact_name = "test/build-shell-buildtree/4a47c98a10df39e65e99d471f96edc5b58d4ea5b9b1f221d0be832a8124b8099"
+
+    cli.configure({"artifacts": {"servers": [{"url": share_with_buildtrees.repo}]}})
+
+    # Run the shell and request that required artifacts and buildtrees should be pulled
+    result = cli.run(
+        project=project,
+        args=[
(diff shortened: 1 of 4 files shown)


gtristan:
I've added some more testing to this branch (locally), as I was trying to see what happens when you try `bst shell <artifact-name>`, instead of using a build tree.

But I encountered:
```
Error launching shell: Error: '--deps run' is not supported for artifact names
```

This is correct because we don't encode runtime dependencies into artifacts (which necessary by design), so I conclude that we cannot support shelling into artifact names except for when running build shells.

I will add a more convenient user facing error in the case the user has attempted to launch a shell on an artifact name without specifying `--build` and `--use-buildtree`.

gtristan:
Updated, added a new test to ensure the correct error messaging when the user tries to use `bst shell` with artifact names but not using build trees.

```
$ bst shell test/autotools-amhello/847d23cd0e61c54a6beca9071b64643388ad2ab783191358b2a499fe77e86563 -- hello

Error launching shell: Only buildtrees are supported with artifact names

    Use the --build and --use-buildtree options to shell into a cached build tree
```

tests/integration/shellbuildtrees.py
```diff
@@ -326,6 +326,40 @@ def test_shell_pull_cached_buildtree(share_with_buildtrees, datafiles, cli, pull
...
+        project=project,
+        args=[
+            "--pull-buildtrees",
+            "shell",
+            "--build",
+            "--use-buildtree",
+            artifact_name,
+            "--",
+            "cat",
+            # We don't preserve the working directory in artifacts, so we will executing be at /
```
ghost: "so we will executing be at /" -- word order / typo ?

ghost (approved):
Looks fine to me.

PR #1721, opened 2022-08-16
doc/source/main_install.rst: Document where to obtain latest buildbox binaries

### LINKED ISSUE #1189: Check versions of buildbox

gtristan:
This effectively closes #1189

### DIFF
--- a/doc/source/main_install.rst
+++ b/doc/source/main_install.rst
@@ -84,12 +84,13 @@ These components can be installed from binaries, or built from source.
 
 Install binaries
 ~~~~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


gtristan:
This should be merged after merging https://gitlab.com/BuildGrid/buildbox/buildbox-integration/-/merge_requests/3 and pushing a tag so that the buildbox release is actually created.

PR #1723, opened 2022-08-17
Build and release wheel packages

### LINKED ISSUE #1712: Distribute buildstream Wheel packages that contain BuildBox
# Intro

Bst2 is available [in PyPI](https://pypi.org/project/BuildStream/), but only as a [source distribution](https://packaging.python.org/en/latest/specifications/source-distribution-format/). Many packages on PyPI can be installed simply with `pip3 install $package`. Installing BuildStream is not so simple:

  * A working `cc` or `gcc` must be available, along with CPython headers and libraries, so Cython modules can be compiled to binaries
  * BuildBox must be installed, this currently implies building from source (see: [buildbox issue about releasing binaries](https://gitlab.com/BuildGrid/buildbox/buildbox-integration/-/issues/1))

The Python [Wheel](https://peps.python.org/pep-0427/#installing-a-wheel-distribution-1-0-py32-none-any-whl) format can be used to bundle and distribute binaries for our target platforms. When you run `pip3 install buildstream` and a suitable `buildstream-*.whl` is available on PyPI, the install should work even in a 'stock' container image:

  * Cython modules are prebuilt so GCC isn't needed
  * BuildBox binaries can be bundled in the wheel and installed in '$prefix/bin/' by Pip

# Goals

The goal is a way to install BuildStream 2 in a single command on a wide range of machines, assuming:

  * the OS doesn't provide its own BuildStream package (we prefer distro packages where available)
  * the user trusts PyPI and the BuildStream release team

Non goals:

  * replace distro builds. Anyone redistributing BuildStream should use the sdist or Git repo and build from source.

# What about Docker?

The [Bst 1.6 install instructions](https://docs.buildstream.build/1.6/install_docker.html) recommend running in Docker via [bst-here](https://gitlab.com/BuildStream/buildstream/blob/master/contrib/bst-here). The [Bst 1.9.95 install instructions](https://docs.buildstream.build/1.95/main_install.html#install-container) also link to the [buildstream-docker-images instructions](https://gitlab.com/BuildStream/buildstream-docker-images/-/blob/master/USING.md). (At time of writing the [buildstream.build website](https://buildstream.build/install.html) doesn't mention Docker directly).

Provisioning Bst via a container image is sensible for many situations and we will always ensure it is possible. **Recommending it as the best way to use BuildStream** is more nuanced.

Advantages of recommending users obtain Bst via Docker:

  * Single-command setup that "just works" for most cases
  * We already build and publish usable Docker images at https://gitlab.com/BuildStream/buildstream-docker-images
  
Disadvantages of recommending users obtain Bst via Docker:

  * Dependency on a container host (we use hub.docker.io at the moment, but could use Github container hosting)
  * Requirement that user can start a container - not easy if they're already in a container, or don't have the necessary privileges
  * Builds have different requirements for host tools depending on what plugins are used, so there's no 'one size fits all' container
  * Most build tools don't require a separate container to run, so this makes `bst` look like a 'heavyweight' tool
  * Docker images are larger than wheels

gtristan:
This updates the CI config to:

  * build x86_64 wheel packages on each update of 'master', which can be downloaded as Action artifacts. The packages contain release binaries of BuildBox casd/fuse/run-bubblewrap.
  * build sdist and wheel packages on each release tag, test the wheels, and upload them to Test PyPI at https://test.pypi.org/project/BuildStream/

Before actually landing this, we need to make some changes:

  * [x] update documentation around release process and installation (https://gitlab.com/BuildStream/website/-/merge_requests/153)
  * [x] release to the real PyPI instead of [Test PyPI](https://test.pypi.org/) (in release.yml)
  * [x] create and add a [PyPI API token](https://pypi.org/help/#apitoken) as a [repository secret](https://docs.github.com/en/codespaces/managing-codespaces-for-your-organization/managing-encrypted-secrets-for-your-repository-and-organization-for-github-codespaces)
  * [x] review and merge latest [buildbox-fuse patches](https://gitlab.com/BuildGrid/buildbox/buildbox-fuse/-/merge_requests/46) so we can have a new buildbox release
  * [x] tag [buildbox-integration](https://gitlab.com/BuildGrid/buildbox/buildbox-integration/) so that we can have the binaries uploaded with the [new descriptive filenames](https://gitlab.com/BuildGrid/buildbox/buildbox-integration/-/merge_requests/5)
  * [x] apply new patch 85a8953940ee8a03d85b0277457583b67acd1656 to update the permalinks for downloading buildbox binaries

Fixes #1712

### DIFF
--- a/.github/compose/ci.docker-compose.yml
+++ b/.github/compose/ci.docker-compose.yml
@@ -85,3 +85,27 @@ services:
   mypy:
     <<: *tests-template
     command: tox -e mypy
+
+  # Test that each BuildStream+BuildBox wheel package can install and run.
+  # on the PyPA official 'manylinux' images that define the base ABI for
+  # Python binary packages.
+
+  wheels-manylinux_2_28-cp37:
+    <<: *tests-template
+    image: quay.io/pypa/manylinux_2_28_x86_64
+    command: .github/wheel-helpers/test-wheel-manylinux.sh cp37-cp37m-manylinux_2_28_x86_64 /opt/python/cp37-cp37m/bin/python3
+
+  wheels-manylinux_2_28-cp38:
+    <<: *tests-template
+    image: quay.io/pypa/manylinux_2_28_x86_64
+    command: .github/wheel-helpers/test-wheel-manylinux.sh cp38-cp38-manylinux_2_28_x86_64 /opt/python/cp38-cp38/bin/python3
+
+  wheels-manylinux_2_28-cp39:
+    <<: *tests-template
+    image: quay.io/pypa/manylinux_2_28_x86_64
+    command: .github/wheel-helpers/test-wheel-manylinux.sh cp39-cp39-manylinux_2_28_x86_64 /opt/python/cp39-cp39/bin/python3
+
+  wheels-manylinux_2_28-cp310:
+    <<: *tests-template
+    image: quay.io/pypa/manylinux_2_28_x86_64
+    command: .github/wheel-helpers/test-wheel-manylinux.sh cp310-cp310-manylinux_2_28_x86_64 /opt/python/cp310-cp310/bin/python3
--- a/.github/wheel-helpers/fetch-latest-buildbox-release.sh
+++ b/.github/wheel-helpers/fetch-latest-buildbox-release.sh
@@ -6,10 +6,15 @@
 
 set -eux
 
-curl -L -O https://gitlab.com/buildgrid/buildbox/buildbox-integration/-/releases/permalink/latest/downloads/binaries.tgz
+#
+# For now we only support building wheels for linux x86_64 linked against glibc
+#
+tarball="buildbox-x86_64-linux-gnu.tgz"
+
+curl -L -O "https://gitlab.com/buildgrid/buildbox/buildbox-integration/-/releases/permalink/latest/downloads/${tarball}"
 
 mkdir -p src/buildstream/subprojects/buildbox
-tar --extract --file ./binaries.tgz --directory src/buildstream/subprojects/buildbox
+tar --extract --file "./${tarball}" --directory src/buildstream/subprojects/buildbox
 
 cd src/buildstream/subprojects/buildbox
 rm buildbox-run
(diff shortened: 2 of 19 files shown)


gtristan:
This branch is a cleanup of #1714 and addresses some of the last review comments (because @ssssam is on vacation for the rest of the week).

This effectively replaces #1714

gtristan:
We recently changed the name of the binaries generated by `buildbox-integration` in https://gitlab.com/BuildGrid/buildbox/buildbox-integration/-/merge_requests/5

So, we are currently waiting for:

* A new tag on `buildbox-integration` repo, so that the new binary name gets published
  * Someone please review https://gitlab.com/BuildGrid/buildbox/buildbox-fuse/-/merge_requests/46 so that this can happen now
* Me to apply the local change I'm holding onto which changes the binary permalink we download binaries from
* A pipeline to pass here with the new change applied

And then we can finally release and tag and see automatically generated and published wheels which include buildbox.

ghost:
The generated wheels cannot build gnome-build-meta due to https://github.com/apache/buildstream/issues/1733. We should resolve that before final release.

ghost:
Here's a patch that fixes buildbox-casd operation with minimal churn: https://github.com/apache/buildstream/commit/c529c615f2dbfb216d03e66c315114e5bc77b3fa

We could alternatively simplify this branch by removing all the _get_host_tool_internal() helpers and setting `os.environ["PATH"]` on startup to include `os.path.join(_site.subprojects, "buildbox")`.

gtristan:
> Here's a patch that fixes buildbox-casd operation with minimal churn: [c529c61](https://github.com/apache/buildstream/commit/c529c615f2dbfb216d03e66c315114e5bc77b3fa)
> 
> We could alternatively simplify this branch by removing all the _get_host_tool_internal() helpers and setting `os.environ["PATH"]` on startup to include `os.path.join(_site.subprojects, "buildbox")`.

I've applied the patch to this branch thanks.

I think I also like the patch this way because the `PATH` setting resides in codepaths which are clearly tied to launching buildbox binaries, whereas globally setting `PATH` at startup would have to live in a disconnected/unrelated codepath.

Global environment setting was unavoidable for setting `GRPC_ENABLE_FORK_SUPPORT` in cfe3d71c71 (which is unfortunatly confusing since the generated entrypoint is nowhere close to grpc usage/initialization).

PR #1725, opened 2022-08-18
_artifactcache/cascache.py: Only check push capabilities for push remotes

### LINKED ISSUE #1724: [bst-1] checks for push in a pull only remote
bst 1 tries to check whether the remote supports asset push even if the remote is not configured for push. This combined with a grpcio bug (that produces weird errors when getting http 404) means that initializing remotes fails with an error like

```
[--:--:--][][] WARNING Failed to initialize remote https://cache.freedesktop-sdk.io:11001: Stream removed
```

gtristan:
Also, handle permission denied response as an unsupported PushDirectory().

Fixes #1724

### DIFF
--- a/buildstream/_artifactcache/cascache.py
+++ b/buildstream/_artifactcache/cascache.py
@@ -1118,19 +1118,6 @@ class CASRemote():
                 if e.code() != grpc.StatusCode.UNIMPLEMENTED:
                     raise
 
-            # Check whether the server supports BatchUpdateBlobs()
-            self.batch_update_supported = False
-            try:
-                request = remote_execution_pb2.BatchUpdateBlobsRequest()
-                for attempt in _retry():
-                    with attempt:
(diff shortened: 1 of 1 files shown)


abderrahim:
While this is a good thing to have, I think it doesn't fix #1724.

In the case of #1724, we still have an HTTP 404 error which should map to UNIMPLEMENTED. For some reason (likely a bug in grpcio) it is being mapped to some other error (the "stream removed" thing)

gtristan:
> While this is a good thing to have, I think it doesn't fix #1724.
> 
> In the case of #1724, we still have an HTTP 404 error which should map to UNIMPLEMENTED. For some reason (likely a bug in grpcio) it is being mapped to some other error (the "stream removed" thing)

I've updated the patch to additionally only check for these push capabilities on push remotes.

abderrahim (approved):
LGTM. I'll do more tests later, but it fixes the immediate issue.

PR #1727, opened 2022-08-19
Continue post processing when quitting the scheduler

### LINKED ISSUE #534: Failed builds not pushing artifacts on quit
[See original issue on GitLab](https://gitlab.com/BuildStream/buildstream/-/issues/534)
In GitLab by [[Gitlab user @knownexus]](https://gitlab.com/knownexus) on Aug 1, 2018, 10:12

## Summary

[//]: # (Summarize the bug encountered concisely)

If, when a build fails, the user quits instead of continuing, the artifact will not be pushed

## Steps to reproduce

[//]: # (How one can reproduce the issue - this is very important)
Build a project that fails,

When given the options of (t)terminate, (q)quit, (c)continue

select `q`

Check for the artifact in the cache

The artifact should be missing

[//]: # (What you should see instead)
The failed build artifact should be cached even when the quit option is selected

## Possible fixes

[//]: # (If you can, link to the line of code that might be responsible for the problem)
One option is to have the artifact be cached at the point of the build failing, so at the point of choosing whether or not to continue, it is already cached

## Other relevant information

* BuildStream version affected: /milestone %BuildStream_v1.4

----

gtristan:
This changes the meaning of `--on-error quit` such that post-processing of the work most relevant to the command issued continues to happen after a failure occurs.

Fixes #534

### DIFF
--- a/src/buildstream/_scheduler/queues/artifactpushqueue.py
+++ b/src/buildstream/_scheduler/queues/artifactpushqueue.py
@@ -31,8 +31,8 @@ class ArtifactPushQueue(Queue):
     complete_name = "Artifacts Pushed"
     resources = [ResourceType.UPLOAD]
 
-    def __init__(self, scheduler, *, skip_uncached=False):
-        super().__init__(scheduler)
+    def __init__(self, scheduler, *, imperative=False, skip_uncached=False):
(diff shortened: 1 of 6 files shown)


juergbi (approved):
Looks good to me after fixing the remaining test failure.

PR #1728, opened 2022-08-20
_frontend/app.py: Fixed missing messages in frontend main logging

gtristan:
The threaded scheduler causes the frontend message handling to be
multi-threaded, as messages are no longer passed through a message queue
but directly propagated to the frondend.

This, in conjunction with message throttling means that we need to use
a lock to protect the cached message, otherwise sometimes a message gets
cached at the same time we finish printing a message and clear the cache.

This patch also ensures that we flush the message cache more consistently
when exiting the scheduler, and causes local app.py messages to be flushed
immediately just in case.

This fixes the spurious error we've been seeing with tests/frontend/logging.py::test_failed_build_listing().

### DIFF
--- a/src/buildstream/_frontend/app.py
+++ b/src/buildstream/_frontend/app.py
@@ -19,6 +19,7 @@
 from contextlib import contextmanager
 import os
 import sys
+import threading
 import traceback
 import datetime
 from textwrap import TextWrapper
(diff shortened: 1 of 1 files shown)


gtristan:
Example of a pipeline where the logging test spuriously failed: https://github.com/apache/buildstream/runs/7929845153?check_suite_focus=true

PR #1734, opened 2022-08-23
buildelement.py: Documentation fixes and enhancements

gtristan:
o Fix links to elements which no longer reside in BuildStream

 o Provide an example of running a srcdir != builddir build with autotools

 o Reworded some things a bit more clearly with regards to `command-subdir`,
   `conf-root` and `install-root`

### DIFF
--- a/src/buildstream/buildelement.py
+++ b/src/buildstream/buildelement.py
@@ -81,35 +81,58 @@ an alternative location while staging some elements in the sandbox root.
 
 Location for running commands
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-The ``command-subdir`` variable sets where the build commands will be executed,
-if the directory does not exist it will be created, it is defined relative to
-the buildroot.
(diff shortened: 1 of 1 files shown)


src/buildstream/buildelement.py
```diff
@@ -81,35 +81,58 @@
...
+This can be achieved in build elements which use ``conf-root`` as follows:
+
+.. code:: yaml
+
+   variables:
+     # Specify that build configuration scripts are found in %{build-root}
+     conf-root: "%{build-root}"
+
+     # The build will run in the `_build` subdirectory
+     command-subdir: _build
```
nanonyme, line 121: Does this make any sense in BuildStream core? cmake and meson elements do not do it like this, instead they have build-dir variable which is passed to build system.

gtristan: It is a more concrete and practical example of using conf-root and command-subdir which were added for this purpose specifically.

Compared to the confusing text before this is an improvement.

A similar example could be added to the autotools plugin docs, though.

PR #1736, opened 2022-08-23
.github/workflows/merge.yml: Fix up job dependency after rename

ghost:
Fixes the following error in "Merge actions" workflow:

    The workflow is not valid. .github/workflows/merge.yml (Line: 46, Col: 12): Job 'publish_docs' depends on unknown job 'build'.

This is a regression introduced in https://github.com/apache/buildstream/pull/1723

### DIFF
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -43,7 +43,7 @@ jobs:
           docs.tgz
 
   publish_docs:
-    needs: build
+    needs: build_docs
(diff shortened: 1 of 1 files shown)


ghost:
Since Github seems to only validate the workflow post-merge, I have merged this commit to master in my fork to show that it is fixed: https://github.com/ssssam/buildstream/actions/runs/2912720130

gtristan (approved):

PR #1737, opened 2022-08-23
.github/workflows/release.yml: Fix names of 'test_wheel' jobs

ghost:
The 'test_wheel' job is defined twice, in ci.yml and release.yml. This
copies a small fix from ci.yml to release.yml so the jobs have useful
names.


ghost:
Fix under test in my fork: https://github.com/ssssam/buildstream/actions/runs/2912766727

PR #1738, opened 2022-08-23
doc/source/hacking: Document wheel building infrastructure

ghost:
There is a lot to remember in terms of wheel building, this is an
attempt to write it all down before I forget.

### DIFF
--- a/doc/source/hacking/updating_python_deps.rst
+++ b/doc/source/hacking/updating_python_deps.rst
@@ -20,3 +20,89 @@ changing the ``.in`` file, run the following to update the matching ``.txt``
 file::
 
    make -C requirements
+
+Adding support for a new Python release
+---------------------------------------
+
+When a new stable release of Python 3 appears, we must explicitly declare
+our support for it in the following places.
+
+tox.ini
+~~~~~~~
+
+The ``tox.ini`` file defines the environments where the BuildStream test suite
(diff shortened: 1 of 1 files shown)


jjardon:
This looks good to me: @gtristan @abderrahim @juergbi  ok to merge?

doc/source/hacking/updating_python_deps.rst
```diff
@@ -20,3 +20,89 @@ changing the ``.in`` file, run the following to update the matching ``.txt``
...
+~~~~~~~
+
+The ``tox.ini`` file defines the environments where the BuildStream test suite
+runs.  Every ``py{3.x,3.y}`` list must be updated to contain the new version
+number such as ``311`` for CPython 3.11.
+
+Use ``tox -e py311-nocover`` to run the test suite with the new version of
+Python.
+
+pyproject.toml
```
gtristan, line 40: Good addition to the hacking !

note that here we also have to bump the minimum required version of cython which includes support for the new cpython version.

ghost: i see, and i assume this is a change request so will push an update when I have a chance

gtristan: I didn’t notice how old this was, don’t worry I’ll do it :)

PR #1741, opened 2022-08-25
Update install instructions for final BuildStream 2.0 release

ghost:
See commits for rationale for the various changes.

### DIFF
--- a/doc/source/main_install.rst
+++ b/doc/source/main_install.rst
@@ -224,15 +224,16 @@ Installing completions
 ----------------------
 
 BuildStream integrates with Bash to provide helpful tab-completion. These
-are provided by the script ``src/buildstream/data/bst`` which requires
-separate installation.
+are provided by the `bst` completion script, available online
+`here <https://raw.githubusercontent.com/apache/buildstream/master/src/buildstream/data/bst>`_
+and in your local Git clone at ``src/buildstream/data/bst``. The completion
+script requires manaul installation.
 
-To install for the current user, copy it into
-``share/bash-completion/completions/`` inside the user's ``$XDG_DATA_HOME``
-directory (the path is usually: ``$HOME/.local/share/bash-completion/completions``).
+To install for the current user, paste the contents of the completion script
+into the file ``~/.bash_completion``.
 
-To install completions system-wide, get the installation path as follows::
-
+To install completions system-wide, copy the completion script to the system-wide
+bash-completion installation path, which you can discover as follows::
 
     pkg-config --variable=completionsdir bash-completion
 
--- a/doc/source/main_install.rst
+++ b/doc/source/main_install.rst
@@ -153,9 +153,8 @@ required PyPI dependencies from the internet at the same time::
 
     pip3 install --user .
 
-You can optionally use Pip's
-`editable mode <https://pip.pypa.io/en/stable/topics/local-project-installs/#editable-installs>`_
-(the ``-e`` flag) as a developer convenience.
+We do not recommend using Pip's `editable mode <https://pip.pypa.io/en/stable/topics/local-project-installs/#editable-installs>`_
+(the ``-e`` flag). See `this issue <https://github.com/apache/buildstream/issues/1760>`_ for discussion.
 
 If you want to stop Pip from fetching missing dependencies, use the
 ``--no-index`` and ``--no-deps`` options.
--- a/doc/source/main_install.rst
+++ b/doc/source/main_install.rst
@@ -24,7 +24,7 @@ BuildStream requires the following Python environment to run:
 
 - python3 >= 3.7
 - PyPI packages as specified in
-  `requirements.txt <https://github.com/apache/buildstream/blob/master/requirements/requirements.txt>`_.
+  `requirements.in <https://github.com/apache/buildstream/blob/master/requirements/requirements.in>`_.
 
 Some :mod:`Source <buildstream.source>` plugins require specific tools installed
 on the host. Here is a commonly used subset based on the
--- a/doc/source/main_install.rst
+++ b/doc/source/main_install.rst
@@ -1,96 +1,104 @@
-Installing
-==========
+Installing from Source
+======================
 
-Until BuildStream is available in your distro, you may need to install
-it yourself from source. The necessary steps are:
+This page explains how to build and install this version of BuildStream from
+source. For general purpose installation instructions consult the
+`website <https://buildstream.build/install.html>`_.
+
+For full install instructions, read on:
 
 * :ref:`Install dependencies<install-dependencies>`
 * :ref:`Install BuildBox<install-buildbox>`
-* :ref:`Install BuildStream<install-buildstream>` (from a git checkout, or from PyPi)
-* :ref:`Update PATH<post-install>`
-
-Alternatively, BuildStream can be run in :ref:`a container<install-container>`.
-
+* :ref:`Install BuildStream<install-buildstream>`
+* :ref:`Install completions<install-completions>`
 
 .. _install-dependencies:
 
 Installing Dependencies
 -----------------------
 
-Before installing BuildStream from source, it is necessary to first install
-the system dependencies. Below are some linux distribution specific instructions
-for installing these dependencies.
+Runtime requirements
+~~~~~~~~~~~~~~~~~~~~
 
-BuildStream requires the following base system requirements:
+BuildStream requires the following Python environment to run:
 
 - python3 >= 3.7
-- pip
-- lzip (optional, for ``.tar.lz`` support)
-- :ref:`buildbox-casd<install-buildbox>`
+- PyPI packages as specified in
+  `requirements.txt <https://github.com/apache/buildstream/blob/master/requirements/requirements.txt>`_.
 
-BuildStream also depends on the host tools for the :mod:`Source <buildstream.source>` plugins.
-Refer to the respective :ref:`source plugin <plugins_sources>` documentation for host tool
-requirements of specific plugins.
+Some :mod:`Source <buildstream.source>` plugins require specific tools installed
+on the host. Here is a commonly used subset based on the
+:ref:`core source plugins <plugins_sources>`
+and `buildstream-plugins <https://apache.github.io/buildstream-plugins/>`_.
 
+- git (for ``git`` sources)
+- lzip (for ``.tar.lz`` support in ``tar`` sources)
+- patch (for ``patch`` sources)
 
-Arch Linux
-~~~~~~~~~~
-Install the dependencies with::
+Some BuildBox tools used by BuildStream require additional host tools:
 
+- bubblewrap (for ``buildbox-run-bubblewrap``)
+- fusermount3 (for ``buildbox-fuse``)
 
-    sudo pacman -S python python-pip
+Install-time requirements
+~~~~~~~~~~~~~~~~~~~~~~~~~
 
+BuildStream contains Cython code which implies the following extra
+dependencies at install-time only:
 
-Debian
-~~~~~~
-Install the dependencies with::
+- C and C++ toolchain
+- Python development headers
 
+These instructions use ``pip3`` to install necessary PyPI packages.
+Packagers and integrators may use a different tool and can ignore
+the `pip` dependency below.
 
-    sudo apt-get install \
-        python3 python3-pip python3-dev
+Distribution-specific guides
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+This table gives you a list of packages for specific distros:
 
-Fedora
-~~~~~~
-For recent fedora systems, the following line should get you the system
-requirements you need::
+.. list-table::
 
-
-    dnf install -y \
-        python3 python3-pip python3-devel
-
-
-Ubuntu
-~~~~~~
-Install the dependencies with::
-
-
-    sudo apt install \
-        python3 python3-pip python3-dev
-
+  * - **Distribution**
+    - **Runtime requires**
+    - **Install requires**
+  * - Arch Linux
+    - bubblewrap fuse3 git lzip patch python
+    - gcc python-pip
+  * - Debian
+    - bubblewrap fuse3 git lzip patch python3
+    - g++ python3-dev python3-pip
+  * - Fedora
+    - bubblewrap fuse3 git lzip patch python3
+    - gcc-c++ python3-devel python3-pip
+  * - Ubuntu
+    - bubblewrap fuse3 git lzip patch python3
+    - g++ python3-dev python3-pip
 
 .. _install-buildbox:
 
 Installing BuildBox
 -------------------
 
-BuildStream master now depends on buildbox-casd to manage the local CAS cache
-and communicate with CAS servers. buildbox-run is used for sandboxing. BuildBox
-components are still in development and there are no stable releases yet.
-Thus, they're not available yet in Linux distros and they have to be manually
-installed.
+BuildStream depends on the following tools from
+`BuildBox <https://gitlab.com/BuildGrid/buildbox/>`_:
+
+  * ``buildbox-casd`` (to manage local and remote content-addressed storage)
+  * ``buildbox-fuse`` (to check out content from the local CAS)
+  * ``buildbox-run-bubblewrap`` (to run element commands in a controlled sandbox)
 
 These components can be installed from binaries, or built from source.
 
 Install binaries
 ~~~~~~~~~~~~~~~~
-Linux x86-64 users can download the `latest statically linked binaries here
-<https://gitlab.com/BuildGrid/buildbox/buildbox-integration/-/releases/permalink/latest/downloads/binaries.tgz>`_,
-or browse the `release history of static binaries here
+Browse the `release history of static binaries here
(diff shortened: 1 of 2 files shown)


doc/source/main_install.rst
```diff
@@ -285,13 +327,23 @@ A regular way to do this is to add the following line to the end of your ``~/.ba
...
+`here <https://gitlab.com/BuildStream/buildstream-docker-images/-/blob/master/USING.md>`_
+
+You can also produce your own container images, either by adapting the
+`buildstream-docker-images project <https://gitlab.com/BuildStream/buildstream-docker-images/>`_,
+or by following the full installation instructions above.
+
+Note that some special configuration is often needed to run BuildStream in a container:
+
+  * User namespaces are used to isolate and control builds. This requires the
+    Docker ``--privileged`` mode.
```
nanonyme: With podman at least this is not required. We (freedesktop-sdk) have an image that works *just fine* in unprivileged toolbox under user namespaces.

ghost: I suspect Toolbx containers are `--privileged` by default. I had a quick look at the code, and it certainly passes `--privileged` to Podman in a couple of places. User namespaces require at least CAP_SYS_ADMIN, and maybe more capabilities that your random "run container for free" CI service often won't allow

nanonyme: Looks to me it's `podman --log-level error exec --detach-keys  --interactive --tty --user username --workdir /var/home/username/git/freedesktop-sdk --env=COLORTERM=truecolor --env=DBUS_SESSION_BUS_ADDRESS=unix:path=/run/user/1000/bus --env=DESKTOP_SESSION=gnome --env=DISPLAY=:0 --env=LANG=en_US.UTF-8 --env=SHELL=/bin/bash --env=SSH_AUTH_SOCK=/run/user/1000/keyring/ssh --env=TERM=xterm-256color --env=TOOLBOX_PATH=/usr/bin/toolbox --env=USER=username --env=VTE_VERSION=6800 --env=WAYLAND_DISPLAY=wayland-0 --env=XAUTHORITY=/run/user/1000/.mutter-Xwaylandauth.NU99Q1 --env=XDG_CURRENT_DESKTOP=GNOME --env=XDG_DATA_DIRS=/var/home/username/.local/share/flatpak/exports/share:/var/lib/flatpak/exports/share:/usr/local/share/:/usr/share/ --env=XDG_MENU_PREFIX=gnome- --env=XDG_RUNTIME_DIR=/run/user/1000 --env=XDG_SESSION_DESKTOP=gnome --env=XDG_SESSION_TYPE=wayland bst2 capsh --caps= -- -c exec "$@" /bin/sh /bin/bash -l` more or less.

nanonyme: Right, so --privileged I think means different things here. With Docker it means "you can privilege escalate to root", with podman it means "you can privilege escalate to user".

nanonyme: The whole point of user namespaces is they work without privileges.

ghost: I will be happy to be wrong here, but I want to test a bit more to be sure.

I will try a Bst build in one of the FDSDK docker images and see if that works without anything more than `--device /dev/fuse`.

nanonyme: You can check with `registry.gitlab.com/freedesktop-sdk/infrastructure/freedesktop-sdk-docker-images/bst2:latest`

ghost: I didn't get as far as creating namespaces as it turns out `fusermount3` can't mount:

```
> podman run --device /dev/fuse -i -t --rm registry.gitlab.com/freedesktop-sdk/infrastructure/freedesktop-sdk-docker-images/bst2
[root@de014e636534 /]# bst --version
1.95.2.dev0
[root@de014e636534 /]# cd
[root@de014e636534 ~]# git clone https://github.com/apache/buildstream
[root@de014e636534 buildstream]# cd doc/examples/running-commands/
[root@de014e636534 running-commands]# bst build hello.bst
...

Failure Summary
    hello.bst:
    [00:00:00][8cebf405][   build:hello.bst                     ] FAILURE buildbox-run failed with returncode 1

        Printing the last 20 lines from log file:
        /root/.cache/buildstream/logs/running-commands/hello/8cebf405-build.110.log
        ======================================================================
            LC_ALL: C
            HOME: /tmp
            TZ: UTC
            SOURCE_DATE_EPOCH: 1321009871
        [--:--:--] START   [8cebf405] hello.bst: Staging dependencies at: /
        [--:--:--] STATUS  [b59f4520] base/alpine.bst: Staging base/alpine.bst/b59f4520
        [--:--:--] STATUS  [bb393e5a] base.bst: Staging base.bst/bb393e5a
        [00:00:00] SUCCESS [8cebf405] hello.bst: Staging dependencies at: /
        [--:--:--] START   [8cebf405] hello.bst: Integrating sandbox
        [00:00:00] SUCCESS [8cebf405] hello.bst: Integrating sandbox
        [--:--:--] START   [8cebf405] hello.bst: Staging sources
        [00:00:00] SUCCESS [8cebf405] hello.bst: Staging sources
        [--:--:--] START   hello.bst: Running commands

            make PREFIX="/usr"
            make -j1 PREFIX="/usr" DESTDIR="/buildstream-install" install
        2022-08-25T15:20:08.005+0000 [191:139995975073728] [buildboxcommon_casclient.cpp:95] [INFO] Setting d_maxBatchTotalSizeBytes = 4128768 bytes by default
        2022-08-25T15:20:08.155+0000 [191:139995975073728] [buildboxcommon_runner.cpp:437] [ERROR] [actionDigest=ace830804d554a11f87ee8018bde83888d28775d4208e1798a010fdec4cd680f/178] Error executing command: std::runtime_error exception thrown at [buildboxcommon_casclient.cpp:900], errMsg = "Error staging "63db766975a429f55445f8e0a24f4107be339b311f002cb631841a71d91234ed/1438" into "": "std::runtime_error exception thrown at [buildboxcasd_fusestager.cpp:144], errMsg = "The FUSE stager child process unexpectedly died with exit code 1"""
        [00:00:00] FAILURE hello.bst: Running commands
        [00:00:00] FAILURE [8cebf405] hello.bst: buildbox-run failed with returncode 1
        ======================================================================

Pipeline Summary
    Total:       3
    Session:     3
    Fetch Queue: processed 3, skipped 0, failed 0 
    Build Queue: processed 2, skipped 0, failed 1 


[root@de014e636534 running-commands]# mkdir cas local
[root@de014e636534 running-commands]# buildbox-fuse --local=cas local
fusermount3: mount failed: Operation not permitted
```

So it seems I can't usefully build stuff if I don't supply `--privileged` to podman. Adding that to the commandline got me a successful build.

anyway, I will check if `bwrap` does work unprivileged and reword the doc. thanks for the pointers!

ghost: I also see differences in `bwrap` behaviour with and without `--privileged`:

Without --privileged:
```
# bwrap --unshare-all --ro-bind / /  --dev-bind /dev /dev --uid 0 /bin/bash
bwrap: Can't bind mount /oldroot/ on /newroot/: Unable to mount source on destination: Operation not permitted
```

With --privileged: I get a running `bash` instance.

nanonyme: Fair. So the documentation is correct, it just might just be in podman case be less privileged than docker.

doc/source/main_install.rst
```diff
@@ -130,28 +116,23 @@ Installing BuildStream
...
-For the latest dev snapshot of BuildStream 2, simply run the following command::
+For the latest pre-release of BuildStream 2, including the necessary Python
+dependencies and BuildBox tools, run the following command::
 
     pip3 install --user --pre BuildStream
 
-This will install latest dev snapshot of BuildStream and its pure python
-dependencies into your user's homedir in ``~/.local``.
+This will install BuildStream and its dependencies into your user's homedir in
+``~/.local``.  Pip will use binary "wheel" packages from PyPI where these are
```
gtristan: Nit:

> Pip will use binary "wheel" packages from PyPI where these are available for your platform.

Would probably read better with *"if these are available for your platform"* (whereas *"where"* would be more suitable if *"for your platform"* was removed).

gtristan:
Overall, I like the install from source instructions here, however it looks like this is unfortunately again growing into a lot of overlap with the user facing instructions at: https://buildstream.build/install.html

Lets keep the user facing install paths published at https://buildstream.build/install.html and have the version specific developer/packager facing instructions at https://docs.buildstream.build/master/main_install.html. This makes the two pages serve obviously different purposes, and makes the reading of either of these less crowded and intense.

Maybe it would be helpful to have a little map:

```
                            Main page (A)
                             /       \
                           /           \
            Source (B-1), (B-2)       Docker (C)

(A): https://buildstream.build/install.html
(B-1): https://docs.buildstream.build/master/main_install.html
(B-2): https://docs.buildstream.build/1.6/main_install.html
(C): https://gitlab.com/BuildStream/buildstream-docker-images/-/blob/master/USING.md
```

While sub-pages might link back to the main user facing install page on the main website (A), we should always assume (A) as a starting point, and (A) should always be our answer for anyone's question of *"How do I install BuildStream"*

### Introduction

I think the statement:

> BuildStream is packaged in some Linux distributions. If your distro has an up-to-date package we recommend using that.

Can go away, this is already mentioned where it is better suited, at: https://buildstream.build/install.html

It could be good to have a forward in `main_install.rst` saying something like:

> This page explains how to build and install this version of BuildStream from sources. If you are trying to install BuildStream on your computer it is recommended to try following the [general purpose installation instructions here](https://buildstream.build/install.html) first, and return here only if you need to build and install from source.

### Installation from PyPI

I think this can all go away and is not relevant here, and is already explained where it is better suited at: https://buildstream.build/install.html

### Docker images

I think the whole *"Using docker images"* section here can also go away completely.

If there are user facing images ready to be used by users, we should trust https://gitlab.com/BuildStream/buildstream-docker-images/-/blob/master/USING.md to document how to use the latest BuildStream 1 and BuildStream 2 images respectively, and link to that page from the user facing documentation at https://buildstream.build/install.html (without going into a whole detailed shpiel about it there either)...

And of course we should touch up https://gitlab.com/BuildStream/buildstream-docker-images/-/blob/master/USING.md a bit to make sure it at least tells you the registry and image names which are useful to use, and posts some useful invocation like:

```shell
docker run -priviliged -it -v `pwd`:/build <bstimagename:latest> /bin/bash
```

### Build/run dependencies

When discussing installing dependencies, here we are talking about the dependencies needed to install *from source*, so it is unnecessary to split up dependencies as we currently have them down (redundantly on a per-distro basis on top of it), with statements like: *"Run this command"* and then if that is not enough, also *"Run this other command too"*.

Also, this would look so much more concise if it were simply a condensed table with two columns:

| Distribution        | Command           |
| ------------- |-------------|
| Arch      | sudo pacman -S bubblewrap fuse3 git lzip patch python python-pip gcc |
| Debian  | sudo apt-get install git lzip patch bubblewrap fuse3 python3-pip python3 g++ python3-dev |


Otherwise right now, it looks like you have to scroll through pages of documentation just to install the distro dependencies needed to build and install BuildStream.

### Checking out git

When talking about checking out git (which are the only installs we need to talk about on this page I guess), it is unnecessary to talk about *Upgrading* separately.

If we have instructed the user how to checkout a useful tag/branch which contains what they want, then they already have the ammunition they need go ahead and install a different or newer version later.

### Install instructions

When discussing the actual installation instructions, it is worth noting:

* `pip3 install --user .` as a default just to conveniently install
* `python3 setup.py install` it is important to note that this is supported because this is the regular distro packager path, which does not involve downloading dependencies from the non-existing internet

We should also have a simple reminder here that depending on where you have decided to install buildstream, you may need to adjust `$PATH` or *activate* your virtual environment in order to use it.

This last part will allow us to nuke that whole **Post-install setup** section which merely talks about `$PATH`

#### Virtual environments

It is also worth noting as part of install instructions that you can use a python virtual environment if you want to have multiple versions of BuildStream installed on the same machine.

An interesting question is whether to duplicate virtual environment instructions on the main page (using PyPI packages) and here on the source installation page (installing from git).

If we continue to link to this page from the main page in order to discuss virtual environments, we should use an encoded link anchor in the reStructuredText for this:

```
.. _install-virtual-environment:
```

Because the main website has a link to: https://docs.buildstream.build/master/main_install.html#installing-in-virtual-environments which only happens to work because we didn't yet haphazardly rename the title of a section in our docs.

### Completions

We should restore the section about *completions* as this is an important detail for distro packagers and also users.

For this we can merely link: https://github.com/apache/buildstream/blob/master/src/buildstream/data/bst

We can safely say (like the file says) where it should be installed, and we can even just paste the contents of that script and inform users that they can just append that text to `~/.bash_completion` (this completion script never changes so it's quite safe).

---

After trimming out all that fat above, this would leave us with a more manageable structure of only:

* Introduction
* Dependencies for building/running this version of buildstream
* Installing buildbox
  * Instructions about binaries
  * Link(s) to how to build buildbox from sources in upstream buildbox
* Cloning the right git tag/branch
* Installation instructions (from the checkout)
  * Virtual environments
* Completions

doc/source/main_install.rst
abderrahim: @nanonyme AFAIK toolbx uses a `--privileged` podman container. You're probably confusing rootless with non-privileged.

ghost:
I'm taking another look at this today, thanks for the detailed guidance.

On dependencies and package lists, I will move them into a table, but we should still separate the install-time requirements (gcc, pip, python3-devel) from the runtime requirements. Let's try a table with 2 columns.

About a suitable 'install' command for distro packagers, it's useful to give some guidance on this. The `python3 setup.py install` command will work (as long as Bst doesn't change buildsystem), but it may not be what distro package guidelines recommend now. did some quick research to satisfy my own curiousity:

  * Fedora recommends use of [a %pyproject_install macro](https://docs.fedoraproject.org/en-US/packaging-guidelines/Python/#_build_macros), which runs `pip install --no-deps --no-index --use-pep517 .` ([link](https://src.fedoraproject.org/rpms/pyproject-rpm-macros/blob/rawhide/f/macros.pyproject#_60))
  * Debian doesn't give any strong recommendation, the existing [buildstream package](https://tracker.debian.org/pkg/buildstream) uses [pybuild](https://manpages.debian.org/bullseye/dh-python/pybuild.1.en.html) via debhelper, which autodetects one of multiple backends. Running `pybuild --detect` in buildstream.git reports `distutils`, so they do indeed [run `setup.py` directly](https://salsa.debian.org/python-team/tools/dh-python/-/blob/master/dhpython/build/plugin_distutils.py), although there is a `pyproject` backend that may become default in future which [runs `python3 -m build`](https://salsa.debian.org/python-team/tools/dh-python/-/blob/master/dhpython/build/plugin_pyproject.py) and post-processes the generated `.whl`.

All this to highlight there's no "one size fits all" distro packager command, they should use whatever the Python guidelines for that distro recommends, and that can be either running setuptools directly or any PEP517 compatible frontend.

doc/source/main_install.rst
ghost: Removed in new version

ghost:
Updated following the above guidance.

doc/source/main_install.rst
```diff
@@ -1,221 +1,182 @@
...
+user's homedir in ``~/.local``, and will attempt to fetch and install any
+required PyPI dependencies from the internet at the same time::
+
+
     pip3 install --user .
 
-This will install BuildStream's pure python dependencies into
-your user's homedir in ``~/.local`` and will run BuildStream directly
-from the git checkout directory.
+You can optionally use Pip's
```
gtristan: I don't believe the `-e` option has worked for a very long time (with buildstream specifically, this might have been since the transition to use cython).

doc/source/main_install.rst
```diff
@@ -258,40 +219,23 @@ and set up other necessary environment variables::
...
-
-  export PATH="${PATH}:${HOME}/.local/bin"
-
-.. note::
+Installing completions
+----------------------
 
-   You will have to restart your terminal in order for these changes to take effect.
+BuildStream integrates with Bash to provide helpful tab-completion. These
+are provided by the script ``src/buildstream/data/bst`` which requires
```
gtristan: Wouldn't it be good to expand the content of `src/buildstream/data/bst` here (with `::literalinclude`) or at least provide a link to the master branch version of the `bst` completions script on github ?

Also I thought there was an easier way to get this working for a single user, by simply copying the content of the script (from this page) and pasting the content directly into `~/.bash_completion`.

doc/source/main_install.rst
ghost: OK, will remove. In fact, probably better to specifically note that it can't be used.

ghost: Opened https://github.com/apache/buildstream/issues/1760 so we have somewhere to answer the question "Why can't I use `pip -e`" and will link that from the docs

doc/source/main_install.rst
ghost: We expect that the user has a Git clone if they're reading these instructions.

I'm not convinced that 'add this text to this file' is much simpler to do than 'copy this file to this location'. It seems simpler to express it though :)

I will rework as described

ghost:
Last 2 review comments now addressed in 7925769d9

PR #1745, opened 2022-08-30
Add release notes for bst 1.6.8

nanonyme:


nanonyme:
Ping @jjardon

abderrahim:
I'd like to backport https://github.com/apache/buildstream/pull/1436 too before the next release.

nanonyme:
Fair. We can pull that in. We IMHO need the new release asap though far this Fall Fedora release

nanonyme:
@abderrahim rebased and mentioned your change in release notes.

abderrahim:
Maybe add the python 3.11 fix if it works now? (or maybe we still need to change the requirements?)

nanonyme:
I don't think there's much we can do at the moment to fix Python 3.11 for pip installs. Someone needs to file a bug against ruamel.yaml.clib.

nanonyme:
But this allows fixing the Fedora-shipped version of bst which is nice.

nanonyme:
I can add comment about the backport, sure.

nanonyme:
Done.

nanonyme:
I would prefer not going as far as claiming Python 3.11 support is fixed though it probably is.

PR #1750, opened 2022-09-09
cachequeryqueue.py: support querying sources when artifacts are cached

### LINKED ISSUE #1749: BUG when using the `storage-service` cache option
After adding the `storage-service` cache option, I get the following error. The cache wasn't completely empty (was used as an artifact cache and source cache before) but isn't completely up to date either.

It seems it is trying to push sources for an element without first querying the cache to check that the source is cached.

```
[05:36:29][--:--:--][        ][    main:core activity                 ] BUG     
      File "/usr/lib/python3.10/asyncio/events.py", line 80, in _run
        self._context.run(self._callback, *self._args)
      File "/usr/lib/python3.10/site-packages/buildstream/_scheduler/scheduler.py", line 438, in real_schedule
        self._sched_queue_jobs()
      File "/usr/lib/python3.10/site-packages/buildstream/_scheduler/scheduler.py", line 378, in _sched_queue_jobs
        queue.enqueue(elements)
      File "/usr/lib/python3.10/site-packages/buildstream/_scheduler/queues/queue.py", line 192, in enqueue
        self._enqueue_element(elt)
      File "/usr/lib/python3.10/site-packages/buildstream/_scheduler/queues/queue.py", line 385, in _enqueue_element
        status = self.status(element)
      File "/usr/lib/python3.10/site-packages/buildstream/_scheduler/queues/sourcepushqueue.py", line 36, in status
        if element._skip_source_push():
      File "/usr/lib/python3.10/site-packages/buildstream/element.py", line 1917, in _skip_source_push
        return not (self.__sourcecache.has_push_remotes(plugin=self) and self._cached_sources())
      File "/usr/lib/python3.10/site-packages/buildstream/element.py", line 2279, in _cached_sources
        return self.__sources.cached()
      File "/usr/lib/python3.10/site-packages/buildstream/_elementsources.py", line 342, in cached
        assert self._cached is not None
    AssertionError
```

abderrahim:
Fixes #1749

### DIFF
--- a/src/buildstream/_scheduler/queues/cachequeryqueue.py
+++ b/src/buildstream/_scheduler/queues/cachequeryqueue.py
@@ -28,18 +28,27 @@ class CacheQueryQueue(Queue):
     complete_name = "Cache queried"
     resources = [ResourceType.PROCESS, ResourceType.CACHE]
 
-    def __init__(self, scheduler, *, sources=False):
+    def __init__(self, scheduler, *, sources=False, sources_if_cached=False):
         super().__init__(scheduler)
 
         self._sources = sources
+        self._sources_if_cached = sources_if_cached
 
     def get_process_func(self):
-        if not self._sources:
+        if self._sources_if_cached:
+            return CacheQueryQueue._query_artifacts_and_sources
+        elif not self._sources:
             return CacheQueryQueue._query_artifacts_or_sources
         else:
             return CacheQueryQueue._query_sources
 
     def status(self, element):
+        if element._can_query_cache():
+            # Cache status already available.
+            # This is the case for artifact elements, which load the
+            # artifact early on.
+            return QueueStatus.SKIP
+
         if not element._get_cache_key(strength=_KeyStrength.WEAK):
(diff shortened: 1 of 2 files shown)


juergbi:
The changes to match the serial cache query logic make sense to me. A test case would be good but I don't know how tricky this is to add. Build with `storage-service` and source push enabled with a target that is already built and cached?

nanonyme:
This seems to fix the immediate problem based on our testing integration testing in freedesktop-sdk.

jjardon:
@abderrahim can we remove the Draft status from this or something else is still needed?

abderrahim:
I've removed the draft status. I think it's ready, but I'm not sure how to test it.

Maybe we should simply make this the default path, and find a way to simplify the output (as it's currently very verbose).

nanonyme:
@juergbi note we have done end-to-end testing a lot with this in freedesktop-sdk and it seems to fix the problems. Using storage service works reliably with this change.

juergbi (approved):

PR #1754, opened 2022-09-16
Use subprocess umask support

nanonyme:
This requires Python 3.9 or newer

### DIFF
--- a/src/buildstream/utils.py
+++ b/src/buildstream/utils.py
@@ -13,12 +13,6 @@
 #
 #  Authors:
 #        Tristan Van Berkom <tristan.vanberkom@codethink.co.uk>
-
-# Disable this for the file, because pylint is not picking it up
-# when specifying it on the specific line.
-#
-# pylint: disable=subprocess-popen-preexec-fn
-#
 """
 Utilities
 =========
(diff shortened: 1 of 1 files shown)


nanonyme:
The leading idea here is that the time between fork and exec is perilous and we can avoid an unnecessary context switch from C code to Python code by using Python builtin umask setting functionality. The os.umask call itself is fine but the dance to make existing pre-exec function work through creating a Python wrapper means we do a context switch.

nanonyme:
I will remove workaround and wait until BuildStream is ready to drop Python 3.8.

juergbi (approved):
Looks good to me. As the release process for 2.4.0 has already started, it may be a bit late to get this still in before the release, though.

juergbi:
We should be able to remove `pylint: disable=subprocess-popen-preexec-fn` from `utils.py`.

nanonyme:
@juergbi removed the stanza.

src/buildstream/utils.py
```diff
@@ -1260,14 +1254,7 @@ def _call(*popenargs, terminate=False, **kwargs):
 
     process = None
 
-    old_preexec_fn = kwargs.get("preexec_fn")
```
nanonyme, line 1263: @juergbi I'm not 100% sure if this should logically be like this or `kwargs.setdefault("umask"], stat.S_IWGRP | stat.S_IWOTH)`. Is it expected caller can override umask?

nanonyme: I think with original functionality they probably could as caller preexec was called *after* the preexec to set umask.

juergbi: We don't allow plugins to set `preexec_fn` or `umask`, so it wouldn't make a difference right now. However, the point of this umask is to make sure that the umask is deterministic. It's not completely unreasonable to allow plugins to choose a different umask. So I think changing it to `setdefault()` would make sense.

PR #1757, opened 2022-09-18
bst-1: .github: Run tests in fedora 37 as well

jjardon:
And remove soon to be deprecated fedora 35

Currently failing because  python 3.11 is not supported: #1758


nanonyme:
Looks like ruamel.yaml.clib does not support Python 3.11 which is in Fedora 37.

nanonyme:
... no, it's not failing because of that. It's failing because of dependency issues.

nanonyme:
@jjardon please stop rebasing this until adding more fixes. This is missing ruamel.yaml.clib version bump and is guaranteed to fail. The current pinned version *cannot be built on Python 3.11*.

jjardon:
Adding more fixes is what I'm trying to do . Thanks for the points

nanonyme:
Note I had absolutely no idea if updating grpcio is right thing to do but older grpcio does not build on Python3.11. I think the hangs are somehow related to new grpcio.

gtristan:
This was fixed in @nanonyme's branch which I then adapted separately and landed for the last bst-1 release.

Closing.

PR #1759, opened 2022-09-19
.github: Run tests in fedora 37 as well

jjardon:
And remove soon to be deprecated fedora 35

Currently failing because python 3.11 is not supported: #1802

### DIFF
--- a/.github/common.env
+++ b/.github/common.env
@@ -1,6 +1,6 @@
 # Shared common variables
 
-CI_IMAGE_VERSION=master-533491591
-CI_TOXENV_MAIN=py37,py38-nocover,py39-nocover,py310-nocover
-CI_TOXENV_PLUGINS=py37-plugins,py38-plugins-nocover,py39-plugins-nocover,py310-plugins-nocover
+CI_IMAGE_VERSION=master-643533272
+CI_TOXENV_MAIN=py37,py38-nocover,py39-nocover,py310-nocover,py311-nocover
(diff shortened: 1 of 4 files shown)


ghost:
Looks to be failing due to missing `longintrepr.h` needed by pyroaring:

      pyroaring/pyroaring.cpp:220:12: fatal error: longintrepr.h: No such file or directory

This appears to be a bug in Cython which was already fixed last year: https://github.com/cython/cython/issues/4461

Suggested fix would be to update the minimum Cython version 0.29.25

juergbi:
> Suggested fix would be to update the minimum Cython version 0.29.25

This has been merged in the meantime as part of #1778.

jjardon:
Seems we still have he same problem: https://github.com/apache/buildstream/actions/runs/3698106262/jobs/6263904645
```
      pyroaring/pyroaring.cpp:220:12: fatal error: longintrepr.h: No such file or directory
        220 |   #include "longintrepr.h"
            |            ^~~~~~~~~~~~~~~
```

cs-shadow:
pyroraing should be fine after https://github.com/apache/buildstream/pull/1806 - we just had pinned an older release, newer versions are fine.

the next issue is with grpcio - see https://github.com/apache/buildstream/pull/1807 for how that fails. https://github.com/grpc/grpc/pull/31960 should backport 3.11 support to the minor version we're on. However this  also means that BuildStream (or at least its testsuite) doesn't work with grpicio versions 1.49 and above, which might be a problem in its own right.

Issue #1762, opened 2022-09-23
bst2: Impossible to track origin with fallback to mirror


nanonyme:
We in freedesktop-sdk project need BuildStream 2 to support tracking model where origin is tracked first, then mirror. This allows us to discover updates through tracking while preserving functionality if origin goes away. The controls https://docs.buildstream.build/master/using_config.html#track-controls provides make this impossible. This is a regression from bst1. I propose instead changing track in following way

```
track:
  source:
  - aliases
  - mirrors
  - user
```
so the order is *explicit* and user can define how it works.

gtristan:
> We in freedesktop-sdk project need BuildStream 2 to support tracking model where origin is tracked first, then mirror

Are you saying that with bst2, when configured with:

```yaml
track:
  source: all
```

That the mirrors are attempted *before* the upstream instead of *after* the upstream ?

In bst1, while this granular control over *what* is considered when tracking/fetching is not present, we *do* make the assumption that the order at track time is reversed from fetch time, a simple fix might be to re-introduce this reverse ordering for now.

nanonyme:
Yes, it seems like the order is now indeed to try mirrors first. This means that it is impossible to discover any new updates unless tracking mirror fails.

nanonyme:
https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/jobs/3077047426 is example run. It did resolve some updates but these were mostly cases where
1. Mirror was busted
2. We had custom tracking in place that ignored mirror in any case

nanonyme:
I'm fine with simple order reversal. It should make auto-updater work again as expected. I'm not sure if it's possible some user might want a different order, hence why I was suggesting order be configurable.

gtristan:
This is fixed by #1765 which reverts to the old behavior of reversing the list at track time.

For now, there is already a lot of added controls over track/fetch sources in BuildStream 2, so I would prefer to wait until such a time that a user presents a use case where further control on traversal order would be necessary for their needs.

PR #1765, opened 2022-09-29
Use reversed order for sources while tracking

nanonyme:
Related to https://github.com/apache/buildstream/issues/1762

This is currently not controllable so revert to same order bst1 had. Reasoning: when resolving new content, we want to prefer upstream whereas when downloading, we want to prefer mirrors. Ideally this order should be controllable through user config.

### DIFF
--- a/src/buildstream/source.py
+++ b/src/buildstream/source.py
@@ -1372,7 +1372,7 @@ class Source(Plugin):
         # NOTE: We are assuming here that tracking only requires substituting the
         #       first alias used
         last_error = None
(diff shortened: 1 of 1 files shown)


nanonyme:
Locally tested, this seems to be enough to fix freedesktop-sdk tracking.

nanonyme:
@gtristan can I have review on this?

gtristan:
Yup, this got delayed because 2 factor auth was difficult/unavailable to me for a while overseas...

PR #1768, opened 2022-10-03
sandbox: add dummy sandbox for checking out artifacts from different arches

abderrahim:
We've added the ability to pull stuff from different architectures since 1.5.1 but that wasn't usable as we could not check out the result (even if we don't need to run anything)

Fixes https://github.com/apache/buildstream/issues/1352 for bst-1

### DIFF
--- a/buildstream/_platform/linux.py
+++ b/buildstream/_platform/linux.py
@@ -22,12 +22,20 @@ import subprocess
 
 from .. import _site
 from .. import utils
-from ..sandbox import SandboxBwrap
+from ..sandbox import SandboxBwrap, SandboxDummy
 
 from . import Platform
 
 
 class Linux(Platform):
+
+    ARCHITECTURES = {
+        'amd64': 'x86_64',
+        'arm64': 'aarch64',
+        'i386': 'i686',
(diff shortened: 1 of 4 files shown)


buildstream/_platform/linux.py
```diff
@@ -22,13 +22,21 @@
...
 from .. import utils
-from ..sandbox import SandboxBwrap
+from ..sandbox import SandboxBwrap, SandboxDummy
 
 from . import Platform
 
 
 class Linux(Platform):
 
+    ARCHITECTURES = {
```
nanonyme, line 32: I know it was called like this before so it's not really a change but not entirely obvious from name that this is just an alias mapping and unlisted things may be supported.

nanonyme:
Looks good to me.

gtristan:
Overall looks good to me.

One thing is curious to me, it appears that even if we do support 32bit builds on 64bit hosts via `arch-test`, we still bail out and complain if the user does not have `linux32` installed. However this is clearly not a regression looking at the ported code.

EDIT: On further thought; perhaps it is still necessary to have uname return a 32bit arch for these scenarios, so this appears correct in that light.

Issue #1766, opened 2022-09-29
Traceback bug in artifact fetch


staehle:
Hello:

Evaluating BuildStream as a meta build system for an embedded Linux OS, one of the users in my evaluation group attempted to do a first-time build and got this BuildStream traceback/bug.  I can't share too much unfortunately, but this is a from-scratch project and is currently using a temporary BuildBarn instance as a caching server only using the instructions here: https://docs.buildstream.build/1.95/using_configuring_cache_server.html / https://github.com/apache/buildstream/tree/1.95.2/.github/compose 

This "components/target/rsync" element is fairly far down the element dependency list

```
    [00:00:00][e062aa1a][   fetch:components/target/rsync.bst   ] BUG     Fetch

        An unhandled exception occured:
    
        Traceback (most recent call last):
          File "/home/builduser/.local/lib/python3.9/site-packages/buildstream/_scheduler/jobs/job.py", line 441, in child_action
            result = self.child_process()  # pylint: disable=assignment-from-no-return
          File "/home/builduser/.local/lib/python3.9/site-packages/buildstream/_scheduler/jobs/elementjob.py", line 92, in child_process
            return self._action_cb(self._element)
          File "/home/builduser/.local/lib/python3.9/site-packages/buildstream/_scheduler/queues/fetchqueue.py", line 77, in _fetch_not_original
            element._fetch(fetch_original=False)
          File "/home/builduser/.local/lib/python3.9/site-packages/buildstream/element.py", line 2210, in _fetch
            self.__sources.fetch()
          File "/home/builduser/.local/lib/python3.9/site-packages/buildstream/_elementsources.py", line 225, in fetch
            self.fetch_sources()
          File "/home/builduser/.local/lib/python3.9/site-packages/buildstream/_elementsources.py", line 254, in fetch_sources
            self._fetch_source(source)
          File "/home/builduser/.local/lib/python3.9/site-packages/buildstream/_elementsources.py", line 435, in _fetch_source
            source._fetch()
          File "/home/builduser/.local/lib/python3.9/site-packages/buildstream/source.py", line 911, in _fetch
            self.__do_fetch()
          File "/home/builduser/.local/lib/python3.9/site-packages/buildstream/source.py", line 1404, in __do_fetch
            self.fetch(**kwargs)
          File "/home/builduser/.local/lib/python3.9/site-packages/buildstream/downloadablefilesource.py", line 183, in fetch
            sha256 = self._ensure_mirror(
[...]

Issue #1761, opened 2022-09-23
bst2 does not propagate error to console when tracking returns 404


nanonyme:
```
Failure Summary
    components/ladspa-sdk.bst:
    [17:08:49][00:00:03][88b09349][   track:components/ladspa-sdk.bst     ] FAILURE Background process died with error code 0
        Printing the last 80 lines from log file:
        /builds/freedesktop-sdk/freedesktop-sdk/cache/buildstream/logs/freedesktop-sdk/components-ladspa-sdk/88b09349-track.42.log
        ======================================================================
        BuildStream 1.95.2+0.g84f6d5e.dirty - Friday, 23-09-2022 at 17:08:46
        [--:--:--] START   [88b09349] components/ladspa-sdk.bst: Track
        [--:--:--] START   components/ladspa-sdk.bst: Tracking http://www.ladspa.org/download/ladspa_sdk_1.15.tgz
        [00:00:03] FAILURE components/ladspa-sdk.bst: Tracking http://www.ladspa.org/download/ladspa_sdk_1.15.tgz
        [00:00:03] FAILURE [88b09349] components/ladspa-sdk.bst: Background process died with error code 0
```
Upstream URL is apparently dead. bst2 must not in this case give up but instead fallback to mirror.

nanonyme:
As per @gtristan there is behavioural change with how mirrors are dealt for tracking. But there is a bug in https://github.com/apache/buildstream/blob/a54f88bc716eff24066f5289cabc911e49ffdb2f/src/buildstream/downloadablefilesource.py#L223-L224 which prevents user from seeing what the actual error was.

nanonyme:
This is now probably fixed.

PR #1771, opened 2022-10-05
Stop using child process for downloading URL's

nanonyme:
Related to https://github.com/apache/buildstream/issues/1766 and https://github.com/apache/buildstream/issues/1761 This makes things really hard to debug when there are errors and downloading is mostly done with GIL dropped.

### DIFF
--- a/src/buildstream/downloadablefilesource.py
+++ b/src/buildstream/downloadablefilesource.py
@@ -113,7 +113,8 @@ def _download_file(opener, url, etag, directory):
         filename = os.path.basename(filename)
         local_file = os.path.join(directory, filename)
         with open(local_file, "wb") as dest:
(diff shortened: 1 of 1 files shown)


abderrahim:
It just crossed my mind that maybe what was meant by "starving the scheduler" isn't actually about the GIL. Can you please try to terminate a fetch during download and see if it works?

nanonyme:
Good point. Though I don't think termination works right now either. You get a really miserable set of stacktraces which eventually maybe results in stopping.

nanonyme:
I converted this into draft so won't be accidentally merged before the important question is answered.

nanonyme:
@abderrahim has better fix for this at https://github.com/apache/buildstream/pull/1814

PR #1775, opened 2022-10-12
Zsh tab-completions, with related doc changes

staehle:
Hello, I found a related ticket for this: #558

I primarily use Zsh with Prezto, and not having tab-completions was bothering me, so here you go :) 

It just uses your existing Bash completion code, but rearranges it for Zsh's `compdef`. File name must be `_bst`, and I placed in a `zsh` directory in the same location as the bash `bst` script is. Hope that's okay.

Also added related usage instructions to the docs.

For future enhancements in this area, I would recommend doing tab-completions similar to how Poetry does it, but that's up to you:  https://python-poetry.org/docs/#enable-tab-completion-for-bash-fish-or-zsh 

Thanks!

### DIFF
--- a/doc/source/main_install.rst
+++ b/doc/source/main_install.rst
@@ -223,11 +223,15 @@ consider using `pipx <https://docs.python.org/3/tutorial/venv.html>`_.
 Installing completions
 ----------------------
 
-BuildStream integrates with Bash to provide helpful tab-completion. These
-are provided by the `bst` completion script, available online
-`here <https://raw.githubusercontent.com/apache/buildstream/master/src/buildstream/data/bst>`_
-and in your local Git clone at ``src/buildstream/data/bst``. The completion
-script requires manaul installation.
+BuildStream integrates with Bash and Zsh to provide helpful tab-completion.
+These completion scripts require manual installation.
+
+Bash completions
+~~~~~~~~~~~~~~~~
+
+Bash completions are provided by the ``bst`` completion script, available online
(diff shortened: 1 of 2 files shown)


abderrahim:
please mention that it closes https://github.com/apache/buildstream/issues/558

staehle:
> please mention that it closes #558

I did, and it's in the commit message

doc/source/main_install.rst
```diff
@@ -239,3 +243,41 @@ bash-completion installation path, which you can discover as follows::
...
+    cp src/buildstream/data/zsh/_bst $ZSH_CUSTOM/plugins/bst/_bst
+
+You must then add ``bst`` to your plugins array in ``~/.zshrc``::
+
+    plugins(
+      bst
+      ...
+    )
+
+**None**::
```
abderrahim: This would be probably clearer as "No framework" or "Vanilla zsh". Also I don't know how people using zsh do it, but shouldn't this be the first one?

abderrahim (approved):
Looks good, except for a small comment.

Note that we're currently in the process of releasing 2.0, so it might not be merged until after that.

doc/source/main_install.rst
staehle: You're right, I did word that a bit confusing. I might be incorrectly assuming that everyone uses a framework with Zsh, so I reworded that to remove that assumption.
Added one more commit to revise that (as well as rebasing to 1.95.4)! Hopefully that looks better!
Thanks for the review!

Issue #1758, opened 2022-09-18
bst1: buildstream is not compatible with python 3.11


jjardon:
See https://github.com/apache/buildstream/pull/1757

This is a problem as, for example,  the Fedora 37 default installation only includes python 3.11

Issue for master (buildstream2): #1802

abderrahim:
It seems that ruamel.yaml.clib is the problem here, it doesn't provide a 3.11 wheel and fails to build with 3.11. But since it's an optional dependency, we can probably skip it for python 3.11 for now.

Sadly, it's not optional for bst2 so we'll have to wait for bst2.

mcatanzaro:
I see the following error when running `bst build --track-all`:

```
[00:00:00][4043f233][build:freedesktop-sdk.bst:bootstrap/build/base-sdk/filtered.bst] BUG     Build

    An unhandled exception occured:
    
    Traceback (most recent call last):
      File "/usr/lib/python3.11/site-packages/buildstream/_scheduler/jobs/job.py", line 384, in _child_action
        result = self.child_process()
                 ^^^^^^^^^^^^^^^^^^^^
      File "/usr/lib/python3.11/site-packages/buildstream/_scheduler/jobs/elementjob.py", line 94, in child_process
        return self._action_cb(self._element)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/usr/lib/python3.11/site-packages/buildstream/_scheduler/queues/buildqueue.py", line 35, in process
        return element._assemble()
               ^^^^^^^^^^^^^^^^^^^
      File "/usr/lib/python3.11/site-packages/buildstream/element.py", line 1534, in _assemble
        collect = self.assemble(sandbox)
                  ^^^^^^^^^^^^^^^^^^^^^^
      File "/usr/lib/python3.11/site-packages/buildstream/plugins/elements/filter.py", line 108, in assemble
        dep.stage_artifact(sandbox, include=self.include,
      File "/usr/lib/python3.11/site-packages/buildstream/element.py", line 641, in stage_artifact
        files = list(self.__compute_splits(include, exclude, orphans))
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/usr/lib/python3.11/site-packages/buildstream/element.py", line 2329, in __compute_splits
        self.__init_splits()
[...]

PR #1785, opened 2022-11-03
bst1: Support Python 3.11

nanonyme:
Fixes https://github.com/apache/buildstream/issues/1758

### DIFF
--- a/.github/common.env
+++ b/.github/common.env
@@ -1,5 +1,5 @@
 # Shared common variables
 
-CI_IMAGE_VERSION=master-533491591
+CI_IMAGE_VERSION=master-643533272
 CI_TOXENV_MAIN=py36-nocover,py37-nocover,py38-nocover,py39-nocover,py310-nocover
 CI_TOXENV_ALL="${CI_TOXENV_MAIN}"
--- a/.github/common.env
+++ b/.github/common.env
@@ -1,5 +1,5 @@
 # Shared common variables
 
 CI_IMAGE_VERSION=master-643533272
-CI_TOXENV_MAIN=py36-nocover,py37-nocover,py38-nocover,py39-nocover,py310-nocover
+CI_TOXENV_MAIN=py36-nocover,py37-nocover,py38-nocover,py39-nocover,py310-nocover,py311-nocover
 CI_TOXENV_ALL="${CI_TOXENV_MAIN}"
--- a/.github/compose/ci.docker-compose.yml
+++ b/.github/compose/ci.docker-compose.yml
@@ -1,7 +1,7 @@
 version: '3.4'
 
 x-tests-template: &tests-template
-    image: registry.gitlab.com/buildstream/buildstream-docker-images/testsuite-fedora:35-${CI_IMAGE_VERSION:-latest}
+    image: registry.gitlab.com/buildstream/buildstream-docker-images/testsuite-fedora:36-${CI_IMAGE_VERSION:-latest}
     command: tox -vvvvv -- --color=yes --integration
     environment:
       TOXENV: ${CI_TOXENV_ALL}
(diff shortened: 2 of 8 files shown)


mcatanzaro:
Why is this still a draft?

nanonyme:
@mcatanzaro tests don't pass and it's not directly related to Python 3.11

nanonyme:
I am not getting a pipeline, dunno why.

nanonyme:
I'm not entirely sure if tests hang on Fedora  36 and 37 but at least they're massively slower.

nanonyme:
Looks like there are failing tests. The failing tests are failing really slowly. I'm trying to make tests abort after first failure so we see real output.

nanonyme:
Well, that was not overly useful. I got it to actually end and we're maybe a bit closer to understanding why the test is so slow https://github.com/apache/buildstream/actions/runs/3688689726/jobs/6243779599 but the traceback is mostly gibberish and there is no console output

mcatanzaro:
Hi @gtristan, can you look at this please? buildstream 1 is really super broken currently, and for a very long time now (it's not like python 3.11 is new anymore).

nanonyme:
Note that this is about CI and testing on Python 3.11. This has no functional changes.

mcatanzaro:
Oops, I see it is #1758 that requires attention, not this #1785.

nanonyme:
There are unreleased fixes for Python3.11 in bst-1 branch. But without CI it's hard to say whether further things are broken.

mcatanzaro:
Er, why not release them then? This has been blocking a lot of stuff for a long time.

(I agree that CI is very important, but it's not urgent.)

nanonyme:
It would be nice if someone pip installed bst1 from this repo who actually has Python3.11 to give it at least some lightweight manual testing.

jjardon:
@mcatanzaro @nanonyme I _think_ that is what is being done in #1757 (It hangs forever)

nanonyme:
@jjardon I am not taking about CI. Just install it and use it. It will not hang. See if there are further regressions.

nanonyme:
If it seems usable, then we tag a release and worry about CI later.

nanonyme:
The hangs are related to special way bst event loop is used  for unittest purposes in same process restarting it over and over again. They are not indicative of real problems.

nanonyme:
@jjardon it also does not hang "forever". Asyncio gets into bad state and times out with 30 minutes at a time. It just times out dozens of times in a loop so it looks like forever.

mcatanzaro:
> It would be nice if someone pip installed bst1 from this repo who actually has Python3.11 to give it at least some lightweight manual testing.

You don't have access to an F37 toolbx container? That should take maybe two minutes to create. Whatever. I installed this branch and am building gnome-build-meta with it now. Will let you know how it goes....

mcatanzaro:
> Will let you know how it goes....

Well it finished successfully, but I think it pulled everything from the artifact cache without actually building anything.

Want me to test anything else in particular?

nanonyme:
I think that's strong indication we should wrap a quick release. If you can build something (doesn't matter what, you can force rebuild by setting environment variable in some element), that would be even better.

nanonyme:
I created https://github.com/apache/buildstream/pull/1809 anyhow. I would like to have CI actually work before advertising Python 3.11 support in metadata. But this should hopefully unblock downstreams.

mcatanzaro:
> I think that's strong indication we should wrap a quick release. If you can build something (doesn't matter what, you can force rebuild by setting environment variable in some element), that would be even better.

Did this to force building core/epiphany.bst. It worked.

gtristan:
This is replaced by #1827

PR #1786, opened 2022-11-05
_sandboxbuildboxrun.py: Restore terminal after exit of interactive child

### LINKED ISSUE #1690: buildstream goes into the background after entering a shell
At the failed build prompt, answer `s` to get to a shell. When exiting the shell, bst prints the prompt again and goes to the background (or maybe is already in the background when it prints the prompt), and I need to run `fg` to get back to it.

abderrahim:
This is a port of 8f401127c450ea1c2a1e2878d2ac28a7166d355c to the buildbox-run sandbox

Thanks to @adoakley for investigating the issue. This was previously #41

Fixes #1690

### DIFF
--- a/src/buildstream/sandbox/_sandboxbuildboxrun.py
+++ b/src/buildstream/sandbox/_sandboxbuildboxrun.py
@@ -223,6 +223,18 @@ class SandboxBuildBoxRun(SandboxREAPI):
             except psutil.NoSuchProcess:
                 pass
 
+            if interactive:
+                # Make this process the foreground process again, otherwise the
+                # next read() on stdin will trigger SIGTTIN and stop the process.
+                # This is required because the sandboxed process does not have
+                # permission to do this on its own (running in separate PID namespace).
+                #
+                # tcsetpgrp() will trigger SIGTTOU when called from a background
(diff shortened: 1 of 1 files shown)


abderrahim:
@juergbi I guess you're the best placed to review as the author of the original commit.

abderrahim:
Seems a bit more complicated than this: there are a few test failures with `Inappropriate ioctl for device`. I wonder why, since I don't think we have tests that run bst interactively.

cs-shadow:
the logs have disappeared now so it's difficult to say now but I just wanted to respond to this bit:

> Seems a bit more complicated than this: there are a few test failures with Inappropriate ioctl for device. I wonder why, since I don't think we have tests that run bst interactively.

We actually do! I was very happy to add the first batch of them in https://gitlab.com/BuildStream/buildstream/-/merge_requests/1706 :)

it's possible `pexpect` didn't like deal well with this change.

abderrahim:
I pushed a fix to also check that stdin is a terminal before calling ioctl.

juergbi (approved):
LGTM and seems to work fine in a manual test.

Issue #1790, opened 2022-11-23
BuildStream 2 artifacts lose the unix write permission bit for the user (`u+w`)


doraskayo:
## Context
As far as I understand, BuildStream is supposed to preserve unix file permissions to a certain extent. However, the user write permission bit seems to be lost from files when checking out built artifacts. This issue is not seen with bst16.

## Description
When checking out artifacts using `--hardlinks` in bst2, it seems that the `u+w` permission is removed from every file (or almost every file?) compared to the permission for that file when it was originally collected as part of `bst build`.

From limited testing with the [Flatpak repo](https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/blob/9b70e597426135ae4b8631f80c1819677656271e/elements/flatpak-release.bst) built in the `freedesktop-sdk` project, it appears that only the user write bit disapears, while all other unix permission bits are retained.

Slightly unrelated to this, and potentially a different issue is when checking out artifacts _without_ `--hardlinks`, bst2 appears to set file permissions according to `umask`, resulting in file permissions which are yet again different from the file permissions that were present when the files were originally collected.

## Impact
Among other issues, this behavior leads OSTree repos which are collected into artifacts to be considered corrupted by `ostree fsck`. This is because OSTree repo include file permissions in their hash calculaition for objects, so this file permission change is detected as a file a corruption.

This file permission change and its effect on OSTree repos was found to be reason why the `freedesktop-sdk` project couldn't publish its Flatpak repo to its remote release server since moving from bst16 to bst2 in the `master` branch.

The only workaround found for this issue was to check out artifacts without `--hardlinks` and with a specific `umask` that provides the expected file permissions. However, this results in massive disk I/O overhead for the `freedesktop-sdk` project.

## References:
- The OSTree error appears when publishing in the `freedesktop-sdk` CI: (this is from the push job of the bst2 MR, where the issue was first seen)
  https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/jobs/2929504305#L46
- The OSTree error similarly appears when using the local `ostree fsck` command here:
  https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/jobs/3315563117#L1249

cc: @abderrahim, @gtristan, @nanonyme

nanonyme:
@juergbi we were discussing this with Slack. Seems some manner of reflinks would be the ideal solution, of course reflinks are not generally available.

nanonyme:
[...]

PR #1793, opened 2022-12-02
Use copy_file_range when exists

nanonyme:
This uses various optimizations for copy which are present including reflinks. Related to https://github.com/apache/buildstream/issues/1790

### DIFF
--- a/src/buildstream/utils.py
+++ b/src/buildstream/utils.py
@@ -70,6 +70,10 @@ _URI_SCHEMES = ["http", "https", "ftp", "file", "git", "sftp", "ssh"]
 _UMASK = os.umask(0o777)
 os.umask(_UMASK)
 
+# Only some operating systems have os.copy_file_range and even when present
+# it might not work
+_USE_CP_FILE_RANGE = hasattr(os, "copy_file_range")
+
 
 class UtilError(BstError):
     """Raised by utility functions when system calls fail.
@@ -357,6 +361,26 @@ def sha256sum(filename: str) -> str:
     return h.hexdigest()
 
 
+def _copy_file_range(src, dest):
+    global _USE_CP_FILE_RANGE  # pylint: disable=global-statement
+    if not _USE_CP_FILE_RANGE:
+        return False
+    with open(src, "rb") as src_file, open(dest, "wb") as dest_file:
+        num_bytes = os.fstat(src_file.fileno()).st_size
+        while num_bytes > 0:
+            try:
+                bytes_read = os.copy_file_range(src_file.fileno(), dest_file.fileno(), num_bytes)
+                if not bytes_read:
+                    return True
+                num_bytes -= bytes_read
+            except OSError as error:
+                if error.errno in (errno.ENOSYS, errno.EXDEV):
+                    _USE_CP_FILE_RANGE = False
+                    return False
+                raise error from None
+    return True
+
+
 def safe_copy(src: str, dest: str, *, copystat: bool = True, result: Optional[FileListResult] = None) -> None:
     """Copy a file while optionally preserving attributes
 
@@ -381,7 +405,9 @@ def safe_copy(src: str, dest: str, *, copystat: bool = True, result: Optional[Fi
             raise UtilError("Failed to remove destination file '{}': {}".format(dest, e)) from e
 
     try:
-        shutil.copyfile(src, dest)
+        ret = _copy_file_range(src, dest)
+        if not ret:
+            shutil.copyfile(src, dest)
     except (OSError, shutil.Error) as e:
         raise UtilError("Failed to copy '{} -> {}': {}".format(src, dest, e)) from e
 


src/buildstream/utils.py
```diff
@@ -359,6 +359,18 @@ def sha256sum(filename: str) -> str:
     return h.hexdigest()
 
 
+try:
+    os.copy_file_range  # type: ignore[attr-defined] Requires Python 3.8 or newer
+except AttributeError:
+    _copy_file = lambda src, dest: shutil.copyfile(src, dest)
```
abderrahim: Why not simply
```suggestion
    _copy_file = shutil.copyfile
```
 ?

src/buildstream/utils.py
```diff
@@ -359,6 +359,18 @@ def sha256sum(filename: str) -> str:
...
+else:
+    def _copy_file(src, dest):
+        with open(src, "rb") as src_file, open(dest, "wb") as dest_file:
+            num_bytes = os.fstat(src_file.fileno()).st_size
+            while num_bytes > 0:
+                num_bytes -= os.copy_file_range(src_file.fileno(), dest_file.fileno(), num_bytes)
```
juergbi: `copy_file_range()` may fail if the kernel is too old. The syscall was introduced in Linux 4.5 and support for cross-filesystem copies was added in Linux 5.3. Current glibc does not provide a user-space emulation if the kernel doesn't support it.

I.e. at least if `copy_file_range()` fails with `ENOSYS` or `EXDEV`, we need to support fallback to `shutil.copyfile()`.

src/buildstream/utils.py
nanonyme: Mypy complained the two functions have different amount of parameters.

src/buildstream/utils.py
nanonyme: I see. So we would need to wrap always, then fallback in the wrapper.

nanonyme: @juergbi it's a bit problematic there is zero CI testing of BuildStream outside when copy_file_range is supported.

src/buildstream/utils.py
nanonyme: I rewrote the entire piece of code.

src/buildstream/utils.py
nanonyme: I wrote code to fallback during runtime when not supported.

src/buildstream/utils.py
```diff
@@ -357,6 +358,21 @@ def sha256sum(filename: str) -> str:
...
+    with open(src, "rb") as src_file, open(dest, "wb") as dest_file:
+        num_bytes = os.fstat(src_file.fileno()).st_size
+        while num_bytes > 0:
+            try:
+                num_bytes -= os.copy_file_range(src_file.fileno(), dest_file.fileno(), num_bytes)
+            except OSError as error:
+                if error.errno in (errno.ENOSYS, errno.EXDEV):
+                    return False
+                raise error from None
+    return True
```
nanonyme, line 381: Essentially how this works here is:
1. Return True implies there was a success and copy_file_range was supported
2. Return False implies copy_file_range was not supported and there should be a fallback
3. Any other exception implies failure and does not result in fallbacks

nanonyme:
I was suggested on Btrfs that that helpful commands for figuring out if this MR actually works are

 "filefrag -v foobar | grep shared", btrfs fi du foobar, "btrfs-search-metadata file foobar" 

I will verify later once I have disk that actually supports reflinks.

nanonyme:
I have now been testing this locally and I'm not convinced things are working as expected. There is *very* high IO going on.

nanonyme:
Tested, this makes checkout lightning fast on Btrfs.

nanonyme:
We now have prototype CI image available through https://gitlab.com/freedesktop-sdk/infrastructure/freedesktop-sdk-docker-images/-/merge_requests/379 as registry.gitlab.com/freedesktop-sdk/infrastructure/freedesktop-sdk-docker-images/bst2:dev-nanonyme-checkout-reflinks with these changes

src/buildstream/utils.py
```diff
@@ -357,6 +358,21 @@ def sha256sum(filename: str) -> str:
...
+    if not _USE_CP_FILE_RANGE:
+        return False
+    with open(src, "rb") as src_file, open(dest, "wb") as dest_file:
+        num_bytes = os.fstat(src_file.fileno()).st_size
+        while num_bytes > 0:
+            try:
+                num_bytes -= os.copy_file_range(src_file.fileno(), dest_file.fileno(), num_bytes)
+            except OSError as error:
+                if error.errno in (errno.ENOSYS, errno.EXDEV):
+                    return False
```
juergbi, line 379: Should we set `_USE_CP_FILE_RANGE` to `False` (using `global`) to avoid the overhead of repeatedly failing when copying a large number of files?

src/buildstream/utils.py
```diff
@@ -357,6 +358,21 @@ def sha256sum(filename: str) -> str:
...
 
 
+def copy_file_range(src, dest):
+    if not _USE_CP_FILE_RANGE:
+        return False
+    with open(src, "rb") as src_file, open(dest, "wb") as dest_file:
+        num_bytes = os.fstat(src_file.fileno()).st_size
+        while num_bytes > 0:
+            try:
+                num_bytes -= os.copy_file_range(src_file.fileno(), dest_file.fileno(), num_bytes)
```
juergbi: We should detect `os.copy_file_range()` returning zero, indicating end of file. This can only happen if the source file was truncated but it would lead to an infinite loop.

src/buildstream/utils.py
```diff
@@ -381,9 +397,14 @@ def safe_copy(src: str, dest: str, *, copystat: bool = True, result: Optional[Fi
...
     try:
-        shutil.copyfile(src, dest)
+        ret = copy_file_range(src, dest)
     except (OSError, shutil.Error) as e:
         raise UtilError("Failed to copy '{} -> {}': {}".format(src, dest, e)) from e
+    if not ret:
+        try:
+            shutil.copyfile(src, dest)
+        except (OSError, shutil.Error) as e:
+            raise UtilError("Failed to copy '{} -> {}': {}".format(src, dest, e)) from e
```
juergbi: Let's merge the two identical `except` clauses. I first wanted to suggest to keep the separate `except` clauses but use different error messages. However, if e.g. `open()` fails in `copy_file_range()`, raising a `copy_file_range()`-specific error message may be confusing.

juergbi (changes requested):

src/buildstream/utils.py
nanonyme: I can do that, sure.

src/buildstream/utils.py
nanonyme: Fixed.

src/buildstream/utils.py
nanonyme: Fixed.

src/buildstream/utils.py
nanonyme: Fixed.

nanonyme:
@juergbi the ostree source track test is randomly failing. I'm reasonably sure it's unrelated to this change.

BenjaminSchubert (approved):
That looks good to me, I now wonder if we would have benefits from cythonizing this one if it's a bottleneck?

juergbi (approved):

PR #1805, opened 2022-12-28
Remove deprecated etag

### LINKED ISSUE #1795: Remove deprecated etag field from downloadablefilesource
It makes sense to do this before bst2 final.

nanonyme:
Fixes https://github.com/apache/buildstream/issues/1795. It was already deprecated in bst1 so let's just remove it.


gtristan:
Good call thanks :)

PR #1807, opened 2022-12-29
requirements.txt: Bump grpc version to be compatible with Python 3.11

cs-shadow:
missed this in #1806

### DIFF
--- a/requirements/requirements.txt
+++ b/requirements/requirements.txt
@@ -1,5 +1,5 @@
 click==8.1.3
-grpcio==1.46.0
+grpcio==1.49.1
 Jinja2==3.1.2
 pluginbase==1.0.1
 protobuf==4.21.12


cs-shadow:
seems like our testsuite is busted with the new version of grpc, draft for now

nanonyme:
Yeah, there's similarly issues with bst1. In there it looked there was some general timeout in asyncio reactor getting hit with timeout of 30 minutes *per test*.

nanonyme:
As per one of the CI runs that actually *exited*, it looks like there's leftover grpcio threads from tests. We're not running bst like normal (launch bst as a subprocess) but instead running bst main function inside pytest. This leaks grpcio into pytest test runner which can result in all kinds of trouble if it doesn't properly terminate. (but which real users of BuildStream can never encounter) I tried refactoring earlier the setup to instead call bst through subprocess but it turned out we have some exception analysis in tests that is impossible to do if you aren't in same process as bst.

juergbi:
Replaced by #1815, which fixes the tests.

PR #1812, opened 2023-01-16
elementsourcescache: don't try catching grpc.RpcError

abderrahim:
Nothing that it calls raises this exception; everything should raise either AssetCacheError or CasRemoteError.


gtristan:
Nice one, I've been wanting to clean these up as well...

PR #1813, opened 2023-01-16
downloadablefilesource: don't download files if they are already cached

abderrahim:


juergbi (approved):

abderrahim:
This breaks in practice because pickling exceptions thrown in a blocking_activity doesn't work. So this actually depends on #1814 (or a similar solution)

gtristan:
These changes are incorporated in #1814 already, closing.

abderrahim:
> These changes are incorporated in https://github.com/apache/buildstream/pull/1814 already, closing.

I found out they are not: the bug is still present in 2.0.1. I opened #1834 as I can't reopen this.

PR #1814, opened 2023-01-16
plugin: don't try to pickle the exceptions thrown by blocking_activity

abderrahim:


nanonyme:
@abderrahim we could send exception class and args, then reconstruct exception on other side without stack trace.

nanonyme:
But this is certainly an improvement already.

PR #1820, opened 2023-01-29
Stop passing URL opener across process boundaries

### LINKED ISSUE #1766: Traceback bug in artifact fetch

nanonyme:
Should fix https://github.com/apache/buildstream/issues/1766

The object is not picklable. This cannot be supported. Instead create URL opener when needed in child process. Drops URL opener caching which may result in minor performance impact.

### DIFF
--- a/src/buildstream/downloadablefilesource.py
+++ b/src/buildstream/downloadablefilesource.py
@@ -88,7 +88,8 @@ class _NetrcPasswordManager:
             return login, password
 
 
-def _download_file(opener, url, etag, directory):
+def _download_file(opener_creator, url, etag, directory):
+    opener = opener_creator.get_url_opener()
     default_name = os.path.basename(url)
     request = urllib.request.Request(url)
     request.add_header("Accept", "*/*")
@@ -120,7 +121,6 @@ class DownloadableFileSource(Source):
 
     COMMON_CONFIG_KEYS = Source.COMMON_CONFIG_KEYS + ["url", "ref", "etag"]
 
-    __urlopener = None
     __default_mirror_file = None
 
     def configure(self, node):
@@ -217,8 +217,10 @@ class DownloadableFileSource(Source):
                 else:
                     etag = None
 
+                url_opener_creator = _UrlOpenerCreator(self._parse_netrc())
+
                 local_file, new_etag = self.blocking_activity(
-                    _download_file, (self.__get_urlopener(), self.url, etag, td), activity_name
+                    _download_file, (url_opener_creator, self.url, etag, td), activity_name
                 )
 
                 if local_file is None:
@@ -251,6 +253,21 @@ class DownloadableFileSource(Source):
             # ValueError for unknown url types, so we handle it here.
             raise SourceError("{}: Error mirroring {}: {}".format(self, self.url, e), temporary=True) from e
 
+    def _parse_netrc(self):
+        netrc_config = None
+        try:
+            netrc_config = netrc.netrc()
+        except OSError:
+            # If the .netrc file was not found, FileNotFoundError will be
+            # raised, but OSError will be raised directly by the netrc package
+            # in the case that $HOME is not set.
+            #
+            # This will catch both cases.
+            pass
+        except netrc.NetrcParseError as e:
+            self.warn("{}: While reading .netrc: {}".format(self, e))
+        return netrc_config
+
     def _get_mirror_file(self, sha=None):
         if sha is not None:
             return os.path.join(self._mirror_dir, sha)
@@ -260,29 +277,15 @@ class DownloadableFileSource(Source):
 
         return self.__default_mirror_file
 
-    @classmethod
-    def _reset_url_opener(cls):
-        # Needed for tests, in order to cleanup the `netrc` configuration.
-        cls.__urlopener = None  # pylint: disable=unused-private-member
 
-    def __get_urlopener(self):
-        if not DownloadableFileSource.__urlopener:
-            try:
-                netrc_config = netrc.netrc()
-            except OSError:
-                # If the .netrc file was not found, FileNotFoundError will be
-                # raised, but OSError will be raised directly by the netrc package
-                # in the case that $HOME is not set.
-                #
-                # This will catch both cases.
-                #
-                DownloadableFileSource.__urlopener = urllib.request.build_opener()
-            except netrc.NetrcParseError as e:
-                self.warn("{}: While reading .netrc: {}".format(self, e))
-                return urllib.request.build_opener()
-            else:
-                netrc_pw_mgr = _NetrcPasswordManager(netrc_config)
-                http_auth = urllib.request.HTTPBasicAuthHandler(netrc_pw_mgr)
-                ftp_handler = _NetrcFTPOpener(netrc_config)
-                DownloadableFileSource.__urlopener = urllib.request.build_opener(http_auth, ftp_handler)
-        return DownloadableFileSource.__urlopener
+class _UrlOpenerCreator:
+    def __init__(self, netrc_config):
+        self.netrc_config = netrc_config
+
+    def get_url_opener(self):
+        if self.netrc_config:
+            netrc_pw_mgr = _NetrcPasswordManager(self.netrc_config)
+            http_auth = urllib.request.HTTPBasicAuthHandler(netrc_pw_mgr)
+            ftp_handler = _NetrcFTPOpener(self.netrc_config)
+            return urllib.request.build_opener(http_auth, ftp_handler)
+        return urllib.request.build_opener()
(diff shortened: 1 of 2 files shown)


nanonyme:
Looks like this breaks netrfc functionality. Need to debug.

nanonyme:
I suspect this is CI-only problem. Environment is working in unexpected way.

nanonyme:
Okay, done. I think there is some problem with buildgrid that is causing test failures.

src/buildstream/downloadablefilesource.py
```diff
@@ -260,29 +260,29 @@ def _get_mirror_file(self, sha=None):
...
+            pass
+
+    def get_url_opener(self):
+        if self.netrc_config:
+            netrc_pw_mgr = _NetrcPasswordManager(self.netrc_config)
+            http_auth = urllib.request.HTTPBasicAuthHandler(netrc_pw_mgr)
+            ftp_handler = _NetrcFTPOpener(self.netrc_config)
+            return urllib.request.build_opener(http_auth, ftp_handler)
+        else:
+            return urllib.request.build_opener()
```
BenjaminSchubert: A small nitpick but I think

```suggestion
        return urllib.request.build_opener()
```

would be a tad easier to read and is more in line with what we do in other places

src/buildstream/downloadablefilesource.py
```diff
@@ -260,29 +260,29 @@ def _get_mirror_file(self, sha=None):
...
+            self.netrc_config = netrc.netrc()
+        except OSError:
+            # If the .netrc file was not found, FileNotFoundError will be
+            # raised, but OSError will be raised directly by the netrc package
+            # in the case that $HOME is not set.
+            #
+            # This will catch both cases.
+            pass
+        except netrc.NetrcParseError:
+            # netrc file exists but is invalid, passthrough since we can't log here
```
BenjaminSchubert: Don't we have any way of reporting this? It might become very hard to debug for users if they have a typo in it.

BenjaminSchubert (commented):
I like the change, but I am a bit cautious about the invalid netrc error, I am not sure I get why this can't be logged anymore?

src/buildstream/downloadablefilesource.py
nanonyme: The problem is we no longer have access to source here so we can't call BuildStream logging primitives. It never worked anyway

src/buildstream/downloadablefilesource.py
nanonyme: Not unless we make this fatal

nanonyme: Wait, actually, that was not quite true. netrc processing happens in main project side. I will refactor this. still a bit.

src/buildstream/downloadablefilesource.py
nanonyme: Fixed.

src/buildstream/downloadablefilesource.py
nanonyme: I decoupled netrc parsing into a method in DownloadableFileSource since it was anyway parsed in main process so now it has safely access to logging.

BenjaminSchubert (approved):

nanonyme:
Failing test appears to have been failed at ` [00:00:00] FAILURE [4a47c98a] build-shell/buildtree.bst: Failed to link '/home/testuser/buildstream/.tox/py310/tmp/test_shell_pull_cached_buildtr2/cache/artifacts/refs/test/build-shell-buildtree/4a47c98a10df39e65e99d471f96edc5b58d4ea5b9b1f221d0be832a8124b8099 -> /home/testuser/buildstream/.tox/py310/tmp/test_shell_pull_cached_buildtr2/cache/artifacts/refs/test/build-shell-buildtree/38b730aec720dbddba843c9e3ce1ad8073e39d16cc4d25f752bc89d867fbaec1': [Errno 17] File exists: '/home/testuser/buildstream/.tox/py310/tmp/test_shell_pull_cached_buildtr2/cache/artifacts/refs/test/build-shell-buildtree/4a47c98a10df39e65e99d471f96edc5b58d4ea5b9b1f221d0be832a8124b8099' -> '/home/testuser/buildstream/.tox/py310/tmp/test_shell_pull_cached_buildtr2/cache/artifacts/refs/test/build-shell-buildtree/38b730aec720dbddba843c9e3ce1ad8073e39d16cc4d25f752bc89d867fbaec1'`

abderrahim:
> Failing test appears to have been failed at [...]

Which seems to be the same as https://github.com/apache/buildstream/pull/1814. I guess that's just a flaky test.

nanonyme:
It looks strange. Apparently the error originates here https://github.com/apache/buildstream/blob/master/src/buildstream/utils.py#L432 and it looks like bst should upon file existing try to delete it and then try to link again.

nanonyme:
It definitely looks like a race condition with multiple parties trying create same link. Supposedly a chain
1. A creates link
2. B tries to create link and gets error it exists
3. C tries to create link and gets error it exists
4. B deletes link
5. C tries to delete link and silently fails
6. B creates link
7. C tries to create link and throws exception

can result in this failure. safe_link is not multi-thread or multi-process safe.

nanonyme:
The should probably be a follow-up issue on that. I suspect there might be something specific to test runner that results in this since I haven't ever seen this in real world.

gtristan:
Rebased this and fixed conflicts with downloadablefilesource.py, this PR is now replaced by #1826

PR #1825, opened 2023-02-12
Remove guard and just always fail if file is gone

nanonyme:
Someone else is operating on same file as us in this case and proceeding is dangerous


nanonyme:
Right, there is a clear race condition `Failed to remove destination file '/home/testuser/buildstream/doc/run-bst-jbglffex/build/workspace-helloi42ufm_9/Makefile': [Errno 2] No such file or directory: '/home/testuser/buildstream/doc/run-bst-jbglffex/build/workspace-helloi42ufm_9/Makefile'`

PR #1827, opened 2023-02-21
Add support for python 3.11

### LINKED ISSUE #1758: bst1: buildstream is not compatible with python 3.11

gtristan:
Fixes #1758

This branch expands on @nanonyme 's branch and makes most tests pass, with a few tests set to xfail due to limitations such as not being able to use `unittest.mock()`.

This replaces #1785

### DIFF
--- a/.github/common.env
+++ b/.github/common.env
@@ -1,5 +1,5 @@
 # Shared common variables
 
 CI_IMAGE_VERSION=master-784208155
-CI_TOXENV_MAIN=py36-nocover,py37-nocover,py38-nocover,py39-nocover,py310-nocover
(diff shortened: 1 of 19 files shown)


gtristan:
The main current remaining issue here is that fedora fails to provide the /usr/bin/fusermount symlink to /usr/bin/fusermount3.

This is being worked around in https://gitlab.com/BuildStream/buildstream-docker-images/-/merge_requests/216

PR #1834, opened 2023-03-23
downloadablefilesource: don't download files if they are already cached

abderrahim:


doraskayo:
I noticed this as well at some point. It looks like a refactor flipped the logic by accident.

PR #1836, opened 2023-03-25
Use urlretrieve

### LINKED ISSUE #1835: downloadablefilesource does not validate file is not shorter than Content-Lenght
https://github.com/apache/buildstream/blob/master/src/buildstream/downloadablefilesource.py#L125 is catching ContentTooShortError. This doesn't actually work because only high-level function urlretrieve raises it while the lower-level functions we're using do not. We need to manually check file is at least the size of Content-Length. It is commonly allowed file is larger than that.

nanonyme:
Fixes https://github.com/apache/buildstream/issues/1835

### DIFF
--- a/src/buildstream/downloadablefilesource.py
+++ b/src/buildstream/downloadablefilesource.py
@@ -99,21 +99,9 @@ def _download_file(opener_creator, url, etag, directory):
         request.add_header("If-None-Match", etag)
 
     try:
-        with contextlib.closing(opener.open(request)) as response:
-            info = response.info()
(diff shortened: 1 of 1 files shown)


nanonyme:
I'm not entirely sure if this can work. Looks possible the API does not in fact support passing request which would mean we can't use it.

nanonyme:
This cannot work indeed. I have filed feature request for this against CPython since it's a useful thing for urlretrieve but BuildStream cannot assume that exists.

PR #1846, opened 2023-06-11
Raise ContentTooShortError if we don't get enough bytes

### LINKED ISSUE #1835: downloadablefilesource does not validate file is not shorter than Content-Lenght

nanonyme:
This can happen if connection is aborted before we finish the download. urlretrieve higher level API would handle this but we don't use it. Fixes https://github.com/apache/buildstream/issues/1835

### DIFF
--- a/src/buildstream/downloadablefilesource.py
+++ b/src/buildstream/downloadablefilesource.py
@@ -107,12 +107,17 @@ def _download_file(opener_creator, url, etag, directory):
                 return None, None, None
 
             etag = info["ETag"]
+            length = info.get("Content-Length")
 
             filename = info.get_filename(default_name)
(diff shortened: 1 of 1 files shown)


nanonyme:
@gtristan can we have this landed soon? Retry logic with connection failure is currently quite broken.

abderrahim (approved):
It would be nice to be able to continue a download, especially with larger files. But this is an improvement already.

PR #1847, opened 2023-06-26
Add EngFlow to list of RE servers

plaindocs:
Hey folks,

Disclaimer: I'm an EngFlow employee 

I found this in my travels and thought I'd add to the list.

### DIFF
--- a/doc/source/using_configuring_remote_execution.rst
+++ b/doc/source/using_configuring_remote_execution.rst
@@ -26,6 +26,8 @@ Some of these implementations include:
 * `BuildGrid <https://buildgrid.build/>`_
 * `BuildBarn <https://github.com/buildbarn>`_
 * `Buildfarm <https://github.com/bazelbuild/bazel-buildfarm>`_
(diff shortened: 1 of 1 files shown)


nanonyme:
Has it been tested with this?

plaindocs:
I believe so, but don't have a link to proof.

It's mentioned on the main website https://www.engflow.com/product/remoteExecution?r=menu****

image

I can ask around if you need something more specifc?

nanonyme:
It's just that simply "Supports Remote Execution API's" has historically not been enough to mean it actually works due to various implementation details. Afaik there is no automated end-to-end testing in context of this project with EngFlow (assumably due to lack of testing cluster) so I was asking if that had been done by someone manually.

abderrahim:
> I can ask around if you need something more specifc?

I think the main thing buildstream needs is the ability to use the input root as the filesystem root. Like the `chrootIntoInputRoot` option of buildbarn.

Then some testing should indeed be in order just to validate that it indeed works.

PR #1862, opened 2023-08-20
source.py: Check for the availability of alias targets in preflight()

### LINKED ISSUE #1860: Buildstream crashes when instructed to only download from mirrors
Trying to add
```
fetch:
  source: mirrors
```
to my buildstream.conf causes buildstream to crash while fetching sources:

```
[00:00:00][400f0914][   fetch:plugins/bst-plugins-experimental.bst] BUG     Fetch
    
      File "/usr/lib/python3.10/site-packages/buildstream/_scheduler/jobs/job.py", line 438, in child_action
      File "/usr/lib/python3.10/site-packages/buildstream/_scheduler/jobs/elementjob.py", line 91, in child_process
      File "/usr/lib/python3.10/site-packages/buildstream/_scheduler/queues/fetchqueue.py", line 75, in _fetch_not_original
        element._fetch(fetch_original=False)
      File "/usr/lib/python3.10/site-packages/buildstream/element.py", line 2207, in _fetch
        self.__sources.fetch()
      File "/usr/lib/python3.10/site-packages/buildstream/_elementsources.py", line 222, in fetch
        self.fetch_sources()
      File "/usr/lib/python3.10/site-packages/buildstream/_elementsources.py", line 251, in fetch_sources
        self._fetch_source(source)
      File "/usr/lib/python3.10/site-packages/buildstream/_elementsources.py", line 432, in _fetch_source
        source._fetch()
      File "/usr/lib/python3.10/site-packages/buildstream/source.py", line 908, in _fetch
        self.__do_fetch()
      File "/usr/lib/python3.10/site-packages/buildstream/source.py", line 1418, in __do_fetch
        raise last_error
    UnboundLocalError: local variable 'last_error' referenced before assignment

```

gtristan:
Fixes #1860

### DIFF
--- a/src/buildstream/source.py
+++ b/src/buildstream/source.py
@@ -308,6 +308,15 @@ class SourceFetcher:
 
         Args:
            url: The url used to download.
+
+        .. note::
+
+           While this must be called in a SourceFetcher initializer for the URL which
+           will be used by the fetcher, note that any URLs which are known and specified
+           in the Source configuration YAML must be marked with either
+           :func:`Source.mark_download_url() <buildstream.source.Source.mark_download_url>` or
+           :func:`Source.translate_url() <buildstream.source.Source.translate_url>` in
+           the :func:`Plugin.configure() <buildstream.plugin.Plugin.configure>` implementation.
         """
         self.__alias = _extract_alias(url)
 
@@ -791,6 +800,18 @@ class Source(Plugin):
                 raise SourceError(
                     "{}: Invalid alias '{}' specified in URL: {}".format(self, alias, url),
                     reason="invalid-source-alias",
+                )
+            if not project.get_alias_uris(alias, first_pass=self.__first_pass, tracking=False):
+                raise SourceError(
+                    "{}: No fetch URI found for alias '{}'".format(self, alias),
+                    detail="Check fetch controls in your user configuration",
+                    reason="missing-source-alias-target",
+                )
+            if not project.get_alias_uris(alias, first_pass=self.__first_pass, tracking=True):
+                raise SourceError(
+                    "{}: No tracking URI found for alias '{}'".format(self, alias),
+                    detail="Check track controls in your user configuration",
+                    reason="missing-source-alias-target",
                 )
 
     def get_project_directory(self) -> str:
@@ -1401,6 +1422,7 @@ class Source(Plugin):
                 self.fetch(**kwargs)
                 return
 
+            last_error = None
             for uri in project.get_alias_uris(alias, first_pass=self.__first_pass, tracking=False):
                 new_source = self.__clone_for_uri(uri)
                 try:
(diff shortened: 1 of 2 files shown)


gtristan:
Looks like there are issues with calling `Source.get_source_fetchers()` in preflight... might be better off raising a lazy error here rather than a preflight error...

gtristan:
> Looks like there are issues with calling `Source.get_source_fetchers()` in preflight... might be better off raising a lazy error here rather than a preflight error...

Tried a lazy failing approach, that works but still wasn't satisfied with this.

Came back and did the check now in `Source.mark_download_url()`, verifying the correctness of this in the Source API contract.

Added a comment to `SourceFetcher.mark_download_url()` to reinforce the already existing contract, which already specified
that *"This must be called for every URL in the configuration during `Plugin.configure()` if `Source.translate_url()` is not called."*

This existing contract already validates the correctness of this approach to policing the alias, however it helps to reinforce the statement in the `SourceFetcher` docs and avoid misguiding plugin authors into thinking that `SourceFetcher.mark_download_url()` is a *replacement* of calling `Source.mark_download_url()`

**Caveat**: With this approach (including the previous approach), Buildstream will fail at load time in the case that user configuration specifies, for example, invalid tracking configuration (where aliases don't have specified URIs for tracking) when `bst build` or `bst source fetch` is called (i.e. it didn't *need* to be an error for the *purpose* of which bst was launched).

I think this is generally okay because you have to go out of your way to specify invalid tracking information anyway, so you shouldn't have invalid tracking configuration.

The current error in the test for this patch revision is:

```
fetch_source source at test.bst [line 3 column 2]: No fetch URI found for alias 'foo'

    Check fetch controls in your user configuration
```

src/buildstream/source.py
```diff
@@ -308,6 +308,15 @@ def mark_download_url(self, url: str) -> None:
 
         Args:
            url: The url used to download.
+
+        .. note::
+
+           While this must be called in a SourceFetcher initializer for the URL which
```
abderrahim, line 314: Looks like this isn't consistently done. In buildstream-plugins, the `cargo` plugin does it but the `git` plugin doesn't. In bst-plugins-experimental, all plugins using source fetchers (`git_tag`, `git_repo` and `pypi`) do call it in the initializer.

I noticed however that the `git` plugin calls it from get_source_fetchers() instead, which is weird but probably gives the same end result?

abderrahim (approved):

src/buildstream/source.py
gtristan: `SourceFetcher.mark_download_url()` must be called for download URLs and is **not** a replacement for the required call to `Source.mark_download_url()` at `Plugin.configure()` time.

The `git` plugin technically must call `Source.mark_download_url()` *"for any URL that is specified in the source configuration"*, i.e. not necessarily for URLs which it will discover later by virtue of having the git module present and discovering submodules.

Technically without calling this, BuildStream core is unable to validate aliases which are used in the URL configurations in the yaml (although this *reason* need not be communicated as the sole purpose in the contract).

And... so it does in fact: https://github.com/apache/buildstream-plugins/blob/master/src/buildstream_plugins/sources/git.py#L749

The `git` plugin does indeed call `SourceFetcher.mark_download_url()` from the `Source.get_source_fetchers()` method instead of in the individual `SourceFetcher` initializers... and I would agree that it does not make a difference, it is more important that it gets called before the `SourceFetcher` is handed over to the core.

abderrahim: I might have gotten things mixed up, I'll try to have a deeper look this evening.

PR #1863, opened 2023-08-21
Updating versioneer again, properly.

### LINKED ISSUE #1855: Can't build with Python 3.12
Buildstream can't build with Python 3.12.

Python 3.12 deprecated a few things, among which is `configparser.SafeConfigParser`.

https://docs.python.org/3.12/whatsnew/3.12.html#removed

Pretty sure the `versioneer.py` script could instead say something like : 

```
    setup_cfg = os.path.join(root, "setup.cfg")
    try;
        if sys.version_info.major == 3 and sys.version_info.minor >= 12:
            from configparser import ConfigParser

        else:
            from configparser import SafeConfigParser as ConfigParser
```

gtristan:
First time around we were using `versioneer install --no-vendor`, and neglecting to remove the `versioneer.py` file from the toplevel, which is still required by BuildStream's setup.py for a few reasons.

Instead, go back to `versioneer install --vendor` and update versioneer files correctly.

Fixes #1855

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,8 +20,6 @@ requires = [
     # Require a new enough cython for the python versions we support
     "Cython>=0.29.25",
     "packaging",
-    # Versioneer is a build time dependency
-    "versioneer[toml]",
 ]
 build-backend = "setuptools.build_meta"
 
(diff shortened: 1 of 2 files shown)


Callek (approved):
As the versioneer maintainer, I approve.  Though I note we do have `0.29` out now (Released in July), which could be used instead of `0.28` at your discretion.

PR #1864, opened 2023-08-22
Raise exception with more information upon plugin bug

### LINKED ISSUE #1783: Unclear output from BuildStream when translate_url called twice with different URL's as primary
BuildStream in this case hits assert (bad because this might get optimized out, we should instead aise non-retryable SourceError) and also we must add better error context with the variables to make it possible to debug plugin bugs. https://github.com/apache/buildstream/blob/master/src/buildstream/source.py#L754

nanonyme:
Fixes https://github.com/apache/buildstream/issues/1783

### DIFF
--- a/src/buildstream/source.py
+++ b/src/buildstream/source.py
@@ -760,7 +760,9 @@ class Source(Plugin):
 
                 assert (
                     self.__expected_alias is None or self.__expected_alias == expected_alias
-                ), "Primary URL marked twice with different URLs"
+                ), "Attempt to mark primary URL with {}, already marked with {}".format(
+                    expected_alias, self.__expected_alias
+                )
 
                 self.__expected_alias = expected_alias
 


src/buildstream/source.py
```diff
@@ -749,9 +749,12 @@ def mark_download_url(self, url: str, *, primary: bool = True) -> None:
             if primary:
                 expected_alias = _extract_alias(url)
 
-                assert (
-                    self.__expected_alias is None or self.__expected_alias == expected_alias
-                ), "Primary URL marked twice with different URLs"
+                if self.__expected_alias is not None and self.__expected_alias != expected_alias:
+                    raise BstError(
```
gtristan: In this case we explicitly do not want a `BstError`, as these are intended to communicate a clear reason for a failure to the user so that the user can potentially address the problem.

In this case we want it to bubble up as a regular exception, so that the core will treat it as a `BUG` and print a stack trace, which is more helpful for plugin authors to address bugs in their plugins.

gtristan: By the way, if anything... there are a number of places where we explicitly use `assert` for this purpose, but it would be desirable to change *all* of these cases for an explicit (not `BstError` inheriting) exception, because there are conceivable ways to run python code with `assert` lines disabled (although that doesn't really ever happen in practice as far as I know).

nanonyme: Added a new exception called Bug.

gtristan: Unless we're going to make a sweeping change that changes all of our asserts where we intend to result in a BUG message with accompanying stack trace, I'm not happy with adding the new `Bug` exception.

I.e. it will be easily forgotten and will not be an obvious matter of policy.

Also, the problem with this we have to realize is that unfortunately, BuildStream has dependencies.

So if one wanted to run BuildStream without assertions (e.g. using `python3 -O ...` or with `PYTHONOPTIMIZE=1`), we would then be ignoring any assert statements from any of our dependencies... I'm happy to be convinced otherwise, but I think in the wild west of python dependencies, nobody really cares about supporting running with assertions disabled :-/

For this patch, I think you really just wanted to add some context to the message, it would probably be best to just update the message in the existing assertion.

I'm sorry for not having thought this through more deeply when you originally submitted this :-(

nanonyme: Fixed.

nanonyme:
@gtristan so is this now okay after your requested changes?

gtristan:
Yes thankyou ! Sorry again for leading you down a rabbit hole

PR #1865, opened 2023-08-24
job.py: catch Exception raised when opening the log file

abderrahim:
When the disk space is insufficient, this would otherwise terminate the job
thread without the scheduler being aware, which would lead to buildstream
hanging.
    
Related to https://github.com/apache/buildstream/issues/660

### DIFF
--- a/src/buildstream/_scheduler/jobs/job.py
+++ b/src/buildstream/_scheduler/jobs/job.py
@@ -22,6 +22,7 @@ import datetime
 import itertools
 import threading
 import traceback
+from contextlib import ExitStack
 
 # BuildStream toplevel imports
 from ... import utils
@@ -312,11 +313,26 @@ class Job:
             self.action_name, self._message_element_name, self._message_element_key
         )
 
-        # Time, log and and run the action function
-        #
-        with self._messenger.timed_suspendable() as timeinfo, self._messenger.recorded_messages(
-            self._logfile, self._scheduler.context.logdir
-        ) as filename:
+        with ExitStack() as stack:
+            # Time, log and and run the action function
+            #
+            timeinfo = stack.enter_context(self._messenger.timed_suspendable())
+
+            try:
+                filename = stack.enter_context(
+                    self._messenger.recorded_messages(self._logfile, self._scheduler.context.logdir)
+                )
+            except Exception as e:  # pylint: disable=broad-except
+                elapsed = datetime.datetime.now() - timeinfo.start_time
+                self.message(
+                    MessageType.ERROR,
+                    "Error opening log file: {}".format(e),
+                    elapsed=elapsed,
+                    detail=traceback.format_exc(),
+                )
+                self._thread_id = None
+                return _ReturnCode.PERM_FAIL, None
+
             try:
                 self.message(MessageType.START, self.action_name, logfile=filename)
 


abderrahim:
I just saw that there is an except block a bit later in the file that's supposed to catch such things. But unfortunately, the thing that's raising an exception for me isn't in the try block.

https://github.com/apache/buildstream/blob/0e0923f7c8ad3812159c48538b663c6abea2e7ee/src/buildstream/_scheduler/jobs/job.py#L476

The exception I hit is a "No space left on device" raised by `self._messenger.recorded_messages(self._logfile, self._logdir) as filename`

gtristan:
> I just saw that there is an except block a bit later in the file that's supposed to catch such things. But unfortunately, the thing that's raising an exception for me isn't in the try block.
> 
> https://github.com/apache/buildstream/blob/0e0923f7c8ad3812159c48538b663c6abea2e7ee/src/buildstream/_scheduler/jobs/job.py#L476
> 
> The exception I hit is a "No space left on device" raised by `self._messenger.recorded_messages(self._logfile, self._logdir) as filename`

Aha... ok so the `try` block around `result = self.child_process()` runs in a it's own thread, and handles the blind `except Exception` clause in order to properly marshal bugs/crashes through the messaging system...

This is the main area where we need to be very careful with exceptions, as an unhandled one here can result in us not cleaning up `self._thread_id` and not returning properly from that function, leaving a stale entry in the asyncio event loop.

It looks like we're handling errors which can potentially arise from untrusted client code implemented by elements well, but we're not being careful enough for IO errors which can happen in the core proper, so we need to catch exceptions when entering the `child_action()` function... in your case around `self._messenger.recorded_messages()`.

I would suggest that for readability we might use `ExitStack()` here, such that we can use try blocks for entering the context managers, e.g.:

```python
with ExitStack() as stack:

    try:
        filename = stack.enter_context(
            self._messenger.recorded_messages(self._logfile, self._logdir)
        )
    except Exception:
        elapsed = datetime.datetime.now() - timeinfo.start_time
        self.message(MessageType.ERROR, "{}: Error opening log file".format(self.action_name),
                              elapsed=elapsed,
                              detail=traceback.format_exc())
        self._thread_id = None
        return _ReturnCode.PERM_FAIL, None
```

Maybe it would be better to have `Messenger.recorded_messages()` do it's own internal try/catch and format a nicer error message into a nicer exception to handle in this block too.

Note this is related to #660

src/buildstream/_scheduler/jobs/job.py
```diff
@@ -312,11 +313,26 @@ def child_action(self):
...
+
+            try:
+                filename = stack.enter_context(
+                    self._messenger.recorded_messages(self._logfile, self._scheduler.context.logdir)
+                )
+            except Exception:
+                elapsed = datetime.datetime.now() - timeinfo.start_time
+                self.message(
+                    MessageType.ERROR,
+                    "Error opening log file",
```
gtristan: Could be a bit better to have `except Exception as e` and `"Error opening log file: {}".format(e)` here.

It could be (I'm not altogether sure off the top of my head) that the logging configuration might truncate the "detail" string if it has too many lines, so we *might* not get the entire stack trace there.

gtristan: Pushed this fix myself and merging.

PR #1867, opened 2023-09-01
Remove unnecessary dependency on dateutil

### LINKED ISSUE #1866: Remove unused function _parse_timestamp()
The usage of `utils._parse_timestamp()` was removed in 9826e115a4bb6f44524145c33f82ca4af13847e3, the only thing using it now is the tests. The tests can be adjusted to use unix timestamps directly, rather than requiring the use of this function.

If this function is removed, we can remove the dependency on `python-dateutil`, and subsequently the dependency on `six`.

benjamb:
Remove unnecessary utility functions for the parsing and generating of ISO 8601 timestamps since they are no longer used outside of the tests, which have been instead adapted to use unix timestamps directly.

Closes https://github.com/apache/buildstream/issues/1866.

### DIFF
--- a/requirements/requirements.in
+++ b/requirements/requirements.in
@@ -9,4 +9,3 @@ ruamel.yaml.clib >= 0.1.2
 setuptools
 pyroaring
 ujson
-python-dateutil >= 2.7.0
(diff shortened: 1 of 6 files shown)


abderrahim (approved):

gtristan:
Thanks for the nice cleanup Ben !

PR #1868, opened 2023-09-05
Fixes Python 3.12 build for bst-1: Updating versioneer again, properly.

### LINKED ISSUE #1855: Can't build with Python 3.12

cheese:
commit cf7ff12fe0c98bd0e3b86a00c9443004a56600fc in master branch

First time around we were using `versioneer install --no-vendor`, and neglecting to remove the `versioneer.py` file from the toplevel, which is still required by BuildStream's setup.py for a few reasons.

Instead, go back to `versioneer install --vendor` and update versioneer files correctly.

Fixes #1855

### DIFF
--- a/versioneer.py
+++ b/versioneer.py
@@ -1,6 +1,5 @@
-# pylint: skip-file
 
-# Version: 0.18
+# Version: 0.28
 
 """The Versioneer - like a rocketeer, but for versions.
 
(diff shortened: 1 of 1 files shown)


nanonyme:
Is there actually demand for this still? BuildStream 1 is now a legacy tool.

nanonyme:
I'm not convinced this will be the last problem. BuildStream 1 has historically completely broken with every Python update due to its mixing of asyncio and multiprocessing in not really supported ways.

gtristan:
> Is there actually demand for this still? BuildStream 1 is now a legacy tool.

I have client organizations still using bst1, although likely to migrate to bst2.

In any case this is not a concern, our policy since releasing bst2 has been that we are happy to accept bugfix contributions and push out snapshot releases, upstream is not actively maintaining bst1.

> I'm not convinced this will be the last problem. BuildStream 1 has historically completely broken with every Python update due to its mixing of asyncio and multiprocessing in not really supported ways.

Indeed we should at least add Python 3.12 to the tests before making any claims.

PR #1877, opened 2023-10-15
_yaml.pyx: disallow duplicate keys in yaml mappings

abderrahim:
The current behaviour is to silently keep the latst one which is obviously wrong.

ruamel-yaml as used in the round-tripping code also raises an error (at least in the current version)

### DIFF
--- a/src/buildstream/_yaml.pyx
+++ b/src/buildstream/_yaml.pyx
@@ -169,6 +169,8 @@ cdef class Representer:
         return RepresenterState.wait_key
 
     cdef RepresenterState _handle_wait_key_ScalarEvent(self, object ev):
+        if ev.value in self.output[-1]:
+            raise YAMLLoadError(f"Duplicate key {ev.value} at line {ev.start_mark.line} column {ev.start_mark.column}")
         self.keys.append(ev.value)
         return RepresenterState.wait_value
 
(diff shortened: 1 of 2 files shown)


gtristan:
If I understand correctly, this will improve correctness by raising an error when bst loads redundant dict keys such as:

```yaml:
mydict:
  foo: Pony
  bar: Horsy
  foo: Zebry
```

And the update to the test case is required because we're otherwise using our yaml parsing to parse "variables" dictionaries of multiple elements which would cause this error to be raised.

If my understanding is correct, then this looks good to me @abderrahim please feel free to merge :)

As an aside, it really wouldn't hurt to have a test which has bst loading some malformed yaml variables, and asserting the provenance in the error exception as we do in other areas.

fishrockz:
Isn't this going to cause currently working but badly constructed projects to start failing to load?

Would that not make it a braking change?

Wouldn't this ether need to be a configurable failure or a bump to bst3?

Ps I agree this is obviously behaviour we want to call out to our users and generally fail on

nanonyme:
This was supposed to be BuildStream 2 behaviour. BuildStream 1 warned about this and stated it's going to start failing.

nanonyme:
At minimum bring the warnings back.

abderrahim:
> This was supposed to be BuildStream 2 behaviour. BuildStream 1 warned about this and stated it's going to start failing.

Except it wasn't :). Using buildstream 1, you could get either no warning, a warning or an error depending on the version of ruaml-yaml that you're using.

This is why I think it is reasonable to merge this even though it used to "work": depending on the version of a dependency it could work or not.

PR #1879, opened 2023-11-01
Allow integrating to new LocalCAS keepalive time

nanonyme:
This is opt-in by user. Integration for https://gitlab.com/BuildGrid/buildbox/buildbox-casd/-/issues/89

### DIFF
--- a/src/buildstream/_cas/casremote.py
+++ b/src/buildstream/_cas/casremote.py
@@ -13,6 +13,7 @@
 #
 
 import grpc
+from google.protobuf.duration_pb2 import Duration  # pylint: disable=no-name-in-module
 
 from .._protos.google.rpc import code_pb2
 from .._protos.build.buildgrid import local_cas_pb2
@@ -69,6 +70,8 @@ class CASRemote(BaseRemote):
             cas_endpoint.client_key = self.spec.client_key
         if self.spec.client_cert:
             cas_endpoint.client_cert = self.spec.client_cert
+        if self.spec.keepalive_time is not None:
+            cas_endpoint.keepalive_time = Duration().FromSeconds(self.spec.keepalive_time)
         try:
             response = local_cas.GetInstanceNameForRemotes(request)
         except grpc.RpcError as e:
--- a/src/buildstream/_remotespec.py
+++ b/src/buildstream/_remotespec.py
@@ -72,6 +72,7 @@ class RemoteSpec:
         client_key: str = None,
         client_cert: str = None,
         instance_name: Optional[str] = None,
+        connection_config: Optional[MappingNode] = None,
         spec_node: Optional[MappingNode] = None,
     ) -> None:
 
@@ -112,6 +113,9 @@ class RemoteSpec:
         # The grpc credentials object
         self._credentials: Optional[ChannelCredentials] = None
 
+        # Various connection parameters for grpc connection
+        self._connection_config: Optional[MappingNode] = connection_config
+
     #
     # Implement dunder methods to support hashing and
     # comparisons.
@@ -129,6 +133,7 @@ class RemoteSpec:
                 self.server_cert_file,
                 self.client_key_file,
                 self.client_cert_file,
+                self.keepalive_time,
             )
         )
 
@@ -175,6 +180,14 @@ class RemoteSpec:
                 certificate_chain=self.client_cert,
             )
         return self._credentials
+
+    # grpc keepalive time
+    #
+    @property
+    def keepalive_time(self) -> Optional[int]:
+        if self._connection_config:
+            return self._connection_config.get_int("keepalive-time", None)
+        return None
 
     # open_channel()
     #
@@ -230,7 +243,7 @@ class RemoteSpec:
         push: bool = False
         remote_type: str = RemoteType.ENDPOINT
 
-        valid_keys: List[str] = ["url", "instance-name", "auth"]
+        valid_keys: List[str] = ["url", "instance-name", "auth", "connection-config"]
         if not remote_execution:
             remote_type = cast(str, spec_node.get_enum("type", RemoteType, default=RemoteType.ALL))
             push = spec_node.get_bool("push", default=False)
@@ -253,6 +266,8 @@ class RemoteSpec:
         if auth_node:
             server_cert, client_key, client_cert = cls._parse_auth(auth_node, basedir)
 
+        connection_config = spec_node.get_mapping("connection-config", None)
+
         return cls(
             remote_type,
             url,
@@ -261,6 +276,7 @@ class RemoteSpec:
             client_key=client_key,
             client_cert=client_cert,
             instance_name=instance_name,
+            connection_config=connection_config,
             spec_node=spec_node,
         )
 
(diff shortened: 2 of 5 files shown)


src/buildstream/_remotespec.py
```diff
@@ -253,6 +266,10 @@ def new_from_node(
         if auth_node:
             server_cert, client_key, client_cert = cls._parse_auth(auth_node, basedir)
 
+        keepalive_config = spec_node.get_mapping("keepalive-config", None)
+        if keepalive_config is not None:
+            keepalive_config.validate_keys(["keepalive-time"])
```
abderrahim: I think this is too much keepalive. We have only one keepalive setting, so no need to put it in a keepalive-config dictionary. We should either have it directly in the server config or (and I think this would be better) in a generic dictionary for connection settings that allows adding things later.

src/buildstream/_protos/build/bazel/remote/asset/v1/remote_asset_pb2.py
```diff
@@ -4,9 +4,8 @@
 """Generated protocol buffer code."""
 from google.protobuf import descriptor as _descriptor
 from google.protobuf import descriptor_pool as _descriptor_pool
-from google.protobuf import message as _message
```
abderrahim: I think we were using a specific version of protobuf compiler to generate these so that they work with both old and new grpc.

src/buildstream/_remotespec.py
nanonyme: This was suggestion by @juergbi if I understood right. There are about ten keepalive settings.

src/buildstream/_protos/build/bazel/remote/asset/v1/remote_asset_pb2.py
nanonyme: Is it documented which version this was?

src/buildstream/_remotespec.py
nanonyme: This is anyway reasonably localized thing to change. The reasoning here was that full grpc keepalive spec looks like this https://grpc.io/docs/guides/keepalive/#keepalive-configuration-specification

src/buildstream/_protos/build/bazel/remote/asset/v1/remote_asset_pb2.py
nanonyme: Looks like it was 3.20.1. This is btw *really* dangerous, the grpcio-tools that is compatible with this old protobuf is getting increasingly hard to install. It would be more maintainable to drop support for older grpc.

nanonyme: Yay. grpcio-tools needed doesn't build anymore with modern g++.

src/buildstream/_remotespec.py
nanonyme: I converted this to more generic connection-config.

src/buildstream/_remotespec.py
```diff
@@ -112,6 +113,9 @@ def __init__(
         # The grpc credentials object
         self._credentials: Optional[ChannelCredentials] = None
 
+        # Various connection parameters for grpc connection
+        self._connection_config: Optional[int] = connection_config
```
juergbi:
```suggestion
        self._connection_config: Optional[MappingNode] = connection_config
```

src/buildstream/_remotespec.py
```diff
@@ -176,6 +181,13 @@ def credentials(self) -> ChannelCredentials:
             )
         return self._credentials
 
+    # grpc keepalive timeout
+    #
+    @property
+    def keepalive_time(self) -> Optional[int]:
+        if self._connection_config:
+            return self._connection_config.get_int("keepalive-timeout", None)
```
juergbi:
```suggestion
            return self._connection_config.get_int("keepalive-time", None)
```

juergbi (commented):
The documentation also needs to be updated for this. And isn't the addition of this dict a compatibility issue for projects? I.e., do we need to release this as BuildStream 2.2? And if we want flexibility with connection configuration in the future, we may need to be less strict with validation of keys in `connection-config`.

src/buildstream/_remotespec.py
nanonyme: You are right, if this is set in project config, then it clearly needs a version bump. I was assuming (perhaps incorrectly) that the most useful way to use this is through user config where you override project CAS with your own config and set timeout there if you're having connection issues.

src/buildstream/_remotespec.py
nanonyme: Should this say keepalive time or timeout? I know the problem I currently have is being inconsistent. https://grpc.io/docs/guides/keepalive/ talks of keepalive timeout.

nanonyme:
Is there in addition to https://docs.buildstream.build/master/using_config.html#cache-control another place where this should be mentioned?

src/buildstream/_remotespec.py
abderrahim: If you look carefully, it talks about both KEEPALIVE_TIME and KEEPALIVE_TIMEOUT (The time between pings vs how long to wait for a response to a ping). 

This is about the keepalive time.

src/buildstream/_remotespec.py
nanonyme: I fixed this wrong, now it's correct.

src/buildstream/_remotespec.py
```diff
@@ -227,10 +240,11 @@ def new_from_node(
         server_cert: Optional[str] = None
         client_key: Optional[str] = None
         client_cert: Optional[str] = None
+        keepalive_config: Optional[MappingNode] = None
```
abderrahim: linter is complaining this is unused.

src/buildstream/_cas/casremote.py
```diff
@@ -13,6 +13,7 @@
 #
 
 import grpc
+from google.protobuf.duration_pb2 import Duration
```
abderrahim: I don't know if it's fatal, but linter is complaining about this. Maybe we need to add some dependency?

src/buildstream/_remotespec.py
nanonyme: Fixed

src/buildstream/_cas/casremote.py
nanonyme: I tried to suppress this. Remains to be seen if it was successful.

nanonyme: Now it should be correct suppression no-name-in-module

nanonyme:
[00:02:11] FAILURE [ffeab612] hello.bst: tar source at hello.bst [line 16 column 2]: Error mirroring http://ftpmirror.gnu.org/gnu/automake/automake-1.16.tar.gz: <urlopen error [Errno 99] Cannot assign requested address>

jjardon:
@juergbi @abderrahim @gtristan would it be possible to get this merged please?

abderrahim:
Landing now that we have a bst-2.1 branch.

nanonyme:
Opening up on what @juergbi said above:  it was decided this goes out in 2.2.x since it allows projects to require 2.2 as minimum version of BuildStream (and take this feature into use).

PR #1880, opened 2023-11-19
tests: pytest_collection_modifyitems hook to skip tests

abderrahim:
It was previously using pytest_runtest_setup to skip tests, but this only ran for tests that were in the tests/ directory, and not for the sourcetests.

Instead use pytest_collection_modifyitems which runs earlier and has access to all collected tests. Since this runs earlier than the tests, we add a skip marker to those tests we want to skip.

Comparing the tests that ran before and after this change, we see
* Before: `1 failed, 1258 passed, 200 skipped, 85 warnings in 116.92s (0:01:56)`
* After: `1258 passed, 201 skipped, 85 warnings in 119.58s (0:01:59)`

So `source_determinism.py::test_deterministic_source_umask[tar]` is indeed the only test affected by this.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -47,36 +47,37 @@ def pytest_addoption(parser):
     parser.addoption("--remote-cache", action="store_true", default=False, help="Run remote-cache tests only")
 
 
-def pytest_runtest_setup(item):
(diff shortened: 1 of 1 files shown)


gtristan (approved):
I'm fine with this, if the tests on this PR can pass :)

PR #1872, opened 2023-09-16
Add support for Python 3.12

jjardon:
Following the steps documented here:
doc/source/hacking/updating_python_deps.rst

PR #1881, opened 2023-12-01
Add support for Python 3.12

nanonyme:
Continuation of https://github.com/apache/buildstream/pull/1872

### DIFF
--- a/.github/common.env
+++ b/.github/common.env
@@ -1,6 +1,6 @@
 # Shared common variables
 
-CI_IMAGE_VERSION=master-862722333
-CI_TOXENV_MAIN=py37,py38,py39,py310,py311
-CI_TOXENV_PLUGINS=py37-plugins,py38-plugins,py39-plugins,py310-plugins,py311-plugins
+CI_IMAGE_VERSION=master-1091773482
+CI_TOXENV_MAIN=py37,py38,py39,py310,py311,py322
+CI_TOXENV_PLUGINS=py37-plugins,py38-plugins,py39-plugins,py310-plugins,py311-plugins,py312-plugins
 CI_TOXENV_ALL="${CI_TOXENV_MAIN},${CI_TOXENV_PLUGINS}"
--- a/.github/compose/ci.docker-compose.yml
+++ b/.github/compose/ci.docker-compose.yml
@@ -114,3 +114,8 @@ services:
     <<: *tests-template
     image: quay.io/pypa/manylinux_2_28_x86_64
     command: .github/wheel-helpers/test-wheel-manylinux.sh cp311 /opt/python/cp311-cp311/bin/python3
+
+  wheels-manylinux_2_28-cp312:
+    <<: *tests-template
+    image: quay.io/pypa/manylinux_2_28_x86_64
+    command: .github/wheel-helpers/test-wheel-manylinux.sh cp312 /opt/python/cp312-cp312/bin/python3
--- a/.github/compose/ci.docker-compose.yml
+++ b/.github/compose/ci.docker-compose.yml
@@ -30,6 +30,10 @@ services:
     <<: *tests-template
     image: registry.gitlab.com/buildstream/buildstream-docker-images/testsuite-fedora:38-${CI_IMAGE_VERSION:-latest}
 
+  fedora-39:
+    <<: *tests-template
+    image: registry.gitlab.com/buildstream/buildstream-docker-images/testsuite-fedora:39-${CI_IMAGE_VERSION:-latest}
+
   debian-10:
     <<: *tests-template
     image: registry.gitlab.com/buildstream/buildstream-docker-images/testsuite-debian:10-${CI_IMAGE_VERSION:-latest}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -126,7 +126,7 @@ jobs:
         run: ${GITHUB_WORKSPACE}/.github/wheel-helpers/fetch-latest-buildbox-release.sh
 
       - name: Build wheels
-        run: pipx run cibuildwheel==2.11.3
+        run: pipx run cibuildwheel==v2.16.2
 
       - uses: actions/upload-artifact@v3
         with:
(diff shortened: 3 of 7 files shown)


.github/common.env
```diff
@@ -1,6 +1,6 @@
 # Shared common variables
 
-CI_IMAGE_VERSION=master-862722333
+CI_IMAGE_VERSION=master-1091773482
 CI_TOXENV_MAIN=py37,py38,py39,py310,py311
 CI_TOXENV_PLUGINS=py37-plugins,py38-plugins,py39-plugins,py310-plugins,py311-plugins
```
abderrahim: I think you need to add py312 to the above two lines to make it run in CI (I wonder why we have this)

nanonyme: I assume this is so we can test on same Fedora version with multiple Python versions. But if so, no idea why so many Fedora versions are used....

nanonyme: Fixed.

abderrahim:
So I tried to run the tests locally, and it seems we can't easily get 3.7 and 3.12 at the same time. 3.12 doesn't work with the current dependencies and trying to update them seems to break 3.7.

juergbi:
> So I tried to run the tests locally, and it seems we can't easily get 3.7 and 3.12 at the same time. 3.12 doesn't work with the current dependencies and trying to update them seems to break 3.7.

I suppose that's not surprising given that Python 3.7 reached end of life 5 months ago. I think it's time to drop Python 3.7 support in BuildStream master as well.

nanonyme:
@juergbi who needs to be present for that decision?

nanonyme:
@abderrahim both pyroaring and ruamel.yaml.clib *claim* to support Python 3.7 with the new versions that support Python 3.12.  Was the problematic dependency grpcio?

nanonyme:
@juergbi can I have another pipeline? It would be useful to be able to run pipelines without approval every time.

nanonyme:
Looks like grpcio 1.59.0 or newer is required for Python 3.12 support. These should support Python 3.12.

nanonyme:
@abderrahim also how did you handle pluginbase? It looks like there simply is no  released version of it that supports Python 3.12.  There was a deprecation warning about pending removal of asyncore, I guess it's now removed.

nanonyme:
I am seeing if I can surgically create dependencies that work on all Python versions. It would be better to drop Python 3.7 support in a separate PR if possible.

nanonyme:
@juergbi there's some random errors with buildbox-casd. It's problematic that tests do not propagate buildbox-casd logs when they fail.

abderrahim:
> @juergbi there's some random errors with buildbox-casd. It's problematic that tests do not propagate buildbox-casd logs when they fail.

If this is the buildgrid service-tests, you can see that buildgrid is crashing because it's failing to import `redis`

nanonyme:
@abderrahim no, I mean Debian 10, Fedora 37 and Fedora 38. There's comments about buildbox-casd not starting as expected. And we're also seeing the pesky hardlink failures.

```
    [00:00:00] FAILURE [4a47c98a] build-shell/buildtree.bst: Failed to link '/home/testuser/buildstream/.tox/py311/tmp/test_shell_pull_cached_buildtr2/cache/artifacts/refs/test/build-shell-buildtree/4a47c98a10df39e65e99d471f96edc5b58d4ea5b9b1f221d0be832a8124b8099 -> /home/testuser/buildstream/.tox/py311/tmp/test_shell_pull_cached_buildtr2/cache/artifacts/refs/test/build-shell-buildtree/38b730aec720dbddba843c9e3ce1ad8073e39d16cc4d25f752bc89d867fbaec1': [Errno 17] File exists: '/home/testuser/buildstream/.tox/py311/tmp/test_shell_pull_cached_buildtr2/cache/artifacts/refs/test/build-shell-buildtree/4a47c98a10df39e65e99d471f96edc5b58d4ea5b9b1f221d0be832a8124b8099' -> '/home/testuser/buildstream/.tox/py311/tmp/test_shell_pull_cached_buildtr2/cache/artifacts/refs/test/build-shell-buildtree/38b730aec720dbddba843c9e3ce1ad8073e39d16cc4d25f752bc89d867fbaec1'
````

**Comment by nanonyme** [comment]
Beyond these it seems tests now pass.

**Comment by nanonyme** [comment]
Can I have another pipeline? I pushed commit that may stabilize tests.

File: src/buildstream/utils.py [hunk 2]
```diff
@@ -367,6 +367,15 @@ def safe_copy(src: str, dest: str, *, copystat: bool = True, result: Optional[Fi
                 result.failed_attributes.append(dest)
 
 
+def _same_file(src: str, dest: str):
+    src_stat = os.stat(src)
+    dest_stat = os.stat(dest)
+    if not src_stat.st_ino or not dest_stat:
```
nanonyme: There is a bug here so PR must not be merged but I'm letting current pipeline finish.

nanonyme:
The commit did not help.
`tests/integration/shellbuildtrees.py::test_shell_pull_cached_buildtree[pull-without-buildtree] FAILED [ 72%]` this is still esoterically failing due to file existing that should not.

nanonyme:
It looks like it's always the same test that fails.

abderrahim:
> It looks like it's always the same test that fails.

Yeah. It looks like some race condition or something. We need to figure out where it is coming from :/

abderrahim:
> > It looks like it's always the same test that fails.
> 
> Yeah. It looks like some race condition or something. We need to figure out where it is coming from :/

Indeed it is. For some reason, bst is launching 2 jobs at once to pull the same element.

abderrahim:
@nanonyme https://github.com/apache/buildstream/pull/1882 fixes the "pesky hardlink failure", can you please rebase now that it's merged.

nanonyme:
@abderrahim pushed, removed the last commit that was irrelevant.

nanonyme:
Only buildgrid test is now failing.

abderrahim:
Yup, and I believe it's the buildgrid image that's broken (and our tests broke because we're using the nightly image of buildgrid rather than a fixed version)

abderrahim (approved):

PR #1882, opened 2023-12-05
Fix pulling buildtrees in bst shell

abderrahim:
We use _PipelineSelection.NONE when using the buildtrees, which includes the element. But then, we also add the element once more, and end up with two jobs running at once and (sometimes) failing to link the artifacts into place

This fixes test_shell_pull_cached_buildtree[pull-without-buildtree] failing randomly

### DIFF
--- a/src/buildstream/_stream.py
+++ b/src/buildstream/_stream.py
@@ -308,7 +308,7 @@ class Stream:
             element = self.targets[0]
             element._set_required(scope)
 
-            if scope == _Scope.BUILD:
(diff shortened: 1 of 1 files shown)


gtristan (approved):
Awesome ! Nice catch :)

abderrahim:
I see the buildgrid test is failing, but that's likely because of a previous commit that was merged thinking only test_shell_pull_cached_buildtree was failing. Will fix in a follow-up

PR #1883, opened 2023-12-05
Cleanup testing for Fedora versions

nanonyme:
* Add Fedora 39 to one place where it was missing
* Remove Fedora 37 completely because it doesn't actually add to Python versions being tested and is EOL since today

### DIFF
--- a/.github/compose/ci.docker-compose.yml
+++ b/.github/compose/ci.docker-compose.yml
@@ -22,10 +22,6 @@ x-tests-template: &tests-template
 
 services:
 
-  fedora-37:
-    <<: *tests-template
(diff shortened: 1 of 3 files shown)


BenjaminSchubert (approved):
Seems good!

nanonyme:
There's also something wrong with the Fedora 39 build but I will create separate PR for that. I think it is not running all of the tests, that's why it is so much faster.

nanonyme:
It is deeply suspicious why Fedora 39 is always so much faster. Is it not running all the tests?

abderrahim:
> It is deeply suspicious why Fedora 39 is always so much faster. Is it not running all the tests?

Indeed, it's only running the plugin tests, the main tests are looking for python 3.22 apparently

abderrahim:
https://github.com/apache/buildstream/pull/1884

PR #1884, opened 2023-12-08
Run CI tests for python 3.12 instead of the nonexistent 3.22

abderrahim:


juergbi (approved):

nanonyme:
Heh, now it's slowest. Well, at least we're not losing coverage.

PR #1886, opened 2023-12-15
tox.ini: use allowlist_externals instead of whitelist_externals

abderrahim:
whitelist_externals is deprecated since tox 3.18 and removed in tox 4.x


nanonyme:
@abderrahim where does that buildgrid failure need to be fixed? It's kind of annoying that this CI results are showing bogus failures.

nanonyme:
Is that buildgrid age test again some flaky test?

abderrahim:
Yeah, it failed again when I rebased. Retrying it, it seems to work fine.

PR #1888, opened 2024-01-05
Add requirement for ruamel.yaml

nanonyme:
ruamel.yaml 0.18.x has incompatible API chnges


nanonyme:
The alternative is to fix BuildStream 1 to support new ruamel-yaml API but this is presumably a larger change.

gtristan:
This fails to update the requirements after changing the `.in` file (`make -C requirements/`)

Replacing this with #1889 which updates the requirements and makes some modifications to pass the new pylint.

PR #1895, opened 2024-02-27
DownloadableFileSource: Support "auth-header-format" extra data

gtristan:
Using the new `extra_data` parameter reported from SourceMirror objects (See #1890), support the `"auth-header-format"` extra data and use this to construct and add addition auth headers.

E.g.:

```
Authorization: Bearer 1234
```

### DIFF
--- a/src/buildstream/downloadablefilesource.py
+++ b/src/buildstream/downloadablefilesource.py
@@ -25,6 +25,66 @@ Any derived classes must write their own stage() and get_unique_key()
 implementation.
 
 
+SourceMirror extra data "auth-header-format"
+--------------------------------------------
+The DownloadableFileSource, and consequently any :class:`Source <buildstream.source.Source>`
+implementations which derive from DownloadableFileSource, support the "auth-header-format"
+extra data returned by :class:`SourceMirror <buildstream.sourcemirror.SourceMirror>` plugins
+through :func:`Source.translate_url() <buildstream.source.Source.translate_url>`.
+
+This functionality is available **Since: 2.2**.
+
+This allows one to use :class:`SourceMirror <buildstream.sourcemirror.SourceMirror>` plugins
+to add an authorization header to the ``GET`` requests.
+
+
+**Example:**
+
+.. code:: python
+
+   class MySourceMirror(SourceMirror):
+
+        def translate_url(
+            self,
+            *,
+            project_name: str,
+            alias: str,
+            alias_url: str,
+            alias_substitute_url: Optional[str],
+            source_url: str,
+            extra_data: Optional[Dict[str, Any]],
+        ) -> str:
+
+            #
+            # Set the "auth-header-format" extra data
+            #
+            if extra_data is not None:
+                extra_data["auth-header-format"] = "Bearer {password}"
+
+            return alias_substitute_url + source_url
+
+The "auth-header-format" value **must** contain the ``{password}`` expression, which
+will be substituted with the corresponding password found in the user's ``~/.netrc``.
+
+
+**Example:**
+
+If the URL reported by :func:`SourceMirror.translate_url() <buildstream.sourcemirror.SourceMirror.translate_url>`
+is ``http://flying-ponies.com/downloads/pony.tgz``, then a corresponding entry will be expected in the
+user's ``~/.netrc``:
+
+.. code::
+
+   flying-ponies.com
+       password 1234
+
+Assuming the ``"auth-header-format"`` value of ``Bearer {password}`` and the configured password ``1234``,
+the DownloadableFileSource will add the following header to the ``GET`` request to download the file:
+
+.. code::
+
+   Authorization: Bearer 1234
+
 """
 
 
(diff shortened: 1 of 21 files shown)


src/buildstream/downloadablefilesource.py
```diff
@@ -145,7 +221,22 @@ class DownloadableFileSource(Source):
     def configure(self, node):
         self.original_url = node.get_str("url")
         self.ref = node.get_str("ref", None)
-        self.url = self.translate_url(self.original_url)
+
+        extra_data = {}
+        self.url = self.translate_url(self.original_url, extra_data=extra_data)
+        self.auth_header_format = extra_data.get("auth-header-format")
```
juergbi, line 227: Fail if `extra_deps` contains unknown keys?

src/buildstream/downloadablefilesource.py
```diff
@@ -145,7 +221,22 @@ class DownloadableFileSource(Source):
...
+        #
+        # Validate the auth header format for a `{password}` formatting identifier
+        #
+        if self.auth_header_format:
+            try:
+                self.auth_header_format.format(password="dummy")
+            except KeyError as e:
+                raise SourceError(
+                    "SourceMirror specified auth-header-format without a password", detail=self.auth_header_format
+                ) from e
```
juergbi, line 238: As I understand it, this raises an error if `auth_header_format` contains non-`password` replacement fields but it doesn't raise an error if there is no `{password}` field. At least the error message seems wrong but it would be better if we could actually check that issue as well.

src/buildstream/downloadablefilesource.py
abderrahim: I'm afraid that doing so will require mirrors to know what source plugin they are translating for to only provide relevant keys.

src/buildstream/downloadablefilesource.py
```diff
@@ -25,6 +25,66 @@
...
+
+This allows one to use :class:`SourceMirror <buildstream.sourcemirror.SourceMirror>` plugins
+to add an authorization header to the ``GET`` requests.
+
+
+**Example:**
+
+.. code:: python
+
+   class MySourceMirror(SourceMirror):
```
abderrahim, line 45: I don't like that this feature is only available when implementing a custom SourceMirror. I would like this to be available with regular aliases and default mirrors.

I think we could have alias definitions be "string or dict", where the dict would contain "url" and "extra-data". (Of course custom source mirrors would still support this)

src/buildstream/downloadablefilesource.py
```diff
@@ -25,6 +25,66 @@
...
+
+            #
+            # Set the "auth-header-format" extra data
+            #
+            if extra_data is not None:
+                extra_data["auth-header-format"] = "Bearer {password}"
+
+            return alias_substitute_url + source_url
+
+The "auth-header-format" value **must** contain the ``{password}`` expression, which
```
abderrahim, line 66: We could support `{login}` here as well.

nanonyme:
So where exactly is the intent that this secret should be passed? Somewhere outside BuildStream YAML files?

src/buildstream/downloadablefilesource.py
gtristan: Ah good catch, I presumed this was the other way around when trying this in the shell:

```
>>> "foo {pizza}".format(password="cheese")
KeyError
```
But as you point out indeed:
```
"foo".format(password="monkey")
'foo'
```

src/buildstream/downloadablefilesource.py
gtristan: I'm generally in agreement here, the `extra_data` is much like `public-data` and provides some _"extra context"_ which Source plugins _"might observe"_.

Otherwise, as @abderrahim points out, we will need SourceMirror instances to be very precisely targeting sources very specifically.

src/buildstream/downloadablefilesource.py
gtristan: I'm in favor of:
* Extending this such that we can allow overriding the _default SourceMirror_ definition
  * So it can be used in regular URL translation, even for the default aliaes
* Possibly changing the `extra_dict` to be public data
  * As I already pointed out elsewhere, to you @abderrahim and to @juergbi, using a `public_data` like approach was my initial idea
  * This however means further extending the APIs such that `Source.fetch()` and `Source.track()` have access to the `SourceMirror` object, which is not nicely doable in the API either (we cannot _add keyword arguments_ to these APIs without breaking API for existing implementations, so we'd need additional ugly API growth like `Source.track_with_mirror()` and similar warts and blemishes...).
  * I **think** at this point, I'm more happy with having the user provide a plugin with a single `translate_url()` function implementation rather than churning the API that much
    * Please remember: _"I think" != "I'm sure"_

gtristan:
@nanonyme 

> So where exactly is the intent that this secret should be passed? Somewhere outside BuildStream YAML files?

The way the patch currently works:
* The `SourceMirror` implementation is allowed to pass extra data through a return parameter via `Source.translate_url()` which the `Source` can see
* The `DownloadableFileSource` advertizes it's support for the `auth-header-format` extra data, which it will use to construct an authorization header in it's download `GET` request
  * As documented by the `DownloadableFileSource`, the source of the password itself will be retrieved from the `~/.netrc`

src/buildstream/downloadablefilesource.py
gtristan: That is a good point, but probably goes unused.

We can't really use this technique to do basic auth (which we already support by default), as that entails the base64 encoding too.

This brings me back to wondering about whether we should instead simply be handling _"Bearer authentication"_ more specifically, which I raised in the email: https://lists.apache.org/thread/hpwv79xy6ss9kr3op4vo2vcym7khhqlt (although I'm reassured by @juergbi's reply that we probably don't need a complex "URL Opener" implementation here).

juergbi:
I've pushed the https://github.com/apache/buildstream/tree/juerg/downloadable-auth branch based on this PR with a commit to switch to the simpler "http-auth" extra data format implemented by @abderrahim in https://github.com/apache/buildstream-plugins/pull/63 and https://gitlab.com/BuildStream/bst-plugins-experimental/-/merge_requests/289. Only minimally tested.

abderrahim:
We ended up going with #1911 instead.

PR #1896, opened 2024-03-05
source: disable check for URL marking at configure time

abderrahim:
This check is triggered by plugins that have URLs (or URL fragments) as part of their ref, such as cargo. These plugins currently resort to only translating part of the URL and appending the rest later. Dropping the check allows these plugins to call mark_download_url() or translate_url() a second time when tracking.
    
It also allows plugins to postpone calling them until fetch(), which is the behaviour the check was supposed to prevent. This is still considered wrong, but will no longer trigger an AssertionError.

This MR also removes the `_configure()` and `_get_configuring()` methods of `Plugin` as they are no longer needed.

### DIFF
--- a/src/buildstream/element.py
+++ b/src/buildstream/element.py
@@ -2890,7 +2890,7 @@ class Element(Plugin):
         self.__config = self.__extract_config(load_element)
         self.__variables.expand(self.__config)
 
-        self._configure(self.__config)
+        self.configure(self.__config)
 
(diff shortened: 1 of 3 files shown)


nanonyme:
Seems to me the entire marked URL's mechanism should be dropped if this is done. `self.__marked_urls` doesn't seem to be used anywhere else.

abderrahim:
superseded by #1899

PR #1898, opened 2024-03-11
Allow toplevel project to override mirrors for subprojects

abderrahim:
Previously, only the user configuration could override mirrors

### DIFF
--- a/src/buildstream/types.py
+++ b/src/buildstream/types.py
@@ -274,6 +274,10 @@ class _SourceUriPolicy(FastEnum):
     # configuration has not provided a mirror
     USER = "user"
 
+    # Use only URIs defined in the toplevel project (the project on  which
+    # BuildStream was invoked with as opposed to a junctioned subproject.
+    TOPLEVEL = "toplevel"
+
 
 # _PipelineSelection()
 #
--- a/doc/source/core_framework.rst
+++ b/doc/source/core_framework.rst
@@ -29,6 +29,7 @@ useful for working on BuildStream itself.
    buildstream.node
    buildstream.plugin
    buildstream.source
+   buildstream.sourcemirror
    buildstream.downloadablefilesource
    buildstream.element
    buildstream.buildelement
--- a/src/buildstream/__init__.py
+++ b/src/buildstream/__init__.py
@@ -32,6 +32,7 @@ if "_BST_COMPLETION" not in os.environ:
     from .node import MappingNode, Node, ProvenanceInformation, ScalarNode, SequenceNode
     from .plugin import Plugin
     from .source import Source, SourceError, SourceFetcher
+    from .sourcemirror import SourceMirror
     from .downloadablefilesource import DownloadableFileSource
     from .element import Element, ElementError, DependencyConfiguration
     from .buildelement import BuildElement
--- a/src/buildstream/_pluginfactory/__init__.py
+++ b/src/buildstream/_pluginfactory/__init__.py
@@ -18,6 +18,7 @@ from .pluginoriginpip import PluginOriginPip
 from .pluginoriginjunction import PluginOriginJunction
 from .sourcefactory import SourceFactory
 from .elementfactory import ElementFactory
+from .sourcemirrorfactory import SourceMirrorFactory
 
 
 # load_plugin_origin()
--- a/src/buildstream/_pluginfactory/elementfactory.py
+++ b/src/buildstream/_pluginfactory/elementfactory.py
@@ -14,8 +14,16 @@
 #  Authors:
 #        Tristan Van Berkom <tristan.vanberkom@codethink.co.uk>
 
+from typing import TYPE_CHECKING, Type, cast
+
 from .pluginfactory import PluginFactory
 from .pluginorigin import PluginType
+from .._loader import LoadElement
+from ..element import Element
+
+if TYPE_CHECKING:
+    from .._context import Context
+    from .._project import Project
 
 
 # A ElementFactory creates Element instances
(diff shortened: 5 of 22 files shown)


juergbi (commented):
It does seem a bit odd to me to have the `projects` section that is completely ignored if the project is not the top-level project. This violates the principle of least surprise to me as I wouldn't expect a (mirror) behavior difference between operating directly on a project and operating e.g. via a tiny wrapper project that doesn't override anything.

As aliases are generally project-specific, we can't simply apply a project's mirror config to its subprojects. However, could it make sense to support specifying an alias mapping in junctions? I.e., map aliases of the subproject to aliases of the current project (not URLs). And then the mirror configuration of the current project could be applied incl. mirror plugins. An interesting aspect of this would be that this could work recursively and avoid the behavior difference mentioned above.

A possible alternative to specifying a mapping for each alias in the subproject would be to specify an alias prefix.

src/buildstream/types.py
```diff
@@ -274,6 +274,10 @@ class _SourceUriPolicy(FastEnum):
     # configuration has not provided a mirror
     USER = "user"
 
+    # Use only URIs defined in the toplevel project (the project on  which
+    # BuildStream was invoked with as opposed to a junctioned subproject.
+    TOPLEVEL = "toplevel"
```
gtristan, line 279: I think we can avoid adding a policy here, which requires user configuration to decide which sources should be used when fetching and tracking.

Instead, I think that the project itself should be in charge of deciding if it is providing mirrors for it's subprojects, and this behavior is then used unconditionally for the `SourceUriPolicy.MIRRORS` portion.

gtristan:
> It does seem a bit odd to me to have the `projects` section that is completely ignored if the project is not the top-level project. This violates the principle of least surprise to me as I wouldn't expect a (mirror) behavior difference between operating directly on a project and operating e.g. via a tiny wrapper project that doesn't override anything.

If I understand correctly, this patch is trying to selectively apply project declared mirrors onto specific subprojects... this is something I was _hoping_ to avoid with a catch all _"override all the mirrors"_ kind of statement (although I accept that at implementation time it may not be possible).

Can we pause for a moment and just ask ourselves:
* If the SourceMirror plugin's translate_url() API has enough context
* If the toplevel project which wants to override subproject mirrors, can only do so with a provided custom SourceMirror plugin
* Then do we avoid the need of all this specific mapping ?
* If so, is it worthwhile to do it this way ? Or is it more valuable to allow using the default mirror format, without custom plugins, to override subproject mirrors too ?

If we _must_ target specific subprojects, we should do so by matching junction paths rather than project names, as multiple versions of the same project may exist in a tree/graph of loaded projects, with different aliases to resolve in those different versions.

I think that @juergbi's concern of the principal of least surprise could be addressed simply by naming the key something like `subprojects`.

> As aliases are generally project-specific, we can't simply apply a project's mirror config to its subprojects. However, could it make sense to support specifying an alias mapping in junctions? I.e., map aliases of the subproject to aliases of the current project (not URLs). And then the mirror configuration of the current project could be applied incl. mirror plugins. An interesting aspect of this would be that this could work recursively and avoid the behavior difference mentioned above.
> 
> A possible alternative to specifying a mapping for each alias in the subproject would be to specify an alias prefix.

This is an interesting alternative, as it allows overriding subproject aliases and mirrors in cases where a custom `SourceMirror` plugin is not needed.

abderrahim:
> Can we pause for a moment and just ask ourselves:
> 
> * If the SourceMirror plugin's translate_url() API has enough context
> * If the toplevel project which wants to override subproject mirrors, can only do so with a provided custom SourceMirror plugin
> * Then do we avoid the need of all this specific mapping ?
> * If so, is it worthwhile to do it this way ? Or is it more valuable to allow using the default mirror format, without custom plugins, to override subproject mirrors too ?

I think it is reasonable. By introducing source mirror plugins we give ourselves the opportunity to "break" backwards compatibility without really breaking backwards compatibility. What I mean is that we could have a `default` or `legacy` source mirror plugin that mirrors the behaviour and limitations of the previous design (and use it if `kind` is not specified), and then we can have a improved and slightly incompatible plugin that is still a core plugin that can be requested by specifying `kind`.

> If we _must_ target specific subprojects, we should do so by matching junction paths rather than project names, as multiple versions of the same project may exist in a tree/graph of loaded projects, with different aliases to resolve in those different versions.

I'm not sure about this. If we're using two different versions of the same project, it is very unlikely that an alias would change in an incompatible way. The most likely change is that an alias is added or removed, and I wouldn't want to duplicate the aliases in this case.

> > As aliases are generally project-specific, we can't simply apply a project's mirror config to its subprojects. However, could it make sense to support specifying an alias mapping in junctions? I.e., map aliases of the subproject to aliases of the current project (not URLs). And then the mirror configuration of the current project could be applied incl. mirror plugins. An interesting aspect of this would be that this could work recursively and avoid the behavior difference mentioned above.
> > A possible alternative to specifying a mapping for each alias in the subproject would be to specify an alias prefix.
> 
> This is an interesting alternative, as it allows overriding subproject aliases and mirrors in cases where a custom `SourceMirror` plugin is not needed.

This one sounds reasonable too.

src/buildstream/types.py
abderrahim: > Instead, I think that the project itself should be in charge of deciding if it is providing mirrors for it's subprojects, and this behavior is then used unconditionally for the SourceUriPolicy.MIRRORS portion.

It is in this patch. The `toplevel` vs `mirrors` difference mirrors the current `user` vs `mirrors`.

That is, when a subproject has its mirrors overridden by the toplevel project both `toplevel` and `mirrors` behave the same. However, when the toplevel project doesn't override the mirrors, `mirrors` will use the original mirrors and `toplevel` will raise an error.

Thinking abut this again, I'm not sure what would be the interaction between overriding the mirrors both in the toplevel project and in the user config.

abderrahim:
Superseded by #1910

PR #1899, opened 2024-03-14
source: add suffix parameter to translate_url

abderrahim:
Some sources need to download more than one file from a given URL. An instance of this is the cargo plugin, which needs to download multiple crates from the same crate registry. The names of the crates come from the ref, so it can't know all the URLs it needs at configure time (as they may change with tracking). Instead, it appends the path to the crate after calling Source.translate_url() with the crate registry URL. This will be problematic for source mirrors as they need to know the entire URL to translate it correctly.

This change allows passing the suffix separately so that only the first part, which may contain an alias, is marked and the entire URL is translated. This allows keeping in place the check that all URLs are marked (to check and store the aliases for the purpose of mirroring), while allowing source plugins to avoid workarounds and thus allowing the upcoming source mirrors plugins to translate the full URL.

This attempts to implement this proposal on the mailing list https://lists.apache.org/thread/c9xjlvlv1zhd479l5rqr47pgt82fzb78

### DIFF
--- a/src/buildstream/source.py
+++ b/src/buildstream/source.py
@@ -696,7 +696,9 @@ class Source(Plugin):
 
         return self.__mirror_directory
 
-    def translate_url(self, url: str, *, alias_override: Optional[str] = None, primary: bool = True) -> str:
+    def translate_url(
+        self, url: str, *, alias_override: Optional[str] = None, primary: bool = True, suffix: Optional[str] = None
+    ) -> str:
         """Translates the given url which may be specified with an alias
         into a fully qualified url.
 
@@ -704,6 +706,7 @@ class Source(Plugin):
            url: A URL, which may be using an alias
            alias_override: Optionally, an URI to override the alias with.
            primary: Whether this is the primary URL for the source.
+           suffix: an optional suffix to append to the URL (*Since: 2.2*)
 
         Returns:
            The fully qualified URL, with aliases resolved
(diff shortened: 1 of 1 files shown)


juergbi (approved):

src/buildstream/source.py
```diff
@@ -696,14 +696,17 @@ def get_mirror_directory(self) -> str:
...
+        self, url: str, *, alias_override: Optional[str] = None, primary: bool = True, suffix: Optional[str] = None
+    ) -> str:
         """Translates the given url which may be specified with an alias
         into a fully qualified url.
 
         Args:
            url: A URL, which may be using an alias
            alias_override: Optionally, an URI to override the alias with.
            primary: Whether this is the primary URL for the source.
+           suffix: an optional suffix to append to the URL (*Since: 2.2*)
```
gtristan, line 709: I would elaborate in the following `.. note::`

```diff
.. note::
   This must be called for every URL in the configuration during
   :func:`Plugin.configure() <buildstream.plugin.Plugin.configure>` if
   :func:`Source.mark_download_url() <buildstream.source.Source.mark_download_url>`
   is not called.

+   The *suffix* argument may be used to translate URLs for which only the base portion of
+   the URL was previously marked with :func:`Source.mark_download_url() <buildstream.source.Source.mark_download_url>`
+   at :func:`Plugin.configure() <buildstream.plugin.Plugin.configure>` time.

```

gtristan (approved):
I'm happy with this change, I think it's the safest approach.

We should land this either in advance of the `SourceMirror` plugin, or together.

Note that the `SourceMirror.translate_url()` API cannot be extended in the same way we can extend the `Source.translate_url()` optional kwargs.

gtristan:
Is it normal that `mypy` test is failing in this way ? I don't recall it failing, also it's running a python 3.11 interpreter and complaining that something is only supported since 3.8 :-/

abderrahim:
> Is it normal that `mypy` test is failing in this way ? I don't recall it failing, also it's running a python 3.11 interpreter and complaining that something is only supported since 3.8 :-/

It is "normal" because it isn't limiting the versions of the `types-*` dependencies. The "is only supported since 3.8" is a bug in mypy that is triggered by the updated `types-setuptools`. I'll look into fixing it separately.

PR #1900, opened 2024-03-15
tox.ini: Pin version of types-* in mypy tests

abderrahim:
Leaving them unpinned risks breakages (and is already broken right now), especially since we're using a very old version of mypy.


juergbi (approved):

gtristan (approved):
Awesome !

PR #1903, opened 2024-03-23
Add SourceMirror plugins (reworked)

abderrahim:


src/buildstream/_pluginfactory/sourcemirrorfactory.py
```diff
@@ -62,9 +62,9 @@ def create(self, context: "Context", project: "Project", node: MappingNode) -> S
         #
         kind = node.get_str("kind", None)
         if kind is None:
-            plugin_type = SourceMirror
-        else:
-            plugin_type, _ = self.lookup(context.messenger, kind, node)
+            kind = "legacy"
```
gtristan: I see what you're doing here, by making the default way of handling source mirrors a subclass of the base class instead of the base class.

I agree with this logic but I would not like to call it _"legacy"_, as that implies that it is old and unused, maybe we could call it _"default"_ or similar.

src/buildstream/sourcemirror.py
```diff
@@ -103,22 +110,35 @@ def __init__(
...
+        assert self.__aliases is not None, "Didn't set aliases during configuring time"
+        if alias in self.__aliases:
+            return [self]
+        return []
 
     ##########################################################
     #                        Public API                      #
     ##########################################################
+
+    def set_supported_aliases(self, aliases: List[str]):
```
gtristan, line 139: Nit, please use the standard docstrings for public API...

This one is missing:

```
Args:
   aliases: The list of ponies on the rainbow
```

src/buildstream/sourcemirror.py
```diff
@@ -93,6 +93,13 @@ class SourceMirror(Plugin):
     # The SourceMirror plugin type is only supported since BuildStream 2.2
     BST_MIN_VERSION = "2.2"
 
+    COMMON_CONFIG_KEYS = ["name", "kind"]
```
gtristan: I've always considered this to be an imperfection which we did exclusively for `Source` definitions, and feel regret for this inconsistency, it seems delicate to try to share the namespace between the core object and subclasses.

I would be more comfortable with:

```yaml
name: sonya
kind: pony
config:
   fur: pink
   tail: yellow
```

src/buildstream/_pluginfactory/sourcemirrorfactory.py
abderrahim: Yeah, I was considering that. I thought we might want to have `default` mirror plugin with a better usability (adding support for extra data, maybe dropping the multiple urls per alias) while keeping compatibility with the old mirrors.

If you think we can make improvements without breaking compatibility, I can change it to `default`.

src/buildstream/sourcemirror.py
abderrahim: The main reason I did this is to be able to implement the "legacy" plugin as a regular plugin (so that it can have `aliases` in the main dictionary). I have since added some special casing to it, but I'm still not sure how this could be done.

src/buildstream/_project.py
```diff
@@ -382,12 +402,18 @@ def alias_exists(self, alias, *, first_pass=False):
...
-    #    alias (str): The alias.
-    #    first_pass (bool): Whether to use first pass configuration (for junctions)
-    #    tracking (bool): Whether we want the aliases for tracking (otherwise assume fetching)
+    #    alias: The alias.
+    #    first_pass: Whether to use first pass configuration (for junctions)
+    #    tracking: Whether we want the aliases for tracking (otherwise assume fetching)
+    #
+    # Returns:
+    #    A list of (SourceMirror, string) tuples with each alias substitution found along
+    #    with the SourceMirror object which produced it.
```
juergbi: As far as I can tell, this returns a list with elements that can be strings (alias substitution for the legacy/default implementation) or mirror objects, not a list of tuples. These lists/values of imprecise type are passed through various functions that expect a string (according to the parameter type). Is this something you're still working on or am I misreading the code?

src/buildstream/sourcemirror.py
```diff
@@ -0,0 +1,166 @@
...
+
+    def translate_url(
+        self,
+        *,
+        project_name: str,
+        alias: str,
+        alias_url: str,
+        alias_substitute_url: Optional[str],
+        source_url: str,
+        extra_data: Optional[Dict[str, Any]],
```
juergbi: You no longer intend to change this method signature as per the discussion on the mailing list?

src/buildstream/_project.py
abderrahim: Indeed. I changed the code and forgot to change the comment.

src/buildstream/sourcemirror.py
abderrahim: Python :shrug: 

I did indeed change where it is used, but forgot it here (and noticed it when trying to address Tristan's comments).

src/buildstream/_project.py
abderrahim: > These lists/values of imprecise type are passed through various functions that expect a string (according to the parameter type)

Forgot to change the type annotations.

abderrahim: Updated.

src/buildstream/_pluginfactory/sourcemirrorfactory.py
abderrahim: Changed it to default

src/buildstream/source.py
```diff
@@ -721,31 +727,64 @@ def translate_url(
...
+
+            assert override_mirror is not None, f"{alias_override} or {self.__alias_override}"
+
+            if isinstance(override_mirror, str):
+                return override_mirror + url_body
+            #
+            # Delegate the URL translation to the SourceMirror plugin
+            #
+            return override_mirror.translate_url(
+                project_name=project.name,
```
juergbi: The `project_name` parameter has been removed from `SourceMirror.translate_url()`.

tests/frontend/project/sourcemirrors/mirror.py
```diff
@@ -0,0 +1,39 @@
...
+        aliases = node.get_mapping("aliases")
+        for alias_name, url_list in aliases.items():
+            self.aliases[alias_name] = url_list.as_str_list()
+
+        self.set_supported_aliases(self.aliases.keys())
+
+    def translate_url(
+        self,
+        *,
+        project_name: str,
```
juergbi: The `project_name` parameter has been removed from `SourceMirror.translate_url()`.

src/buildstream/plugins/sourcemirrors/default.py
```diff
@@ -0,0 +1,30 @@
+from typing import Optional, Dict, List, Any, TYPE_CHECKING
+
+from buildstream import SourceMirror, MappingNode, SequenceNode, SourceMirrorError
+
+
+class AliasesSourceMirror(SourceMirror):
+    BST_MIN_VERSION = "2.1"  # temporary
```
juergbi: Do you want to fix this up before merging or what's the plan? Is there an issue with simply creating a 2.2.0.dev0 tag now?

src/buildstream/sourcemirror.py
gtristan: I would say:

* Don't expose `COMMON_CONFIG_KEYS` as public API
* Keep `"aliases"` as a special/different configuration for the default SourceMirror type
* Don't allow other/user implemented `SourceMirror` plugins to extend configuration in the main `mirror` namespace (only under `config`)

I'd be happy if the base class simply made a special case for the `"default"` SourceMirror plugin type, passing it the `node` directly to `self._configure()` rather than passing the `config` dictionary for every other case.

src/buildstream/plugins/sourcemirrors/default.py
gtristan: That sounds sensible to me, we should probably branch 2.1 just in case in advance, and we can have as many `dev0`, `dev1` etc tags as we want in master until 2.2.

abderrahim: We already have a [bst-2.1 branch](https://github.com/apache/buildstream/tree/bst-2.1). It's just a matter of making a tag.

src/buildstream/sourcemirror.py
```diff
@@ -0,0 +1,169 @@
...
+    ):
+        # Note: the MappingNode passed here is already expanded with
+        #       the project level base variables, so there is no need
+        #       to expand them redundantly here.
+        #
+
+        # Special case for the default plugin
+        kind = node.get_str("kind", "default")
+        if kind == "default":
+            node.validate_keys(["name", "kind", "aliases"])
```
gtristan: We can drop this validation, as we do it in the "default" plugin at `Plugin.configure()` time anyway.

PR #1906, opened 2024-03-28
Support mapping subproject aliases to aliases of the parent project

abderrahim:


src/buildstream/plugins/elements/junction.py
```diff
@@ -361,6 +369,16 @@ def configure(self, node):
                 )
             self.overrides[key] = junction_name
 
+        # Map from subproject alias to local alias
+        self.aliases = node.get_mapping("aliases", default={})
+        self.map_aliases = node.get_bool("map-aliases", default=False)
```
juergbi, line 374: This isn't a bool.

src/buildstream/_project.py
```diff
@@ -786,7 +815,14 @@ def _load(self, *, parent_loader=None, provenance_node=None):
 
         # Junction configuration
         junctions_node = pre_config_node.get_mapping("junctions", default={})
-        junctions_node.validate_keys(["duplicates", "internal"])
+        junctions_node.validate_keys(["duplicates", "internal", "disallow-subproject-uris"])
+
+        if self.junction and not self.junction._get_project().disallow_subproject_uris:
```
juergbi, line 848:
```suggestion
        if self.junction and self.junction._get_project().disallow_subproject_uris:
```

src/buildstream/_project.py
```diff
@@ -220,6 +221,17 @@ def translate_url(self, url, *, first_pass=False):
...
             if alias_url:
+                if self.junction:
+                    parent_project = self.junction._get_project()
+                    parent_alias = self.junction.get_parent_alias(url_alias)
+                    if parent_alias:
+                        # Delegate translation to parent project
+                        return parent_project.translate_url(
+                            parent_alias + utils._ALIAS_SEPARATOR + url_body, first_pass=first_pass
+                        )
+                    elif parent_project.disallow_subproject_uris:
```
gtristan, line 252: This logic looks backwards, is this an artifact of having changed `allow_subproject_uris` to `disallow_subproject_uris` ?

It looks to me like we should instead have:

```python
if not disallow_subproject_uris and parent_alias:
   return parent_project.translate_url(...)
```

Rather than delegating `translate_url()` to the parent project *before* checking if that is allowed.

src/buildstream/_project.py
```diff
@@ -396,6 +416,15 @@ def get_alias_uris(self, alias, *, first_pass=False, tracking=False):
         if not alias or alias not in config._aliases:  # pylint: disable=unsupported-membership-test
             return [None]
 
+        if self.junction:
+            parent_project = self.junction._get_project()
+            parent_alias = self.junction.get_parent_alias(alias)
+            if parent_alias:
+                # Delegate translation to parent project
+                return parent_project.get_alias_uris(parent_alias, first_pass=first_pass, tracking=tracking)
+            elif parent_project.disallow_subproject_uris:
```
gtristan, line 452: Same comment, is this logic still correct after changing `allow_subproject_uris` for `disallow_subproject_uris` ?

gtristan:
In general I think this patch is probably right, I'm mostly confused by the added project.conf `junctions` configuration `disallow_subproject_uris`.

I think my previous comments can be discarded, but I'd like this patch to:
* Come with the `doc/source/format_project.rst` changes describing what exactly this new configuration does
* Ideally come with a test validating that this does what it is intended to do

src/buildstream/_project.py
juergbi: If the junction has a mapping from the subproject alias to a parent project alias, `translate_url()` should be delegated to the parent project independent of whether the parent project enforces this delegation or not.

If we want to support alias mappings in junctions only if the parent project enforces delegation for all aliases, we should instead report an error when encountering an alias mapping in a junction (as it would otherwise be unused, which would only cause confusion). However, I don't see a reason for such a restriction.

juergbi:
> In general I think this patch is probably right, I'm mostly confused by the added project.conf `junctions` configuration `disallow_subproject_uris`.
> 
> I think my previous comments can be discarded, but I'd like this patch to:
> 
>     * Come with the `doc/source/format_project.rst` changes describing what exactly this new configuration does

I've pushed a branch with initial documentation of the new flag: https://github.com/apache/buildstream/tree/juerg/junction-aliases

Edit: This and other changes are now reviewable at https://github.com/apache/buildstream/pull/1910

abderrahim:
Superseded by #1910

PR #1907, opened 2024-04-03
Fix CI failures

abderrahim:
* ci: use docker compose v2
* Remove unused 'default' values for bst init
* Update default version for bst init

(See commit messages for details)

### DIFF
--- a/src/buildstream/_frontend/app.py
+++ b/src/buildstream/_frontend/app.py
@@ -354,16 +354,16 @@ class App:
     #
     # Args:
     #    project_name (str): The project name, must be a valid symbol name
-    #    min_version (str): The minimum required version of BuildStream (default is current version)
-    #    element_path (str): The subdirectory to store elements in, default is 'elements'
+    #    min_version (str): The minimum required version of BuildStream
+    #    element_path (str): The subdirectory to store elements in
     #    force (bool): Allow overwriting an existing project.conf
     #    target_directory (str): The target directory the project should be initialized in
     #
     def init_project(
         self,
         project_name,
-        min_version=None,
-        element_path="elements",
+        min_version,
+        element_path,
         force=False,
         target_directory=None,
     ):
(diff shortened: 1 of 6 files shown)


src/buildstream/_frontend/app.py
```diff
@@ -909,13 +905,10 @@ def _assert_element_path(self, element_path):
     #    min_version (int): The user selected minimum BuildStream version
     #    element_path (str): The user selected element path
     #
-    def _init_project_interactive(self, project_name, min_version=None, element_path="elements"):
+    def _init_project_interactive(self, project_name, min_version, element_path):
 
         bst_major, bst_minor = utils._get_bst_api_version()
 
-        if min_version is None:
```
gtristan, line 916: This will make the `"default"` to `click.prompt()` be `None` at all times.

I think we should keep providing the default as the currently installed BuildStream version, as `click.prompt()` probably uses this to show the user what the default selection will be if they press ENTER without any value.

src/buildstream/_frontend/app.py
```diff
@@ -375,10 +375,6 @@ def init_project(
 
         project_path = os.path.join(directory, "project.conf")
 
-        if min_version is None:
```
gtristan, line 378: Here also, I'm not sure why we're breaking the behavior of `bst init` here so that it no longer initializes a project with the min-version set to the installed version, if `bst init` is not given the `--version` argument.

src/buildstream/_frontend/app.py
abderrahim: It doesn't. The real default is hardcoded at https://github.com/apache/buildstream/blob/master/src/buildstream/_frontend/cli.py#L411. It is never `None` here.

Another thing we can do is to drop the hardcoded default above, and initialize it here if it is `None`. The downside of that would be that `bst init --help` will no longer show the default.

gtristan: Ok I understand now and I've read through the other usages of the version numbers (checking if the selected versions are "known" by installed buildstream).

What looks ambiguous to me, is the possibility for `min_version` to be `None`, which the function allows but never happens because it is always invoked with a hardcoded value.

Let's at least:
* Update `App.init_project()` docs comment for `min_version` and `element_path`, removing the notes about what the defaults are when you pass `None` here
* Same for `App.init_project_interactive()`

Ideally we should be moving `App` to use type annotations also, but we can do that separately...

gtristan (approved):
LGTM

PR #1910, opened 2024-04-12
Support mapping subproject aliases to aliases of the parent project

juergbi:
This is an improved branch of #1906 with documentation, error handling, passing tests, implicitly fatal `unaliased-url` and an enum for the `map-alias` option.


abderrahim (approved):
Nice job getting it over the line.

PR #1911, opened 2024-04-15
DownloadableFileSource: Support "http-auth" extra data

juergbi:
This is a modification of #1895 to implement the simpler extra data format proposed by @abderrahim in https://github.com/apache/buildstream-plugins/pull/63 and https://gitlab.com/BuildStream/bst-plugins-experimental/-/merge_requests/289.

### DIFF
--- a/src/buildstream/downloadablefilesource.py
+++ b/src/buildstream/downloadablefilesource.py
@@ -25,6 +25,62 @@ Any derived classes must write their own stage() and get_unique_key()
 implementation.
 
 
+SourceMirror extra data "http-auth"
+--------------------------------------------
+The DownloadableFileSource, and consequently any :class:`Source <buildstream.source.Source>`
+implementations which derive from DownloadableFileSource, support the "http-auth"
+extra data returned by :class:`SourceMirror <buildstream.sourcemirror.SourceMirror>` plugins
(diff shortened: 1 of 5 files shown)


src/buildstream/downloadablefilesource.py
```diff
@@ -25,6 +25,64 @@
...
+
+        def translate_url(
+            self,
+            *,
+            project_name: str,
+            alias: str,
+            alias_url: str,
+            alias_substitute_url: Optional[str],
+            source_url: str,
+            extra_data: Optional[Dict[str, Any]],
```
abderrahim:
```suggestion
            alias: str,
            alias_url: str,
            source_url: str,
            extra_data: Optional[Dict[str, Any]],
```

src/buildstream/downloadablefilesource.py
```diff
@@ -25,6 +25,64 @@
...
+        ) -> str:
+
+            #
+            # Set the "http-auth" extra data
+            #
+            if extra_data is not None:
+                extra_data["http-auth"] = "bearer"
+
+            return alias_substitute_url + source_url
```
abderrahim: I think we can just say here

```suggestion
            # ...
```

`alias_substitute_url` is no longer a thing.

abderrahim (approved):

PR #1914, opened 2024-04-21
Fix bst-graph

nanonyme:
Apparently at some point build-deps and runtime-deps were CSV but they are no longer, now they are YAML lists. Fix bst-graph to properly operate on these. Alternative is invoking bst hundreds of times which is *slow*.

### DIFF
--- a/contrib/bst-graph
+++ b/contrib/bst-graph
@@ -40,7 +40,7 @@ def parse_args():
     '''
     parser = argparse.ArgumentParser(description=__doc__)
     parser.add_argument(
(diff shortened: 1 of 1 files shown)


gtristan:
This looks good to me.

As I understand it, the API was broken some time in the past, and it went unnoticed for a long time, so it can make sense to fix the contrib script rather than actually fixing the bst breakage (maybe it was not broken since the release of bst 2.0).

That said, it would really be good to have a `bst show` regression test to ensure this breakage doesn't happen again.

PR #1916, opened 2024-04-28
Properly set keepalive time

nanonyme:


abderrahim (approved):
The change looks good. But I think it would be better to have a test for this (not necessarily testing the keepalive behaviour, as I don't know how to do that) but at least exercising this code path.

nanonyme:
I agree. Where should this test live? which tests use actual CAS? Or would it be enough to trigger the bugs to have config tests?

PR #1918, opened 2024-05-13
Keep track of the parent alias when aliases are mapped

abderrahim:
This also adds tests for non-identity alias mapping and interaction between alias mapping and source mirrors.

### DIFF
--- a/src/buildstream/_project.py
+++ b/src/buildstream/_project.py
@@ -15,7 +15,7 @@
 #        Tristan Van Berkom <tristan.vanberkom@codethink.co.uk>
 #        Tiago Gomes <tiago.gomes@codethink.co.uk>
 
-from typing import TYPE_CHECKING, Optional, Dict, Union, List, Sequence
+from typing import TYPE_CHECKING, Optional, Dict, Union, List, Sequence, Tuple
 
 import os
 import urllib.parse
@@ -453,7 +453,7 @@ class Project:
     #
     def get_alias_uris(
         self, alias: str, *, first_pass: bool = False, tracking: bool = False
-    ) -> Sequence[Union[SourceMirror, str, None]]:
+    ) -> Sequence[Union[Tuple[str, Union[SourceMirror, str]], None]]:
 
         if first_pass:
             config = self.first_pass_config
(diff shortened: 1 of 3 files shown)


src/buildstream/_project.py
```diff
@@ -453,7 +453,7 @@ def alias_exists(self, alias, *, source, first_pass=False):
     #
     def get_alias_uris(
         self, alias: str, *, first_pass: bool = False, tracking: bool = False
-    ) -> Sequence[Union[SourceMirror, str, None]]:
+    ) -> Sequence[Union[Tuple[str, Union[SourceMirror, str]], None]]:
```
juergbi: Can we define a `NamedTuple` to make this more readable?

tests/frontend/mirror.py
```diff
@@ -895,9 +895,29 @@ def test_source_mirror_plugin(cli, tmpdir):
 @pytest.mark.parametrize("unaliased_sources", [True, False])
 @pytest.mark.parametrize("disallow_subproject_uris", [True, False])
 @pytest.mark.parametrize("fetch_source", ["aliases", "mirrors"])
-@pytest.mark.parametrize("alias_override", [["foo"], ["foo", "bar"], "global", "invalid"])
+@pytest.mark.parametrize(
```
gtristan: Can we have a comment around here explaining why this level of parameterized testing is needed, and what defects we are specifically protecting against ?

abderrahim: This is mainly to be able to exclude some elements from the test matrix.

Initially I wanted something like
```
@pytest.mark.parametrize("alias_override", [["foo"], ["foo", "bar"], "global"])
@pytest.mark.parametrize("alias_mapping", ["identity", "project-prefix", "invalid"])
```

but "global,project-prefix" isn't implemented yet (I'd like to have it in a future release) and "global,invalid" doesn't make sense.

As for the reason I'm doing this, the bug in question happens when having a non-identity mapping along with source mirror plugin. When I started, I thought it could be unrelated to source plugins, so wanted to test it.

If you have a better option, feel free to comment. Or I can add the above to a comment?

juergbi: Maybe calling `pytest.xfail()` for these two cases (with a comment) would be reasonable.

juergbi (approved):

tests/frontend/mirror.py
abderrahim: Added a comment.

PR #1921, opened 2024-05-23
Drop the directory argument from the sandbox

abderrahim:
It is currently unused and produces a bogus info message when using remote execution

### DIFF
--- a/src/buildstream/element.py
+++ b/src/buildstream/element.py
@@ -1408,7 +1408,7 @@ class Element(Plugin):
 
         # bst shell and bst artifact checkout require a local sandbox.
         # pylint: disable-next=contextmanager-generator-missing-cleanup
-        with self.__sandbox(None, config=self.__sandbox_config, allow_remote=False) as sandbox:
+        with self.__sandbox(config=self.__sandbox_config, allow_remote=False) as sandbox:
 
             # Configure always comes first, and we need it.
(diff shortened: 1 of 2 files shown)


src/buildstream/element.py
```diff
@@ -2810,14 +2807,11 @@ def __sandbox(self, directory, stdout=None, stderr=None, config=None, allow_remo
         else:
             output_node_properties = None
 
-        if directory is not None and allow_remote and context.remote_execution_specs:
-
-            self.info("Using a remote sandbox for artifact {} with directory '{}'".format(self.name, directory))
-
+        if allow_remote and context.remote_execution_specs:
+            self.info("Using a remote sandbox for artifact {}".format(self.name))
```
abderrahim: Only thing I'm not sure is whether we want to keep this message at all. I think it's pretty clear from the other messages that we're doing a remote build.

gtristan: I don't regularly do remote builds, is it clear ?

I would be happy if the only difference was a clear indication in the startup summary.

We can surly remove this info message.

gtristan (approved):
+1 LGTM.

src/buildstream/element.py
abderrahim: Yeah, it's very clear. You get "uploading build root" and "waiting for the remote build to complete" as subtasks of "running commands".

Issue #1913, opened 2024-04-19
Add Token Based Support for BuildStream to REAPI servers


cphang99:
Currently REAPI clients such as Bazel [support](https://github.com/bazelbuild/bazel/issues/13378#issuecomment-869301764) token based authenication to REAPI server implementations by passing this into authorisation headers that the server implementation then parses.

Adding this support to BuildStream would be very useful, which to the best of my knowledge only supports TLS certificate authentication currently.

A suggested sketch of things to add as part of this:
- BuildStream backend implementation for passing authentication tokens to REAPI servers
- Extending buildstream configuration for users to be able to pass authentication tokens.
- Documentation of the above.

juergbi:
buildbox-casd already supports bearer authentication for remote servers, reading the access token from a file. However, it currently supports this only for remote servers configured on the command line, not with the dynamic configuration used by BuildStream.

Tentative list of tasks to address this issue:

* buildbox-casd
  - Support dynamic remotes also for execution servers (requires small extension to the `LocalCAS` protocol)
  - Support access tokens (+ optional reload interval) for dynamic remotes (requires small extension to the `LocalCAS` protocol)
  - If `netrc` support is desired, also support access tokens sent as strings over the local protocol
* BuildStream
  - Support configuring access tokens as paths (+ reload interval) in the `auth` dict of remotes (as alternative to `client-key` / `client-cert`) and pass that to buildbox-casd if specified
  - If `netrc` support is desired, add a configuration option to enable that, which then sends the password string from `netrc` to buildbox-casd instead of a path
  - Use buildbox-casd as proxy for all remote connections: At least remote execution currently doesn't use buildbox-casd as proxy

juergbi:
> Support dynamic remotes also for execution servers (requires small extension to the LocalCAS protocol)

MR for this: https://gitlab.com/BuildGrid/buildbox/buildbox/-/merge_requests/605.

The BuildStream change to use buildbox-casd as proxy for remote asset connections: https://github.com/apache/buildstream/pull/1922

Support for proxying remote execution is implemented in #1926

juergbi:
> * Support access tokens (+ optional reload interval) for dynamic remotes (requires small extension to the `LocalCAS` protocol)

MR for this: https://gitlab.com/BuildGrid/buildbox/buildbox/-/merge_requests/616

juergbi:
[...]

PR #1922, opened 2024-05-31
Use buildbox-casd as remote asset proxy, if supported

juergbi:
buildbox-casd is already used as CAS proxy. Using buildbox-casd for all remote connections aims to improve robustness (e.g., consistent retry behavior) and will allow adding support for token-based authentication (#1913).

This does not yet use buildbox-casd as remote execution proxy

This depends on https://gitlab.com/BuildGrid/buildbox/buildbox/-/merge_requests/605 for the remote asset proxy to be used. However, BuildStream will fall back to direct gRPC connections if buildbox-casd doesn't support asset-only remotes.

### DIFF
--- a/src/buildstream/_cas/casremote.py
+++ b/src/buildstream/_cas/casremote.py
@@ -59,18 +59,7 @@ class CASRemote(BaseRemote):
 
         local_cas = self.cascache.get_local_cas()
         request = local_cas_pb2.GetInstanceNameForRemotesRequest()
-        cas_endpoint = request.content_addressable_storage
-        cas_endpoint.url = self.spec.url
-        if self.spec.instance_name:
-            cas_endpoint.instance_name = self.spec.instance_name
-        if self.spec.server_cert:
-            cas_endpoint.server_cert = self.spec.server_cert
-        if self.spec.client_key:
-            cas_endpoint.client_key = self.spec.client_key
-        if self.spec.client_cert:
-            cas_endpoint.client_cert = self.spec.client_cert
-        if self.spec.keepalive_time is not None:
-            cas_endpoint.keepalive_time.FromSeconds(self.spec.keepalive_time)
+        self.spec.to_localcas_remote(request.content_addressable_storage)
         try:
             response = local_cas.GetInstanceNameForRemotes(request)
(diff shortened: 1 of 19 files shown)


juergbi:
https://gitlab.com/BuildGrid/buildbox/buildbox/-/merge_requests/605 is now merged.

src/buildstream/_cas/casremote.py
```diff
@@ -59,18 +59,7 @@ def _configure_protocols(self):
...
-            cas_endpoint.instance_name = self.spec.instance_name
-        if self.spec.server_cert:
-            cas_endpoint.server_cert = self.spec.server_cert
-        if self.spec.client_key:
-            cas_endpoint.client_key = self.spec.client_key
-        if self.spec.client_cert:
-            cas_endpoint.client_cert = self.spec.client_cert
-        if self.spec.keepalive_time is not None:
-            cas_endpoint.keepalive_time.FromSeconds(self.spec.keepalive_time)
+        request.content_addressable_storage.CopyFrom(self.spec.to_localcas_remote())
```
abderrahim: I'm not sure what's the idiomatic thing to do with grpcio, but I feel it would be better to pass it like
```
self.spec.to_localcas_remote(request.content_addressable_storage)
```

Would that be possible? Is it something we'd like to do, or is the current way more idiomatic?

abderrahim (approved):

nanonyme:
Is the intent there would be some kind of version check for when to enable this or what?

juergbi:
> Is the intent there would be some kind of version check for when to enable this or what?

The branch includes a fallback path. I.e., with older buildbox-casd it will keep using direct connections to remote asset servers.

We may want to remove the fallback path and error out with older buildbox-casd versions in the future. I think we should wait with that at least until proxy support is also added for remote execution and #1913 is complete (and buildbox and buildbox-integration have been tagged with that support).

src/buildstream/_cas/casremote.py
juergbi: I'm not sure either what's more idiomatic but your suggestion sounds reasonable to me and it avoids the need for `CopyFrom()`. I've updated the branch accordingly.

PR #1925, opened 2024-06-21
Support access tokens for CAS and Remote Asset servers

juergbi:
This adds support for bearer authentication using access tokens for CAS and Remote Asset servers with buildbox-casd 1.2.8 and later as described in #1913.

Bearer authentication is not yet supported for Remote Execution as those gRPC connections are not yet proxied via buildbox-casd. I intend to fix this in a follow-up.

This is not currently covered by the test suite (no simple test server available with token support), however, I have manually verified that buildbox-casd correctly sets up bearer authentication when configured in BuildStream.


doc/source/using_config.rst
```diff
@@ -734,6 +736,10 @@ Attributes
   The client key is used to encrypt traffic when uploading traffic
   to the server.
 
+* ``access-token``
+
+  The path to a token for optional HTTP bearer authentication.
+
 Normally, only the ``server-cert`` is required to securely *download* data
 from remote cache services, while both the ``client-key`` and ``client-cert``
 is required to securely *upload* data to the server.
```
abderrahim: Maybe mention access token here to (as an alternative to securely upload data.

abderrahim (approved):

doc/source/using_config.rst
juergbi: Added, and also fixed inaccuracies in the description of certificate options.

doc/source/using_config.rst
```diff
@@ -710,33 +712,46 @@ The ``auth`` configuration block looks like this:
...
 
 * ``client-cert``
 
-  The client certificate is used by the remote server to decrypt
-  traffic being uploaded to the server.
+  The client certificate is used by the remote server to authenticate the client.
 
   The remote server will have its own copy of this certificate, but the
   client needs to send this certificate's identity to the server so that
   the server knows which certificate to use.
```
abderrahim: Thanks for fixing the descriptions. I believe this paragraph is still slightly incorrect. The server doesn't need to have a copy of the certificate if it is signed by a CA that the server trusts.

abderrahim (approved):

doc/source/using_config.rst
juergbi: True, a CA generally also makes sense for client certificates (although I would expect that to be an internal CA). I don't think these authentication details on the server side are actually relevant or helpful for client configuration, so I've removed that paragraph.

PR #1926, opened 2024-06-22
Use buildbox-casd as remote execution proxy

juergbi:
buildbox-casd is already used as CAS and RA proxy. Using buildbox-casd for all remote connections aims to improve robustness (e.g., consistent retry behavior) and enables support for token-based authentication (#1913).

This depends on BuildBox 1.2.6+ for the remote execution proxy to be used. However, BuildStream will fall back to direct gRPC connections if buildbox-casd doesn't support execution remotes.

### DIFF
--- a/.github/common.env
+++ b/.github/common.env
@@ -1,6 +1,6 @@
 # Shared common variables
 
-CI_IMAGE_VERSION=master-1091773482
+CI_IMAGE_VERSION=master-1361451963
 CI_TOXENV_MAIN=py37,py38,py39,py310,py311,py312
 CI_TOXENV_PLUGINS=py37-plugins,py38-plugins,py39-plugins,py310-plugins,py311-plugins,py312-plugins
 CI_TOXENV_ALL="${CI_TOXENV_MAIN},${CI_TOXENV_PLUGINS}"
--- a/src/buildstream/_cas/casdprocessmanager.py
+++ b/src/buildstream/_cas/casdprocessmanager.py
@@ -31,6 +31,7 @@ from .._protos.build.bazel.remote.asset.v1 import remote_asset_pb2_grpc
 from .._protos.build.bazel.remote.execution.v2 import remote_execution_pb2_grpc
 from .._protos.build.buildgrid import local_cas_pb2_grpc
 from .._protos.google.bytestream import bytestream_pb2_grpc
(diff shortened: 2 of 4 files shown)


gtristan:
How does this factor into the remote service tests ?

It looks like `ci.buildgrid.yml` is using nightly buildbox from `registry.gitlab.com/buildgrid/buildgrid.hub.docker.com/buildbox:nightly`

Does this mean that this new codepath will be exercised in the buildgrid test with BuildBox 1.2.6+ in that image ?

Should we EOL the older codepaths eventually and just have a hard dependency on BuildBox 1.2.6+ ? Maybe in 6 months ? Or should we extend the tests to ensure that we run them with older buildbox ?

juergbi:
> It looks like `ci.buildgrid.yml` is using nightly buildbox from `registry.gitlab.com/buildgrid/buildgrid.hub.docker.com/buildbox:nightly`

That's used for the worker container, which is not relevant to this PR.

buildbox-casd used by BuildStream CI on the client side is taken from https://gitlab.com/BuildStream/buildstream-docker-images/, which uses the latest release from https://gitlab.com/BuildGrid/buildbox/buildbox-integration/.

However, BuildStream CI uses a manually updated version from buildstream-docker-images, so this is actually still running with an old version of BuildBox. I'll push a commit to update the image for CI. (I did run the BuildStream BuildGrid tests locally with both older and updated BuildBox).

> Should we EOL the older codepaths eventually and just have a hard dependency on BuildBox 1.2.6+ ? Maybe in 6 months ?

I think so, yes. I'd like to require BuildBox 1.2.8+ after some grace period to also ensure [buildbox-run-bubblewrap drops all Linux capabilities](https://gitlab.com/BuildGrid/buildbox/buildbox/-/merge_requests/625) (only relevant when running bst as privileged user) and include support for access tokens.

gtristan:
> I think so, yes. I'd like to require BuildBox 1.2.8+ after some grace period to also ensure [buildbox-run-bubblewrap drops all Linux capabilities](https://gitlab.com/BuildGrid/buildbox/buildbox/-/merge_requests/625) (only relevant when running bst as privileged user) and include support for access tokens.

+1

gtristan (approved):
LGTM :)

PR #1931, opened 2024-07-06
Implement `pip` and `junction` origin loading for SourceMirror plugins

gtristan:
This adds support for loading SourceMirror plugins via both `pip` and `junction` plugin origins.

This tests essential mirror behavior when loading the plugins with `pip` or `junction` plugin origins,
and also tests the case where the mirror itself is used to define the mirror required to obtain itself
across a junction boundary, succeeding when the mirror is not used and asserting the failure mode
when mirroring is forced.

### DIFF
--- a/src/buildstream/_frontend/app.py
+++ b/src/buildstream/_frontend/app.py
@@ -282,6 +282,7 @@ class App:
                     self.context,
                     cli_options=self._main_options["option"],
                     default_mirror=self._main_options.get("default_mirror"),
+                    fetch_subprojects=self.stream.fetch_subprojects,
                 )
             except LoadError as e:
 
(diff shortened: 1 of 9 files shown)


gtristan:
Note about remaining imperfections.

We do *not* lazy load the SourceMirror plugins, which means that we load them in a loop, while loading the mirrors configuration in the project configuration, rather than delaying the load until the SourceMirror is actually *used*.

This can still result in some situations where:
* You have two SourceMirror plugins defined through separate junctions
* You need one of these SourceMirrors in order to *access* the other SourceMirror through *its* junction
* The project will not load properly *depending on the order in which the SourceMirrors are declared in `project.conf`*

This is a bit tricky to resolve, and I suspect the current functionality to satisfy the majority of use cases, but we will most likely run into this issue.

juergbi:
> The project will not load properly _depending on the order in which the SourceMirrors are declared in `project.conf`_

As long as the behavior is deterministic and this results in a clear error message, I don't think we need to let that issue block this PR. It may be worth a note in the documentation. However, I agree that we likely want to eliminate this limitation at some point.

PR #1932, opened 2024-07-12
tar.py: Always check member paths

juergbi:
The paths of tarball members are already checked to guard against extraction outside the target directory in the common case. However, if `base-dir` is set to the empty string, the checks were skipped.

This corrects the code to always check member paths. On Python 3.12+, this uses the new extraction filter support, which should also ensure consistent behavior on Python 3.14+, which will use the restrictive 'data' filter by default.


gtristan (approved):
Looks good !

PR #1933, opened 2024-07-18
Drop support for Python 3.7

juergbi:
Python 3.7 is no longer maintained upstream since June 27, 2023 and many Python packages have dropped support for it.

Updating our dependencies is required to get pylint and mypy working on Python 3.12. However, not all of the updated dependencies are compatible with Python 3.7 anymore.

This replaces the Python 3.7-based Debian 10 with the Python 3.8-based Ubuntu 20.04 in CI to keep testing with the oldest supported version of Python.

### DIFF
--- a/.github/common.env
+++ b/.github/common.env
@@ -1,6 +1,6 @@
 # Shared common variables
 
 CI_IMAGE_VERSION=master-1361451963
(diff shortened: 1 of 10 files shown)


gtristan (approved):
Awesome sauce

PR #1934, opened 2024-07-19
Update requirements

juergbi:
This includes an update to Pylint 3, which requires various changes to pass the new linter checks.

With this commit, pylint and mypy now also pass with Python 3.12. The CI image used for pylint and mypy is updated to the Python 3.12-based Fedora 39.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -11,7 +11,7 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 #
-[MASTER]
+[MAIN]
 
 # A comma-separated list of package or module names from where C extensions may
 # be loaded. Extensions are loading into the active Python interpreter and may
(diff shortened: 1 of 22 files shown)


.pylintrc
```diff
@@ -11,7 +11,7 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 #
-[MASTER]
+[MAIN]
```
gtristan, line 14: Is this a necessary change ?

gtristan (approved):
Overall looks good to me :)

.pylintrc
juergbi: The pylint 3.0 release notes state "Following a deprecation period, it's no longer possible to use MASTER or master as configuration section in setup.cfg or tox.ini". It seems that the section name in `.pylintrc` might actually be ignored, but the 3.x documentation exclusively mentions `MAIN`.

gtristan: Got it.

PR #1936, opened 2024-07-26
Fix tests after dependency updates

juergbi:
pytest 8 had breaking changes in test collection, which prevented the main test suite from being picked up in combination with our `sourcetests_collection_hook`.

This sets the pytest `testpaths` config to fix test collection and fixes a couple of issues that caused test failures.


gtristan (approved):
Looks pretty straightforward to me, +1

PR #1939, opened 2024-08-02
scriptelement: Fix time logging of integration commands

abderrahim:
sandbox.batch() needs to happen within the timed_activity for it to count the time it needs to run the command


gtristan (approved):
LGTM !

PR #1944, opened 2024-08-07
Update NEWS for 2.3.0

abderrahim:


juergbi:
I think we also want to update the default version for bst init in `src/buildstream/_frontend/cli.py`.

abderrahim:
It looks like updating it before tagging makes CI unhappy. So I guess we'll just have to merge this with a red pipeline and tag afterwards.

PR #1772, opened 2022-10-05
WIP: Expose casd socket interface in the sandbox

gtristan:
This allows REAPI aware tools running in the sandbox to access /tmp/casd.sock and speak to CAS.

PR #1945, opened 2024-08-07
Allow configuring sandbox option casd-socket

nanonyme:
This declares path in sandbox where casd socket will be mounted to. Continuation of https://github.com/apache/buildstream/pull/1772

### DIFF
--- a/src/buildstream/sandbox/_sandboxbuildboxrun.py
+++ b/src/buildstream/sandbox/_sandboxbuildboxrun.py
@@ -133,6 +133,15 @@ class SandboxBuildBoxRun(SandboxREAPI):
             else:
                 stdin = subprocess.DEVNULL
 
+            if "casd-socket" in self.__config:
+                # Expose the casd socket in the sandbox
+                buildbox_command.append(
+                    "--bind-mount={}:{".format(
+                        casd_process_manager._socket_path,
+                        self.__config["casd-socket"],
+                    )
+                )
+
             self._run_buildbox(
                 buildbox_command,
                 stdin,


juergbi:
Exposing the unfiltered buildbox-casd socket would result in exposing access to the host filesystem via the LocalCAS protocol.

An option that may be safe (to be reviewed) would be to change the instance name to a random token, generated in each BuildStream session, and then use `GetInstanceNameForNamespace()` to create a second instance that is then exposed to the sandbox.

Ideally, REAPI upstream would define a standard way how REAPI access can be exposed to a the action command. This shouldn't be buildbox-casd-specific and should also be supported with remote execution. Maybe configured by platform property.

abderrahim:
> use `GetInstanceNameForNamespace()` to create a second instance that is then exposed to the sandbox.

I'm not sure how this works: does it create a "sandboxed" instance somehow or is it just to have the LocalCas methods apply on the sandbox instead of the host?

---

But this indeed needs to put some thoughts on what exactly we want to provide to the sandboxes: just CAS? CAS and action cache? do we need remote asset too? (we need to at least restrict what can be requested from remote asset). How about remote execution? maybe not by default, but it's probably what would give the most performance benefit. How about LocalCAS? do we need to restrict that?

juergbi:
> I'm not sure how this works: does it create a "sandboxed" instance somehow or is it just to have the LocalCas methods apply on the sandbox instead of the host?

That's kind of the same thing, isn't it? Is there a particular difference between the two options that needs to be clarified? As mentioned, we should review the code to make sure it's sufficiently sandboxed, though.

> just CAS? CAS and action cache? do we need remote asset too?

Yes, we certainly need to give this some more thought.

abderrahim:
> That's kind of the same thing, isn't it? Is there a particular difference between the two options that needs to be clarified?

I meant is access to the CAS somehow sandboxed? (from your answer I guess it's not). Not that I think it should be, I just wanted clarification.

src/buildstream/sandbox/_sandboxbuildboxrun.py
```diff
@@ -133,6 +133,15 @@ def _execute_action(self, action, flags):
             else:
                 stdin = subprocess.DEVNULL
 
+            if "casd-socket" in self.__config:
```
abderrahim, line 136: This is probably a mistake: `self.__config` doesn't seem to exist here.

nanonyme: Hmm, is it the name mangling? self.__config belongs to grandparent of this class, Sandbox.

abderrahim: Yeah, anything that starts with two underscores is private to a class (and python enforces it)

gtristan:
Is it fair to say that #2014, which appears more complete, although in draft state, replaces this ?

juergbi:
Yes, #2014 replaces this PR.

PR #1948, opened 2024-08-11
pluginoriginpip: use importlib instead of pkg_resources

abderrahim:


abderrahim:
It looks like we can use https://pypi.org/project/backports.entry-points-selectable/ to fix the failure on python 3.8.

abderrahim:
Now that we dropped python 3.8 support, we don't need to worry about it.

juergbi:
The Python package tests in CI are failing with:
```
       File "/tmp/venv/lib/python3.9/site-packages/buildstream/_pluginfactory/pluginoriginpip.py", line 35, in get_plugin_paths
        from packaging.requirements import Requirement, InvalidRequirement
      File "/tmp/venv/lib/python3.9/site-packages/pluginbase.py", line 438, in plugin_import
        return self._system_import(import_name, globals, locals,
    ModuleNotFoundError: No module named 'packaging'
```

And Debian 11 (Python 3.9) is also not happy:
```
  File "/home/testuser/buildstream/src/buildstream/_pluginfactory/pluginoriginpip.py", line 75, in get_plugin_paths
    detail="{} {} is installed but {} is required".format(dist.name, dist.version, package),
AttributeError: 'PathDistribution' object has no attribute 'name'
```

juergbi:
There are still a few remaining references to `pkg_resources` in the repository. Can we replace them in the same branch or does it make sense to defer that to a follow-up?

src/buildstream/_pluginfactory/pluginoriginpip.py
```diff
@@ -53,63 +47,56 @@ def get_plugin_paths(self, kind, plugin_type):
...
             ) from e
-        except pkg_resources.VersionConflict as e:
+
+        if dist.version not in package.specifier:
             raise PluginError(
                 "{}: Version conflict encountered while loading {} plugin '{}'".format(
                     self.provenance_node.get_provenance(), plugin_type, kind
                 ),
-                detail=e.report(),
+                detail="{} {} is installed but {} is required".format(dist.name, dist.version, package),
```
nanonyme, line 80: Try doing `dist.metadata["Name"]` here. I think the data is still there, the property helper is just missing.

nanonyme: I looked up from newer Python versions and the property is just short-hand for extracting this data.

abderrahim:
CI is happy now.
* I used `importlib_metadata` for python 3.9 that doesn't have all the needed features in `importlib.metadata`.
* I moved setuptools to dev-requirements, as it's still used by tests. We can port tests as a followup.
* I added packaging to the requirements

PR #1950, opened 2024-08-30
Expand cache config options and improve defaults

juergbi:
* Add `reserved-disk-space` config option to the `cache` section
    BuildStream has been relying on buildbox-casd's default for this, which is currently set at 2 GB. This adds a config option and sets the default to 5% of the size of the filesystem as it doesn't seem reasonable to fill up the filesystem more than 95% by default.
* Add `low-watermark` config option to the `cache` section
    BuildStream currently uses a hard-coded value of 50% for the low watermark. This adds a config option but doesn't change the default. This requires [buildbox-casd 1.2.17+](https://gitlab.com/BuildGrid/buildbox/buildbox/-/merge_requests/669) to be effective. Older versions of buildbox-casd will continue to use 50% as low watermark.
* Change the default `low-watermark` of the cache from 50% to 80%
    Pruning half the cache when hitting the cache quota seems excessive as a default value, resulting in a lower cache hit rate.

### DIFF
--- a/src/buildstream/_context.py
+++ b/src/buildstream/_context.py
@@ -185,6 +185,9 @@ class Context:
         # User specified cache quota, used for display messages
         self.config_cache_quota_string: Optional[str] = None
 
+        # Reserved disk space for local cache in bytes
+        self.config_cache_reserved: Optional[int] = None
+
         # Remote cache server
         self.remote_cache_spec: Optional[RemoteSpec] = None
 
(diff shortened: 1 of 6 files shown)


src/buildstream/_context.py
```diff
@@ -185,6 +185,12 @@ def __init__(self, *, use_casd: bool = True) -> None:
         # User specified cache quota, used for display messages
         self.config_cache_quota_string: Optional[str] = None
 
+        # Reserved disk space for local cache in bytes
+        self.config_cache_reserved: Optional[int] = None
+
+        # Low watermark for local cache in bytes
```
abderrahim: this is not "in bytes"

src/buildstream/utils.py
```diff
@@ -832,6 +832,35 @@ def _parse_size(size, volume):
...
+#
+# Raises:
+#     UtilError if the string is not a valid percentage.
+#
+def _parse_percentage(percentage):
+    if not percentage.endswith("%"):
+        raise UtilError("{} is not a valid percentage.".format(percentage))
+
+    try:
+        num = int(percentage[:-1])
```
abderrahim: It seems `_parse_size()` accepts a float as a percentage value. Maybe we should here too for consistency.

abderrahim (approved):
A couple nits, but otherwise LGTM.

src/buildstream/utils.py
juergbi: Fair enough. Updated.

PR #1958, opened 2024-09-24
Fix SyntaxWarning with utils.py

nanonyme:


nanonyme:
There's in console a warning
`/usr/lib/python3.12/site-packages/buildstream/utils.py:231: SyntaxWarning: invalid escape sequence '\*`
this is because backslashes have to be escaped in docstrings as well if they aren't marked as raw strings.

juergbi (approved):

Issue #1961, opened 2024-10-09
Crash when using remote execution on a large element


abderrahim:
Trying to build a large element (and after raising limit on the number of open files to work around https://github.com/apache/buildstream/issues/1842), I got another crash

```
    An unhandled exception occured:
    
    Traceback (most recent call last):
      File "/home/abderrahimkitouni/.local/pipx/venvs/buildstream/lib/python3.11/site-packages/buildstream/_scheduler/jobs/job.py", line 346, in child_action
        result = self.child_process()  # pylint: disable=assignment-from-no-return
                 ^^^^^^^^^^^^^^^^^^^^
      File "/home/abderrahimkitouni/.local/pipx/venvs/buildstream/lib/python3.11/site-packages/buildstream/_scheduler/jobs/elementjob.py", line 81, in child_process
        return self._action_cb(self._element)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/home/abderrahimkitouni/.local/pipx/venvs/buildstream/lib/python3.11/site-packages/buildstream/_scheduler/queues/buildqueue.py", line 55, in _assemble_element
        element._assemble()
      File "/home/abderrahimkitouni/.local/pipx/venvs/buildstream/lib/python3.11/site-packages/buildstream/element.py", line 1687, in _assemble
        collect = self.assemble(sandbox)  # pylint: disable=assignment-from-no-return
                  ^^^^^^^^^^^^^^^^^^^^^^
      File "/home/abderrahimkitouni/.local/pipx/venvs/buildstream/lib/python3.11/site-packages/buildstream/buildelement.py", line 315, in assemble
        with sandbox.batch(root_read_only=True, label="Running commands"):
      File "/usr/lib/python3.11/contextlib.py", line 144, in __exit__
        next(self.gen)
      File "/home/abderrahimkitouni/.local/pipx/venvs/buildstream/lib/python3.11/site-packages/buildstream/sandbox/sandbox.py", line 265, in batch
        batch.execute()
      File "/home/abderrahimkitouni/.local/pipx/venvs/buildstream/lib/python3.11/site-packages/buildstream/sandbox/_sandboxreapi.py", line 234, in execute
        self.sandbox._run_with_flags(
      File "/home/abderrahimkitouni/.local/pipx/venvs/buildstream/lib/python3.11/site-packages/buildstream/sandbox/sandbox.py", line 374, in _run_with_flags
[...]

Issue #1842, opened 2023-05-13
"Too many open files" when running with remote execution


abderrahim:
When building with remote execution, I get the following error. I don't know if it's consistent, but retrying once gives the same error. 

```
[16:35:19][00:11:56][e8a2758a][   build:bootstrap/build/gcc-stage1.bst] BUG     Build

    An unhandled exception occured:
    
    Traceback (most recent call last):
      File "/home/abderrahimkitouni/.local/pipx/venvs/buildstream/lib/python3.11/site-packages/buildstream/_scheduler/jobs/job.py", line 438, in child_action
        result = self.child_process()  # pylint: disable=assignment-from-no-return
                 ^^^^^^^^^^^^^^^^^^^^
      File "/home/abderrahimkitouni/.local/pipx/venvs/buildstream/lib/python3.11/site-packages/buildstream/_scheduler/jobs/elementjob.py", line 91, in child_process
        return self._action_cb(self._element)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/home/abderrahimkitouni/.local/pipx/venvs/buildstream/lib/python3.11/site-packages/buildstream/_scheduler/queues/buildqueue.py", line 55, in _assemble_element
        element._assemble()
      File "/home/abderrahimkitouni/.local/pipx/venvs/buildstream/lib/python3.11/site-packages/buildstream/element.py", line 1689, in _assemble
        collect = self.assemble(sandbox)  # pylint: disable=assignment-from-no-return
                  ^^^^^^^^^^^^^^^^^^^^^^
      File "/home/abderrahimkitouni/.local/pipx/venvs/buildstream/lib/python3.11/site-packages/buildstream/buildelement.py", line 315, in assemble
        with sandbox.batch(root_read_only=True, label="Running commands"):
      File "/usr/lib/python3.11/contextlib.py", line 144, in __exit__
        next(self.gen)
      File "/home/abderrahimkitouni/.local/pipx/venvs/buildstream/lib/python3.11/site-packages/buildstream/sandbox/sandbox.py", line 265, in batch
        batch.execute()
      File "/home/abderrahimkitouni/.local/pipx/venvs/buildstream/lib/python3.11/site-packages/buildstream/sandbox/_sandboxreapi.py", line 234, in execute
        self.sandbox._run_with_flags(
[...]

PR #1962, opened 2024-10-11
sandbox: request remote execution output as a directory digest

abderrahim:
This gives us an output that is easier to manipulate given we're using buildbox-casd, and allows to avoid a whole class of bugs such as #1842 and #1961

The old code for handling the output as a Tree is kept for now as the remote execution spec recommends using it as a fallback

### DIFF
--- a/src/buildstream/sandbox/_sandboxremote.py
+++ b/src/buildstream/sandbox/_sandboxremote.py
@@ -294,12 +294,20 @@ class SandboxRemote(SandboxREAPI):
 
         # Fetch outputs
         for output_directory in action_result.output_directories:
-            tree_digest = output_directory.tree_digest
-            if tree_digest is None or not tree_digest.hash:
-                raise SandboxError("Output directory structure had no digest attached.")
-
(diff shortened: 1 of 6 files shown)


src/buildstream/sandbox/_sandboxremote.py
```diff
@@ -294,12 +294,20 @@ def _execute_action(self, action, flags):
 
         # Fetch outputs
         for output_directory in action_result.output_directories:
+            # Now do a pull to ensure we have the full directory structure.
+            # We first try the root_directory_digest we requested, then fall back to tree_digest
+
+            root_directory_digest = output_directory.root_directory_digest
+            if root_directory_digest and root_directory_digest.hash:
+                cascache._fetch_directory(casremote, root_directory_digest)
```
juergbi: Nit, `_fetch_directory()` should be marked as public (internal) by removing the leading underscore as we're calling it from outside `cascache.py`.

juergbi (approved):

PR #1965, opened 2024-10-31
_stream: Only query cache for the requested elements

abderrahim:
query_cache() used to query the cache for the elements passed and all their (build and runtime) dependencies. This is wasteful since most of the time, we don't need all this cache querying. And for the cases we need, it is the job of the query_cache() caller to include the dependencies they need.

As it turns out, this is only needed by `bst show` to be able to figure out the cache key in all circumsances.

This commit makes this behaviour optional, and only enables it from `bst show`.

### DIFF
--- a/src/buildstream/_frontend/cli.py
+++ b/src/buildstream/_frontend/cli.py
@@ -649,7 +649,7 @@ def show(app, elements, deps, except_, order, format_):
 
         # Don't spend time interrogating the cache if we don't need to show element state
         if need_state:
-            app.stream.query_cache(dependencies)
+            app.stream.query_cache(dependencies, need_state=True)
 
         if order == "alpha":
             dependencies = sorted(dependencies)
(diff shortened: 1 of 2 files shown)


juergbi:
I think this makes sense. However, do I see this right that it doesn't make a difference for typical builds as we always load all elements to support dynamic planning (unless `--deps build` is specified)? So this is mainly an optimization for non-build commands, or am I missing something?

abderrahim:
> I think this makes sense. However, do I see this right that it doesn't make a difference for typical builds as we always load all elements to support dynamic planning (unless --deps build is specified)? So this is mainly an optimization for non-build commands, or am I missing something?

That's correct. The main optimisation here is for artifact checkout.

The commit message here hasn't been updated to cover my most recent findings. I'll try to update it later today.

juergbi (approved):

gtristan (approved):
LGTM

PR #1966, opened 2024-11-01
cascache.py: Use `UploadTree` with `storage-service`, if available

juergbi:
Cache query is currently relatively slow when `storage-service` is enabled as the tree traversals are handled by Python code. Calling `UploadTree` moves that logic to buildbox-casd, which is significantly faster.

In local tests I've seen a 2-4x improvement of the cache query speed.

The corresponding buildbox-casd change is pending in https://gitlab.com/BuildGrid/buildbox/buildbox/-/merge_requests/757. The current code is kept as fallback path, i.e., current/older versions of buildbox-casd continue to be supported, without the optimization.

### DIFF
--- a/src/buildstream/_artifact.py
+++ b/src/buildstream/_artifact.py
@@ -339,7 +339,7 @@ class Artifact:
 
         buildroot_digest = self._get_field_digest("buildroot")
         if buildroot_digest:
-            return self._cas.contains_directory(buildroot_digest, with_files=True)
+            return self._cas.contains_directory(buildroot_digest)
         else:
             return False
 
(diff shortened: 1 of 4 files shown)


abderrahim (approved):

abderrahim:
Is this the last missing piece of https://github.com/apache/buildstream/issues/1042? Or is there more to it?

juergbi:
> Is this the last missing piece of #1042? Or is there more to it?

There is still room for further optimizations (not related to `storage-service`) with `UploadTree` (and maybe also `FetchTree`). That said, I'm fine with closing #1042 and opening new issues as needed.

PR #1971, opened 2024-11-22
Drop support for Python 3.8 and add support for Python 3.13

juergbi:
Python 3.8 is no longer maintained upstream and some Python packages have dropped support for it. Updating dependencies is required for Python 3.13 support but will break Python 3.8 support.


abderrahim:
Please check the code, we might have some python 3.8 specific codepaths that we can drop.

juergbi:
> Please check the code, we might have some python 3.8 specific codepaths that we can drop.

I don't see any Python 3.8-specific code paths. However, I've removed one left-over conditional from Python 3.7.

abderrahim (approved):

PR #1972, opened 2024-11-22
tests/testutils/casd.py: Set cache quota

juergbi:
This fixes `test_cache_usage_monitor` with recent versions of buildbox-casd that no longer track the disk usage of blobs if no absolute quota is configured.

https://gitlab.com/BuildGrid/buildbox/buildbox/-/merge_requests/670

Example of test failure: https://gitlab.com/BuildStream/buildstream-docker-images/-/jobs/8443036863

### DIFF
--- a/tests/testutils/casd.py
+++ b/tests/testutils/casd.py
@@ -20,7 +20,13 @@ from buildstream._cas import CASCache, CASDProcessManager, CASLogLevel
 @contextmanager
 def casd_cache(path, messenger=None):
     casd = CASDProcessManager(
-        str(path), os.path.join(str(path), "..", "logs", "_casd"), CASLogLevel.WARNING, None, None, True, None
+        str(path),
+        os.path.join(str(path), "..", "logs", "_casd"),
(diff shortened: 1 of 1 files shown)


tests/testutils/casd.py
```diff
@@ -20,7 +20,13 @@
 @contextmanager
 def casd_cache(path, messenger=None):
     casd = CASDProcessManager(
-        str(path), os.path.join(str(path), "..", "logs", "_casd"), CASLogLevel.WARNING, None, None, True, None
+        str(path),
+        os.path.join(str(path), "..", "logs", "_casd"),
+        CASLogLevel.WARNING,
+        16 * 1024 * 1024,
```
abderrahim, line 26: Is this 16M. I'm not sure that's enough. But let's let the tests decide.

juergbi: Yes, it seems to be enough for the tests but can increase it if you think this doesn't leave enough headroom. (The Buildbarn test failure is not related to this branch, I'll fix that separately.)

PR #1975, opened 2024-11-29
_loader/loader.py: Add `None` checks for shallow loading

juergbi:
When performing shallow loads without loading subprojects, `get_loader()` may return `None`. Add checks for that to fix crashes dereferencing `None`.


gtristan:
It would be good to document the functions which can *newly* now return `None`, and hopefully even document *why* they can now return `None`.

gtristan (approved):
Looks good to me, this might or might not fix the crashes but doesn't break tests and is otherwise harmless, possibly helping us track down issues which were leading to this None dereferencing.

PR #1978, opened 2024-12-20
Update default min-version for bst init to 2.4

juergbi:
A test fails without this now that the first 2.4 tag has been created:

```FAILED tests/frontend/init.py::test_defaults - AssertionError: assert '2.3' == '2.4'```


nanonyme:
Makes me wonder if we could somehow dynamically figure out the correct major.minor from Python metadata.

PR #1979, opened 2025-01-07
contrib/bst-graph: Escape names when using as node_id

harrysarson:
Colons in node_ids generate invalid graphviz, therefore escape the name
and use a `label` attribute to render correct name in the graph.

This PR also uses a raw string for the regex to fix `SyntaxWarning: invalid escape sequence '\|'`.

### DIFF
--- a/contrib/bst-graph
+++ b/contrib/bst-graph
@@ -29,6 +29,7 @@ installed.
 import argparse
 import subprocess
 import re
+import urllib.parse
 
 from graphviz import Digraph
 from ruamel.yaml import YAML
@@ -53,6 +54,22 @@ def parse_args():
         help='Open the rendered graph with the default application'
     )
     return parser.parse_args()
+
+
+def unique_node_name(s):
+    '''Generate unique node name for `s`.
+
+    Graphviz node names cannot contain colons or backslashes so we use
+    url-encoding to generate the unique node name. (A cryptographic hash could
+    be used instead but that would make the graphviz file less readable.)
+
+    Args:
+       s: element name
+
+    Returns:
+       A string containing the unique node name
+    '''
+    return urllib.parse.quote_plus(s)
 
 
 def parse_graph(lines):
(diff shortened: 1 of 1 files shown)


contrib/bst-graph
```diff
@@ -55,6 +56,10 @@ def parse_args():
     return parser.parse_args()
 
 
+def escape(s):
+    return urllib.parse.quote_plus(s.encode())
```
juergbi: Is this the right function to use for escaping? Might [graphviz.escape()](https://graphviz.readthedocs.io/en/stable/api.html#graphviz.escape) be a better fit (I haven't tested it)?

harrysarson: `graphviz.escape()` does not escape colons:
image




See also the note here on colons https://graphviz.readthedocs.io/en/stable/manual.html#node-ports-compass

juergbi: I see. What confused me is that graphviz doesn't actually expect URI/percent-encoded strings. As far as I can tell, this is using URI percent-encoding simply as a convenient way to generate unique identifiers that don't contain colons or backslashes. As labels are used, the identifiers don't really matter. (A cryptographic hash could be used instead but that would make the graphviz file less readable).

Can you please add a small comment that URI percent-encoding is used to generate unique identifiers that don't contain colons or backslashes?

Also, is the `.encode()` part actually necessary? `urllib.parse.quote_plus()` accepts strings and the default encoding is `utf-8`, so it seems to me that `.encode()` is redundant, or am I missing something?

juergbi:
Please prefix your commit summaries with `contrib/bst-graph: `

contrib/bst-graph
harrysarson: > Can you please add a small comment that URI percent-encoding is used to generate unique identifiers that don't contain colons or backslashes?

Added a small comment, also renamed the `escape` function to `unique_node_name`. Hopefully that makes the intent easier to grok for readers.

> Also, is the .encode() part actually necessary? urllib.parse.quote_plus() accepts strings and the default encoding is utf-8, so it seems to me that .encode() is redundant, or am I missing something?

Fixed

harrysarson:
> Please prefix your commit summaries with `contrib/bst-graph: `

Done

juergbi (approved):

PR #1987, opened 2025-01-24
Support gRPC connection options for retry and timeout

juergbi:


nanonyme:
Looks more or less reasonable to me. The mypy failure doesn't look related to this PR?

juergbi:
> Looks more or less reasonable to me. The mypy failure doesn't look related to this PR?

It was related to the regeneration of protobuf code with updated tooling. Fixed now.

doc/source/using_config.rst
```diff
@@ -145,6 +145,9 @@ toplevel of your configuration file, like so:
      # are used.
      connection-config:
        keepalive-time: 60
+       retry-limit: 4
+       retry-delay: 1000
+       request-timeout: 300
```
abderrahim: It would be a good idea to document the units given that retry-delay is in milliseconds while the other durations are in seconds.

abderrahim (approved):
Otherwise looks good.

doc/source/using_config.rst
juergbi: I've improved the documentation.

PR #1989, opened 2025-01-31
Write order key based on total order

nanonyme:
We follow these four rules always in order

1. Elements that are not runtime dependencies come first
2. Elements whose names are alphabetically earlier come first
3. Elements which are junctioned come first
4. Elements whose junction is alphabetically earlier come first

### DIFF
--- a/src/buildstream/_loader/loadelement.pyx
+++ b/src/buildstream/_loader/loadelement.pyx
@@ -14,8 +14,6 @@
 #  Authors:
 #        Tristan Van Berkom <tristan.vanberkom@codethink.co.uk>
 
-from functools import cmp_to_key
-
 from pyroaring import BitMap, FrozenBitMap  # pylint: disable=no-name-in-module
 
 from .._exceptions import LoadError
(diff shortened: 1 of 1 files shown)


src/buildstream/_loader/loadelement.pyx
```diff
@@ -379,44 +377,11 @@ cdef class LoadElement:
...
-            return -1
-    elif element_a.junction:
-        return -1
-    elif element_b.junction:
-        return 1
-
-    # This wont ever happen
-    return 0
+def _cmp_key(Dependency dep):
+    if element.junction:
```
nanonyme, line 385: If it's desired to use DependencyType, we could make first item in tuple an integer with value 0, 1, 2 corresponding build vs runtime vs all types.

nanonyme: Oh. Now I get why it's done like how it was.

juergbi:
We're not merging this right now as it changes staging order and cache keys with all Python versions. #1993 solves the issue with Python 3.13 without changing cache keys for Python <= 3.12.

A new algorithm, such as the one in this PR, may follow as optional alternative in the future.

nanonyme:
> We're not merging this right now as it changes staging order and cache keys with all Python versions. #1993 solves the issue with Python 3.13 without changing cache keys for Python <= 3.12.
> 
> A new algorithm, such as the one in this PR, may follow as optional alternative in the future.

I agree, this solution does not seem acceptable. The more I've been reading, the nore it seems real solution needs to be topological sort. We actually want specific order from elements where dependencies are taken into account. It's not certain if Python 3.12 order is correct either. Python 3.11 changed sort function and after that all bets are off.

Issue #1751, opened 2022-09-13
Recc inside BuildStream


ghost:
BuildStream is a tool to execute full "package" builds, including awkward parts like `configure` scripts, stripping, integration commands, in a repeatable way using a sandbox. It can distribute these builds to build farm workers via the REAPI, at a granularity of one element => one worker. This means, that a large element like WebKit is only compiled on a single worker which can be slow.

[Recc](https://gitlab.com/BuildGrid/recc) is a tool to distribute individual C/C++ compile tasks across build farm workers. At present, it cannot be used inside a BuildStream sandbox because it has no way to access the REAPI endpoint, but ... could we allow a controlled opening in the sandbox to enable Recc to be used inside element builds?

The ideal scenario would be a large element building as follows:

  1. `bst build` runs the element `configure-commands` on a single worker, with `recc` replacing `cc` etc.
  2. `bst build` runs the element `build-commands`:
    a. For each C/C++ compile operation, `recc` hashes the inputs for the given compile operation to produce a cache key, and checks the remote cache via configured REAPI endpoint
    b. If needed, compile is invoked on a remote worker. Then, the object file is downloaded from the cache.
  3. Link operations run on the main worker and still probably take ages to complete.
  4. Profit!

This is not guaranteed to be faster than building on a single node, a lot depends on the infrastructure being used and the element being built. I think it would be worth exploring as part of a larger project to try and reduce build times of large BuildStream projects like gnome-build-meta.

juergbi:
This could certainly be interesting if either no big workers are available in a build farm or if the project could benefit from more fine-grained caching. The idea of using recc with nested remote execution has been around for a while but nobody has attempted to implement it yet, as far as I know.

I think this would mainly be a feature of the REAPI server/worker side. E.g. BuildGrid/BuildBox could support a platform property that allows network access to the CAS and Execution services and sets environment variables to communicate the URIs. It might make sense to extend buildbox-casd to work as an Execution service proxy, making it possible to restrict the sandbox hole to the worker's local buildbox-casd.

BuildStream would simply need to support setting that platform property (possibly in a generic way, see #1313). BuildStream projects could then include recc in their sandbox and configure it using the environment variables set by the worker.

If recc supported buildbox-casd's `CaptureFiles` method, it could skip hashing of input files when buildbox-fuse is used as the hash would already available in the `user.checksum.sha256` xattr, reducing the recc overhead.

staehle:
[...]

PR #1991, opened 2025-02-05
buildelement: Add the digest-environment config property

### LINKED ISSUE #1984: Add dependency configuration to export a CAS digest as an environment variable
Context: https://lists.apache.org/thread/scsz1ybc25d3x1oy33onfmm8rhr0kddg

We want to be able to export the CAS digest of certain dependencies to the sandbox. This can be used in conjunction with something like https://github.com/apache/buildstream/pull/1772 to have more correct cache keys for bazel/RECC, but can be more generally useful on its own to include this information about dependencies in its own artifacts.

This is to be implemented via a [dependency configuration](https://docs.buildstream.build/2.4/main_glossary.html#term-Dependency-configuration), similar to the already existing [`location`](https://docs.buildstream.build/2.4/buildstream.buildelement.html#location-for-staging-dependencies). As this happens pretty late in the element lifecycle, it cannot be exported as a buildstream variable. It will instead be exported as an environment variable in the sandbox.

For example:

```
dependencies:
- foo.bst
- bar.bst
- filename: gcc.bst
  config:
    digest-environment: RECC_TOOLCHAIN_KEY
```

One thing to take into consideration implementeng this is that we don't yet have any way for a plugin to find the CAS digest of a dependency except for staging it into the sandbox. We will need to add some public API to allow this. Either public API to manipulate dependency artifacts as `Directory` instances directly, or a way to create a throwaway dummy sandbox, that can't run any commands but can be passed to existing APIs such as `Element.stage_artifact()` and `Element.stage_dependency_artifacts()`.

Kekun:
This allows setting an environment variable inside the sandbox to the CAS
digest of one or more dependencies.

Fixes https://github.com/apache/buildstream/issues/1984

### DIFF
--- a/src/buildstream/buildelement.py
+++ b/src/buildstream/buildelement.py
@@ -219,6 +219,7 @@ class BuildElement(Element):
     def configure_dependencies(self, dependencies):
 
         self.__layout = {}  # pylint: disable=attribute-defined-outside-init
+        self.__digest_environment = {}  # pylint: disable=attribute-defined-outside-init
 
         # FIXME: Currently this forcefully validates configurations
         #        for all BuildElement subclasses so they are unable to
@@ -227,9 +228,18 @@ class BuildElement(Element):
         for dep in dependencies:
             # Determine the location to stage each element, default is "/"
             location = "/"
+
             if dep.config:
-                dep.config.validate_keys(["location"])
-                location = dep.config.get_str("location")
+                dep.config.validate_keys(["digest-environment", "location"])
+
+                location = dep.config.get_str("location", "/")
+
+                digest_var_name = dep.config.get_str("digest-environment", None)
+
+                if digest_var_name is not None:
+                    element_list = self.__digest_environment.setdefault(digest_var_name, [])
+                    element_list.append(dep.element)
+
             try:
                 element_list = self.__layout[location]
             except KeyError:
@@ -285,10 +295,17 @@ class BuildElement(Element):
             command_dir = build_root
         sandbox.set_work_directory(command_dir)
 
-        # Setup environment
-        sandbox.set_environment(self.get_environment())
-
     def stage(self, sandbox):
+        # Setup environment
+        env = self.get_environment()
+
+        for digest_variable, element_list in self.__digest_environment.items():
+            dummy_sandbox = sandbox.create_sub_sandbox()
+            self.stage_dependency_artifacts(dummy_sandbox, element_list)
+            digest = dummy_sandbox.get_virtual_directory()._get_digest()
+            env[digest_variable] = "{}/{}".format(digest.hash, digest.size_bytes)
+
+        sandbox.set_environment(env)
 
         # First stage it all
         #
--- a/src/buildstream/sandbox/_sandboxremote.py
+++ b/src/buildstream/sandbox/_sandboxremote.py
@@ -261,9 +261,17 @@ class SandboxRemote(SandboxREAPI):
                 "Uploading input root", element_name=self._get_element_name()
             ):
                 # Determine blobs missing on remote
+                root_digests = [action.input_root_digest]
+
+                # Add virtual directories for subsandboxes
+                for subsandbox in self._get_subsandboxes():
+                    vdir = subsandbox.get_virtual_directory()
+                    root_digests.append(vdir._get_digest())
+
+                missing_blobs = []
                 try:
-                    input_root_digest = action.input_root_digest
-                    missing_blobs = list(cascache.missing_blobs_for_directory(input_root_digest, remote=casremote))
+                    for root_digest in root_digests:
+                        missing_blobs.extend(cascache.missing_blobs_for_directory(root_digest, remote=casremote))
                 except grpc.RpcError as e:
                     raise SandboxError("Failed to determine missing blobs: {}".format(e)) from e
 
--- a/src/buildstream/sandbox/_sandboxremote.py
+++ b/src/buildstream/sandbox/_sandboxremote.py
@@ -24,7 +24,7 @@ from .._remote import BaseRemote
 from .._protos.build.bazel.remote.execution.v2 import remote_execution_pb2, remote_execution_pb2_grpc
 from .._protos.build.buildgrid import local_cas_pb2
 from .._protos.google.rpc import code_pb2
-from .._exceptions import BstError, SandboxError
+from .._exceptions import SandboxError
 from .._protos.google.longrunning import operations_pb2, operations_pb2_grpc
 from .._cas import CASRemote
 
@@ -247,9 +247,7 @@ class SandboxRemote(SandboxREAPI):
         stdout, stderr = self._get_output()
 
         context = self._get_context()
-        project = self._get_project()
         cascache = context.get_cascache()
-        artifactcache = context.artifactcache
 
         action_digest = cascache.add_object(buffer=action.SerializeToString())
 
@@ -268,15 +266,6 @@ class SandboxRemote(SandboxREAPI):
                     missing_blobs = list(cascache.missing_blobs_for_directory(input_root_digest, remote=casremote))
                 except grpc.RpcError as e:
                     raise SandboxError("Failed to determine missing blobs: {}".format(e)) from e
-
-                # Check if any blobs are also missing locally (partial artifact)
-                # and pull them from the artifact cache.
-                try:
-                    local_missing_blobs = cascache.missing_blobs(missing_blobs)
-                    if local_missing_blobs:
-                        artifactcache.fetch_missing_blobs(project, local_missing_blobs)
-                except (grpc.RpcError, BstError) as e:
-                    raise SandboxError("Failed to pull missing blobs from artifact cache: {}".format(e)) from e
 
                 # Add command and action messages to blob list to push
                 missing_blobs.append(action.command_digest)
--- a/src/buildstream/sandbox/sandbox.py
+++ b/src/buildstream/sandbox/sandbox.py
@@ -85,13 +85,18 @@ class Sandbox:
         self.__env = None  # type: Optional[Dict[str, str]]
         self.__mount_sources = {}  # type: Dict[str, str]
         self.__allow_run = True
+        self.__subsandboxes = []
 
         # Plugin element full name for logging
         plugin = kwargs.get("plugin", None)
         if plugin:
             self.__element_name = plugin._get_full_name()
         else:
-            self.__element_name = None
+            parent = kwargs.get("parent", None)
+            if parent:
+                self.__element_name = parent._get_element_name()
+            else:
+                self.__element_name = None
 
         # Configuration from kwargs common to all subclasses
         self.__config = kwargs["config"]
@@ -263,6 +268,24 @@ class Sandbox:
                 self.__batch = None
 
             batch.execute()
+
+    def create_subsandbox(self, **kwargs):
+        """Create an empty sandbox
+
+        This allows an element to use a secondary sandbox for manipulating artifacts
+        that does not affect the build sandbox
+        """
+
+        sub = Sandbox(
+            self.__context,
+            self.__project,
+            parent=self,
+            stdout=self.__stdout,
+            stderr=self.__stderr,
+            config=self.__config,
+        )
+        self.__subsandboxes.append(sub)
+        return sub
 
     #####################################################
     #    Abstract Methods for Sandbox implementations   #
@@ -564,6 +587,9 @@ class Sandbox:
     #
     def _disable_run(self):
         self.__allow_run = False
+
+    def _get_subsandboxes(self):
+        return self.__subsandboxes
 
 
 # SandboxFlags()


src/buildstream/buildelement.py
```diff
@@ -228,14 +228,23 @@ def configure_dependencies(self, dependencies):
...
                 element_list = self.__layout[location]
             except KeyError:
                 element_list = []
                 self.__layout[location] = element_list
 
+            # Determine the optional digest environment variable to stage each
+            # element,default is None
+            digest_environment = None
+            if dep.config:
+                dep.config.validate_keys(["digest-environment", "location"])
```
abderrahim, line 233: You only need to validate the keys once (above is good)

src/buildstream/buildelement.py
```diff
@@ -228,14 +228,23 @@ def configure_dependencies(self, dependencies):
...
+                if "digest-environment" in dep.config:
+                    digest_environment = dep.config.get_str("digest-environment")
+            self.__digest_environment = digest_environment
```
abderrahim: This should  be a dict so as to allow more than one element to set the new config option.

src/buildstream/buildelement.py
```diff
@@ -285,10 +297,19 @@ def configure_sandbox(self, sandbox):
             command_dir = build_root
         sandbox.set_work_directory(command_dir)
 
+    def stage(self, sandbox):
         # Setup environment
```
abderrahim, line 299: I had to move this from `configure_sandbox()` to `stage()` because buildstream wouldn't let me stage dependency artifacts outside of `stage()`

src/buildstream/buildelement.py
```diff
@@ -285,10 +297,19 @@ def configure_sandbox(self, sandbox):
...
-        sandbox.set_environment(self.get_environment())
+        env = self.get_environment()
 
-    def stage(self, sandbox):
+        for digest_variable, element_list in self.__digest_environment.items():
+            dummy_sandbox = SandboxDummy(
+                self._get_context(), self._get_project(), plugin=self, stdout=None, stderr=None, config={}
+            )
```
abderrahim: I wonder if we should try to have a nicer way to create a dummy sandbox

src/buildstream/buildelement.py
```diff
@@ -285,10 +297,19 @@ def configure_sandbox(self, sandbox):
...
+            self.stage_dependency_artifacts(dummy_sandbox, element_list)
+            digest = dummy_sandbox.get_virtual_directory()._get_digest()
```
abderrahim, line 305: The linter complains here because `_get_digest()` is only defined for `CasBasedDirectory` (not the base `Directory` class) and `get_virtual_directory()` is advertised to return a `Directory`.

What would be the best way forward?

src/buildstream/buildelement.py
```diff
@@ -227,9 +229,19 @@ def configure_dependencies(self, dependencies):
...
+
             if dep.config:
-                dep.config.validate_keys(["location"])
-                location = dep.config.get_str("location")
+                dep.config.validate_keys(["digest-environment", "location"])
+
+                if "location" in dep.config:
+                    location = dep.config.get_str("location")
+
+                if "digest-environment" in dep.config:
```
gtristan: Instead of doing a double dictionary lookup, we should use the default `None` to `MappingNode.get_str()`.

```python
   var_name = dep.config.get_str("digest-environment", None)
   if var_name is not None:
       element_list = blablabla
```

src/buildstream/buildelement.py
```diff
@@ -285,10 +297,19 @@ def configure_sandbox(self, sandbox):
...
         # Setup environment
-        sandbox.set_environment(self.get_environment())
+        env = self.get_environment()
 
-    def stage(self, sandbox):
+        for digest_variable, element_list in self.__digest_environment.items():
+            dummy_sandbox = SandboxDummy(
+                self._get_context(), self._get_project(), plugin=self, stdout=None, stderr=None, config={}
+            )
+            self.stage_dependency_artifacts(dummy_sandbox, element_list)
```
gtristan, line 304: In the case that we are staging elements in `element_list` which don't depend on eachother, I don't believe that `Element.stage_dependency_artifacts()` will do any sorting.

In the case that we have resulting overlaps in the resulting staged trees, I am worried that the resulting CAS digest may differ depending on the order in which the dependencies are declared in the buildstream `.bst` element.

Also there is a concern that overlaps may not turn out to be the same as when staging the element's overall dependencies.

I think abderrahim also noted the sandbox staging seems a bit weird, or could be done nicer.

I don't believe it should be algorithmically necessary to do the actual staging in order to calculate what the resulting CAS digest would be, probably there should be a way to support this better in the `CasBasedDirectory` code without doing an actual staging process.

abderrahim: > In the case that we are staging elements in element_list which don't depend on eachother, I don't believe that Element.stage_dependency_artifacts() will do any sorting.

I find that statement weird given that `stage_dependency_artifact()` is advertised to do just that.

I would also argue that if this is true, then it is orthogonal to the changes in this pull request. There is no difference between

```
filename:
- a.bst
- b.bst
- c.bst
config:
  location: /sysroot
```

and 

```
filename:
- a.bst
- b.bst
- c.bst
config:
  digest-environment: DEPENDENCIES_CAS_DIGEST
```

in the current code, except for the fact that the former stages in a subdirectory of the build sandbox and the latter stages in a dummy sandbox.

> I don't believe it should be algorithmically necessary to do the actual staging in order to calculate what the resulting CAS digest would be, probably there should be a way to support this better in the CasBasedDirectory code without doing an actual staging process.

Well, it depends. Currently an element plugin has no way to create a `CasBasedDirectory` using public API. The only way to get a `CasBasedDirectory` is throgh `Sandbox.get_virtual_directory()` (and even that advertises it returns a `Directory` base class). It also can't get the artifact contents of an element as a `Directory`: it only has access to `Element.stage_artifact()`. While this `BuildElement` is in core and can technically use private APIs, I tried to avoid going that route.

I think the difference we need to consider between the two approaches is whether or not to consider the dependencies? i.e. if I depend on gcc, is your expectation that this only gets the digest of the gcc artifact or should it be gcc and its runtime dependencies?

src/buildstream/buildelement.py
abderrahim: Done.

src/buildstream/buildelement.py
gtristan: @abderrahim writes:

> >   In the case that we are staging elements in element_list which don't depend on eachother, I don't believe that Element.stage_dependency_artifacts() will do any sorting.
>
> I find that statement weird given that stage_dependency_artifact() is advertised to do just that.

Right, this text could be clarified further, this function says:

> *"This is primarily a convenience wrapper around `Element.stage_artifact()` which takes care of staging all the dependencies in staging order and issueing the appropriate warnings."*

Which is not false, it does stage *all of the specified element* and it does so *in staging order*, however if you look at what the code does, it does not do any toplevel sorting of the input element sequence.


> I would also argue that if this is true, then it is orthogonal to the changes in this pull request. There is no difference between:
>
> ```yaml
> depends:
> - a.bst
> - b.bst
> - c.bst
>   config:
>     location: /sysroot
> ```
> and
>
> ```yaml
> depends:
> - a.bst
> - b.bst
> - c.bst
>   config:
>     digest-environment: DEPENDENCIES_CAS_DIGEST
> ```
> in the current code, except for the fact that the former stages in a subdirectory of the build sandbox and the latter stages in a dummy sandbox.

I think you are missing my meaning with your example.

So here is my hypothesis; given your `a`, `b`, `c` example, let us assume that the toplevel element `e` in question depends on `b` and `c` which may also depend on `a`, where `b` and `c` install the same file.

e.g.:

```
            e
          /   \
         b     c  <--- both "b" and "c" install /etc/abc.conf
          \   / 
            a
```

What I am saying, is that with this implementation, the value of `DEPENDENCIES_CAS_DIGEST` may differ in the following two configurations.

### Configuration 1

Here we stage element `e` deterministically into the sandbox, and stage `c`'s version of `/etc/abc.conf` over `b`'s version of `/etc/abc.conf` in an orthogonal, extra stage operation which we use to determine the cas digest `DEPENDENCIES_CAS_DIGEST`.

```yaml
depends:
- a.bst
- b.bst
  config:
    digest-environment: DEPENDENCIES_CAS_DIGEST
- c.bst
  config:
    digest-environment: DEPENDENCIES_CAS_DIGEST
```

### Configuration 2

Here we do the same as configuration 1, except that `b`'s version of `/etc/abc.conf` is staged on top of `c`'s version of the same file, resulting in a *different digest*.

```yaml
depends:
- a.bst
- c.bst
  config:
    digest-environment: DEPENDENCIES_CAS_DIGEST
- b.bst
  config:
    digest-environment: DEPENDENCIES_CAS_DIGEST
```

### A closer look

The body of `Element.stage_dependency_artifacts()` is doing the following:

```python
with self._overlap_collector.session(action, path):
    for dep in self.dependencies(selection):
        dep._stage_artifact(sandbox, path=path, include=include, exclude=exclude, orphans=orphans, owner=self)
```

And, looking at `Element.dependencies()`, we can quickly see that when the `selection` argument is given to select *"Subsets of the dependency graph"*, the `self` element is not considered at all in the algorithm.

This means that the overall deterministic staging order *of element `e`* is not considered when looping over the dependencies of the `selection` argument given to `Element.stage_dependency_artifacts()`.

Arguably this algorithm in `Element.stage_dependency_artifacts()` needs to be made more deterministic in this regard, although I wouldn't necessarily say that the docs are lying here - they are yielded *in staging order* relative to *eachother*, but they are not yielded in *the staging order of the given element*.

And so I come back to your original argument:

> I would also argue that if this is true, then it is orthogonal to the changes in this pull request.

I have a point of contention here.

I have not yet looked at whether the `location` dependency configuration suffers from this issue as well, however, if we have this issue with the `location` dependency configuration, it needs to be fixed, and I also do not want to go as far as landing *this change* in the knowledge that this is indeed an issue.

### Prove me wrong :)

There may be other ways that I am wrong about this bug, but I can think of one possibility:

It is possible that in the `Element` loading/instantiation process, that the dependency configurations are exposed to the plugin *in staging order* rather than in the order in which they are listed in the element declaration (`.bst` file)... if this is the case, `Element.dependencies()` may be sufficient for both of these use cases.

It would however be good to add a test with this merge request, asserting that we get the same CAS digest environment variable using both orderings of element `e` in my overlapping example above.

gtristan:
I'll open a completely separate thread to follow up on the separate topic of "doing a whole stage" (I probably should have made that comment separately initially).

@abderrahim writes:

> Well, it depends. Currently an element plugin has no way to create a CasBasedDirectory using public API. The only way to get a CasBasedDirectory is throgh Sandbox.get_virtual_directory() (and even that advertises it returns a Directory base class). It also can't get the artifact contents of an element as a Directory: it only has access to Element.stage_artifact(). While this BuildElement is in core and can technically use private APIs, I tried to avoid going that route.

I should point out that `SandboxDummy` is declared in `sandboxing/_sandboxdummy.py`, and not exposed in the toplevel `__init__.py`, and so it is not a public API, I'm not sure there is a sensible reason to make this public.

It does appear in `sandbox/__init__.py` which may strictly speaking be a bug on our part, as it may mislead people into thinking it is public.

> I think the difference we need to consider between the two approaches is whether or not to consider the dependencies? i.e. if I depend on gcc, is your expectation that this only gets the digest of the gcc artifact or should it be gcc and its runtime dependencies?

I think this definitely needs to consider dependencies, although I'm not sure how that is related to using a `SandboxDummy`.

Frankly, the use of `SandboxDummy` is not a huge deal breaker for me if it really saves you some lines of code, my concerns about using it are:
* Bringing in unrelated lines of code, i.e. *the use of a sandbox object*
  * Is it possible that sharing code in this way may cause sandbox related changes to break this feature ?
* The possibility that those orthogonal lines of code may incur an unnecessary runtime overhead
  * Does this increase I/O in the calculation of the `digest-environment` ?

src/buildstream/buildelement.py
abderrahim: > It would however be good to add a test with this merge request, asserting that we get the same CAS digest environment variable using both orderings of element e in my overlapping example above.

I agree with this. I was just pointing out that if this behaviour is non-deterministic, then it's a pitfall that also applies to all plugins that make use of `Element.stage_dependency_artifacts()` and is not something specific to this use case.

abderrahim:
> I should point out that `SandboxDummy` is declared in `sandboxing/_sandboxdummy.py`, and not exposed in the toplevel `__init__.py`, and so it is not a public API, I'm not sure there is a sensible reason to make this public.
> 
> It does appear in `sandbox/__init__.py` which may strictly speaking be a bug on our part, as it may mislead people into thinking it is public.

I understand it's not intended to be public, and I commented about it earlier. However my main point is that the only public way to gain access to an element's artifact is through `Element.stage_artifact()` (and `Element.stage_dependency_artifacts()`) which depends on having a sandbox.

I feel that adding an API to expose a dummy sandbox is easier than trying to figure out what part of `self.__artifact.get_files()` to expose. Do we expose the `Artifact` class? or just the files virtual directory?

> > I think the difference we need to consider between the two approaches is whether or not to consider the dependencies? i.e. if I depend on gcc, is your expectation that this only gets the digest of the gcc artifact or should it be gcc and its runtime dependencies?
> 
> I think this definitely needs to consider dependencies, although I'm not sure how that is related to using a `SandboxDummy`.

`Element.stage_dependency_artifacts()` is the easiest (and I believe the recommended) way to get an element's artifact including its dependencies. And that requires a sandbox. It doesn't have to be a dummy sandbox, but a dummy sandbox is already used elsewhere as a fallback when a "real" sandbox is not available.
 
> Frankly, the use of `SandboxDummy` is not a huge deal breaker for me if it really saves you some lines of code, my concerns about using it are:
> * Bringing in unrelated lines of code, i.e. _the use of a sandbox object_
>   * Is it possible that sharing code in this way may cause sandbox related changes to break this feature ?
> * The possibility that those orthogonal lines of code may incur an unnecessary runtime overhead
>   * Does this increase I/O in the calculation of the `digest-environment` ?

I don't think we need to worry about any of these. When staging, the `Sandbox` object is only used for `Sandbox.get_virtual_directory()`. The only unrelated lines of code that get executed are the `Sandbox.__init__()` and `SandboxDummy.__init__()`, which are only doing attribute assignment. I don't think the runtime overhead is significant, and there is no extra I/O involved compared to staging directly to a virtual (cas-based) directory.

In an ideal world, I'd prefer that `Element.stage_directory()` take a `Directory` rather than a `Sandbox`. Then for this MR, I'd want a way for a plugin to create a cas-based `Directory` instance. But we're stuck with `Element.stage_directory()` taking a `Sandbox` for now.

I'm happy to do it differently if you can think of a better way of doing it.

gtristan:
> In an ideal world, I'd prefer that `Element.stage_directory()` take a `Directory` rather than a `Sandbox`. Then for this MR, I'd want a way for a plugin to create a cas-based `Directory` instance. But we're stuck with `Element.stage_directory()` taking a `Sandbox` for now.

Right, I think you mean `Element.stage_artifact()`, and yeah it *might* be cool if this was `Directory` based.

On the surface, I feel mostly concerned with the overhead of *staging the artifact*. This activity, at least on the surface, looks like an activity which involves I/O, when all we are actually looking for is *what CAS digest would the root `Directory` have, if this set of artifacts were to be staged*.

Looking at the operation, I see that we're essentially using `CasBasedDirectory._import_files_internal(Directory, ...)`. For a `CasBasedDirectory` being imported - this essentially lands up in `CasBasedDirectory.__partial_import_cas_into_cas()`.

While the code around here is a bit tricky to follow, I think I'm not really seeing any I/O, instead what generally happens during a build is:
* `Sanbox` creates it's vdir
* Elements stage artifacts into that vdir
* This process only plays with data structures in the vdir `CasBasedDirectory` object hierarchy
* `SandboxREAPI._run()` then serializes the in memory `CasBasedDirectory` into protobufs and delegates this to `SandboxREAPI._execute_action`, which can be handled by either `SandboxBuildboxRun` or `SandboxRemote`

So it looks like the I/O of managing the cache only ever happens immediately before a build activity or after when we cache the artifact.

In summary, I think I've looked at this enough to feel fairly confident that this activity is not really I/O intensive.

> 
> I'm happy to do it differently if you can think of a better way of doing it.

I think I'm pretty happy with this approach using a dummy sandbox for this, after having looked deeper into the implementation details :)

src/buildstream/buildelement.py
```diff
@@ -285,10 +296,19 @@ def configure_sandbox(self, sandbox):
...
+            digest = dummy_sandbox.get_virtual_directory()._get_digest()
+            env[digest_variable] = "{}/{}".format(digest.hash, digest.size_bytes)
```
abderrahim, line 306: Another issue I discovered when testing with remote execution: If using this with recc/bazel (https://github.com/apache/buildstream/issues/1751), for setting a remote execution platform property (e.g. chrootRootDigest) then we also need to upload it to the remote execution CAS.

src/buildstream/buildelement.py
nanonyme: Would cast work if you know here that you're actually dealing with specific subtype?

abderrahim:
Pushed this into the main repository as https://github.com/apache/buildstream/pull/2036. Closing.

I will try to transfer the non-resolved discussions to the new PR.

src/buildstream/buildelement.py
nanonyme: > > It would however be good to add a test with this merge request, asserting that we get the same CAS digest environment variable using both orderings of element e in my overlapping example above.
> 
> I agree with this. I was just pointing out that if this behaviour is non-deterministic, then it's a pitfall that also applies to all plugins that make use of `Element.stage_dependency_artifacts()` and is not something specific to this use case.

Maybe even worth regarding this a bug (design flaw) that should have its own ticket.

PR #1994, opened 2025-02-14
_frontend/widget: Allow showing the CAS digest of built artifacts

Kekun:
This adds the `artifact-cas-digest` format string to the show command.

This is a first step implementing https://github.com/apache/buildstream/issues/1984.

- [x] Add tests, e.g. building an artifact and comparing the printed digest


src/buildstream/storage/_casbaseddirectory.py
```diff
@@ -529,6 +529,10 @@ def is_dir_in(entry: _IndexEntry, directory: CasBasedDirectory) -> bool:
 
         self.__invalidate_digest()
 
+    def get_cas_digest(self):
```
abderrahim: This adds a public method to a private class, and it feels weird.

If we want to add a public method, it should be added to the base `Directory` class and be implemented by both subclasses.

That said, I feel we should not add public method at this time as nothing is to be exposed to plugins. We can consider adding the public method later in #1991 and just use the internal methods for now.

src/buildstream/element.py
```diff
@@ -1272,6 +1272,22 @@ def _get_cache_key(self, strength=_KeyStrength.STRONG):
...
+    #
+    # Returns:
+    #    (str|None): The CAS digest for this Element, or None
+    #
+    # None is returned if no CAS artifact is available for this element.
+    #
+    def _get_cas_digest(self):
+        if self.__artifact is None:
+            return None
+        if not self.__artifact.query_cache():
```
abderrahim: Is this still needed? Now that you're correctly querying the cache at the beginning?

src/buildstream/_frontend/widget.py
```diff
@@ -437,6 +437,14 @@ def show_pipeline(self, dependencies, format_):
                 runtime_deps = [e._get_full_name() for e in element._dependencies(_Scope.RUN, recurse=False)]
                 line = p.fmt_subst(line, "runtime-deps", _yaml.roundtrip_dump_string(runtime_deps).rstrip("\n"))
 
+            # CAS Digest
+            if "%{cas-digest" in format_:
+                cas_digest = element._get_cas_digest()
```
abderrahim: I kinda think we should implement this here for the moment. Call `element._get_artifact()` to get the artifact, and try to format the output here.

Later, we could have a better API to expose to the plugins, and we can use it here too.

man/bst-show.1
```diff
@@ -45,6 +45,7 @@ Show elements in the pipeline
         %{deps}           A list of all dependencies
         %{build-deps}     A list of build dependencies
         %{runtime-deps}   A list of runtime dependencies
+        %{cas-digest}     The CAS digest
```
abderrahim: A bit of bike shedding:
* "The CAS digest" is too short to explain what this does. The CAS digest of what exactly?
* I feel that `cas-digest` is too generic. This is actually the cas digest of the built artifact, we could (maybe not now, but in the future) have cas digests of other things (like the source code, or the buildtree if we have one)

Kekun:
I added a test, it's probably not enough though…

gtristan:
> I feel that cas-digest is too generic. This is actually the cas digest of the built artifact, we could (maybe not now, but in the future) have cas digests of other things (like the source code, or the buildtree if we have one)

I tend to agree, while the “CAS” is an implementation detail it’s fairly well known, I think I’d rather avoid exposing the word in APIs if we don’t already have it.

Also valid point about source caches.

let’s use another name, perhaps `content-hash` ? Maybe that doesn’t capture the implication that it is the *artifact* content… maybe `output-hash` or `artifact-content-hash` ?

Also, was this man page updated using the automated generation ? We should only update the command description in `cli.py` and use the python command to automatically update all the man pages and commit that as a separate commit… that command is documented… somewhere in our `HACKING` file(s)….

Kekun:
Isn't "hash" more technical than "digest" too (genuine question, I don't know)? What about just `artifact-digest` (or `artifact-hash` if you prefer)? Any reason to prefer "content"? I'll let you two decide the right words to use as you are the BuildStream specialists. :)

The man page what updated manually, I'll update it in a different commit then.

gtristan:
Regarding testing in this PR: Yes the added test is good to add, but not at all sufficient IMO.

What I would like to see here...

* Take the `tests/frontend/artifact_checkout.py` test as an example of a test that:
  * Actually builds something
  * Pushing the artifact to a remote artifact cache
* Make sure you create a new directory for the test project data
  * In the past we had many tests using the same directory, e.g. `tests/frontend/project` still has a bit of this mess going on
  * We made efforts to tear this apart ensuring that individual tests have their own static committed project directories, please take this new approach as it is easier to maintain, and modifying one test does not subtly affect adjacent tests
* You can add this new test to `tests/frontend/show.py`, or to a new python file (do not sneak the new unrelated test additional test into `artifact_checkout.py` of course ;-) )
* The test, like the artifact checkout test, can use a simple `import` element of a committed text file
  * This is expected to be 100% reproducible, so the CAS digest should always be the same
  * Record this CAS digest (whether we call this a "hash" or whatever in the final CLI UI, just take note of it)
* The test body essentially does this:
  * Build the element
  * Run `bst show --format ...` to obtain the CAS digest
  * Assert that the CAS digest is _exactly the one you previously recorded_, i.e. hard coded CAS digest
* Test multiple scenarios
  * Building with and without the remote cache enabled
  * Building with remote cache enabled, deleting the artifact locally, before `bst show`
    * In this scenario I think we expect to not have the CAS digest
  * Building with remote cache enabled, delete the artifact locally, then `bst artifact pull` the artifact, and `bst show`
    * In this scenario we expect to have exactly the hard coded CAS digest

I realize this is a lot of text, but it shouldn't be too difficult to do in practice, there are examples of all of this in the tests already.

tests/frontend/show.py
```diff
@@ -553,6 +554,64 @@ def test_strict_dependencies(cli, datafiles, target, expected_state):
...
+#  * Modify some local files which are imported
+#    by an import element which the target depends on
+#  * Assert that the cached state of the target element
+#    is as expected
+#
+# We run the test twice, once with an element which strict
+# depends on the changing import element, and one which
+# depends on it regularly.
+#
+@pytest.mark.datafiles(os.path.join(DATA_DIR, "project"))
```
gtristan: I specifically asked that we create a _new_ directory with a small project rather than sharing a directory with adjacent tests, this is an old mistake which is mostly (at least 50%) corrected.

src/buildstream/_frontend/widget.py
```diff
@@ -437,6 +437,21 @@ def show_pipeline(self, dependencies, format_):
...
+                    artifact = None
+                if artifact is not None:
+                    artifact_files = artifact.get_files()
+                    # We call the private CasBasedDirectory._get_digest() for
+                    # the moment, we should make it public on Directory.
+                    artifact_digest = artifact_files._get_digest()
+                    formated_artifact_digest = "{}/{}".format(artifact_digest.hash, artifact_digest.size_bytes)
+                    line = p.fmt_subst(line, "artifact-cas-digest", formated_artifact_digest)
+                else:
+                    line = p.fmt_subst(line, "artifact-cas-digest", "(no artifact CAS digest)", fg="yellow")
```
gtristan: I think that this should output an empty string rather than a user readable indicator.

This is expected to be usable as an interface for scripting with, and I expect that an empty string is easier to distinguish a non-existent digest compared to a human readable output in stderr.

Instead we could consider a user facing log. i.e. we could call `Plugin.warn()` here and say `element.warn("Cannot obtain CAS digest because artifact is not cached")`, or better yet we could collect all of the elements that are not cached and issue a single warning message.

The difference is that warnings and logs are sent to `stderr` and are considered "UI" and as such do not interfere with scripting output.

tests/frontend/show.py
```diff
@@ -553,6 +554,64 @@ def test_strict_dependencies(cli, datafiles, target, expected_state):
...
+    [
+        ("target.bst", {
+            "compose-all.bst": "c742f599e198f348ba7600bf50194ae45af6cba759e0005dcd980ad596c51959/78",
+            "import-bin.bst": "594334e3e9f15c9eac5b8325befc3f53af8ffbaa664424bb308484abb892944f/77",
+            "import-dev.bst": "42203f9284db4817a2fc2d57714c6c92ec318f82a23a25e5325852e5dd6effc3/77",
+            "target.bst": "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855/0",
+        }),
+    ],
+)
+def test_format_artifact_cas_digest(cli, datafiles, target, expected_digests):
```
gtristan: This covers some of the conditions I asked for, but I'm interested in really determining all of the possible states.

* If the element was never built locally
  * And the element is not existing in the remote cache (covered already, no remote used in this test)
  * And the element is available in the remote (... lets find out what the behavior is and test for it...)
* If the element was built locally (covered already)
* If the element was built locally and then deleted (... lets test, do the right bits of the underlying objects go away ? ...)

tests/frontend/show.py
```diff
@@ -553,6 +554,64 @@ def test_strict_dependencies(cli, datafiles, target, expected_state):
...
+#
+# We run the test twice, once with an element which strict
+# depends on the changing import element, and one which
+# depends on it regularly.
+#
+@pytest.mark.datafiles(os.path.join(DATA_DIR, "project"))
+@pytest.mark.parametrize(
+    "target, expected_digests",
+    [
+        ("target.bst", {
```
gtristan: Note that, as I am asking for a _separate project directory_ be created for this test to stand on its own and not mingle data with other tests... I really think it is sufficient to just have a project with less elements.

I'd be happy enough with one element, but it would be useful to test various different import elements:
* A regular file with some simple example content e.g. *"The perverted pink pony mounts the unwilling hippopotamus"*
* A regular file with execute permission
* A symbolic link

The value here might be to check that we indeed get differing content hashes for individual files with different modes/types, however... while I didn't think of asking for this, and I'm highly confident that such content hashes are going to *differ*, it might be interesting to actually know.

Kekun: I confirm the hash changes when a file is executable or not. I tested it with a working test, then running `chmod +x` on an imported file, ran the tests and got a different digest, then `chmod -x` on the same file and the old digest matched anew.

gtristan: Right so lets include some different samples / hard coded digests in the tests if it's not too much trouble.

tests/frontend/show.py
gtristan: Extra tip: in case you didn't find it yet... see for example `tests/frontend/buildcheckout.py`, this test uses the `create_artifact_share` utility which sets up a remote artifact cache (locally) which can be used to interact with in tests.

tests/frontend/show.py
Kekun: Nah it's trivial, I did it already (locally). :slightly_smiling_face:

tests/frontend/show.py
Kekun: The tests now use the `show_artifact_cas_digest_project` project, let me know if it's not correct.

src/buildstream/_frontend/widget.py
Kekun: I replaced the output for no artifact digest by an empty string and a warning message.

I didn't do the collection though.

tests/frontend/show.py
Kekun: I'm not sure how to test that an element is cached remotely before building it locally, so far I test it by deleting the local artifact. Any suggestion?

tests/frontend/show.py
```diff
@@ -438,8 +438,9 @@ def setup_test():
         ("%{deps}", "- import-dev.bst\n- import-links.bst\n- import-bin.bst"),
         ("%{build-deps}", "- import-dev.bst\n- import-links.bst"),
         ("%{runtime-deps}", "- import-links.bst\n- import-bin.bst"),
+        ("%{artifact-cas-digest}", "(no artifact CAS digest)"),
```
abderrahim: This now prints nothing.

```
E           AssertionError: Expected output:
E           (no artifact CAS digest)
E           Instead received output:
```

but maybe this whole change was supposed to be dropped now that we have a new file for these tests?

tests/frontend/show_artifact_cas_digest.py
```diff
@@ -0,0 +1,129 @@
...
+
+    # Configure a local cache
+    local_cache = os.path.join(str(tmpdir), "cache")
+    cli.configure({"cachedir": local_cache})
+
+    with create_artifact_share(os.path.join(str(tmpdir), "artifactshare")) as share:
+
+        cli.configure({"artifacts": {"servers": [{"url": share.repo}]}})
+
+        # Check the element and its dependencies have not been built locally and are not existing in the remote cache
```
abderrahim: This feels unnecessary, every time the test run it should start with a fresh directory.

tests/frontend/show_artifact_cas_digest.py
```diff
@@ -0,0 +1,129 @@
...
+        for component in sorted(expected_digests.keys()):
+            assert cli.get_element_state(project, component) == "buildable"
+            assert_not_shared(cli, share, project, component)
+
+        # Check the element and its dependencies have no artifact digest
+        result = cli.run(project=project, silent=True, args=["show", "--format", "%{name},%{artifact-cas-digest}", target])
+        result.assert_success()
+
+        digests = dict(line.split(",", 2) for line in result.output.splitlines())
```
abderrahim: I feel this is over engineered, you're testing a single element each time.

tests/frontend/show_artifact_cas_digest.py
```diff
@@ -0,0 +1,129 @@
...
+        }),
+        ("import-executable-files.bst", {
+            "import-executable-files.bst": "133a9ae2eda30945a363272ac14bb2c8a941770b5a37c2847c99934f2972ce4f/170",
+        }),
+        ("import-symlinks.bst", {
+            "import-symlinks.bst": "95947ea55021e26cec4fd4f9de90d2b7f4f7d803ccc91656b3e1f2c9923ddf19/131",
+        }),
+    ],
+)
+def test_show_artifact_cas_digest(cli, tmpdir, datafiles, target, expected_digests):
```
abderrahim: The whole test like this takes a very long time, more than 10s on my computer (per test). I wonder if it would be possible to split it such that each part is tested separately? We probably need @gtristan's guidance on this one.

tests/frontend/show_artifact_cas_digest.py
Kekun: It is because I carried it over from my previous test which also tested dependencies. :slightly_smiling_face: I'm considering adding dependencies to a test, just so we can check they were built and we can access their digest. Otherwise indeed it could and should be simplified.

tests/frontend/show.py
Kekun: Woops indeed, I only tested the new test and forgot to remove this, thanks.

tests/frontend/show_artifact_cas_digest.py
Kekun: Yeah I considered splitting it up too, but I don't know how to test that an element exists in the remote cache but hasn't been built locally without first pushing it and removing it locally, which isn't a really good way to test it in the first place anyway…

tests/frontend/show.py
gtristan: Deleting the local artifact is a correct thing to do, I'm not sure *why* you need to test if it is built remotely.

The tests always start with a clean slate, so if you are using the `create_artifact_share` utilities, it will be in a clean empty state at the beginning of each test, or you can explicitly delete from the artificial remote (there are some tools for that if you look into that utility toolbox).

man/bst-show.1
```diff
@@ -45,6 +45,7 @@ Show elements in the pipeline
         %{deps}           A list of all dependencies
         %{build-deps}     A list of build dependencies
         %{runtime-deps}   A list of runtime dependencies
+        %{artifact-cas-digest} The CAS digest of the built artifact
```
gtristan, line 49: This has already been changed in docstring of the `show()` command in `_frontend/cli.py`.

Updating the man pages is [documented here](https://docs.buildstream.build/master/hacking/writing_documentation.html#man-pages) and is done as a single action in a separate commit. (I know, its a bit of an obscure location and the contributing guide is a bit too much and comes from days when we had way too many contributors).

Also the man page generation will not work right now, this is fixed in the `tristan/sboms` branch in 0c2a92d2667f7f88938bea8779d50484ed6b6d94

Actually you may have run into this brokenness of man page generation and I'm ambivalent if this lands in advance of `tristan/sboms`, if this lands after `tristan/sboms`, it will be better for you to remove this change, follow the updated documentation, and add a "man pages update" commit to this branch.

tests/frontend/show_artifact_cas_digest.py
```diff
@@ -0,0 +1,291 @@
...
+
+    with create_artifact_share(os.path.join(str(tmpdir), "artifactshare")) as share:
+
+        cli.configure({"artifacts": {"servers": [{"url": share.repo}]}})
+
+        # Build the target locally
+        result = cli.run(project=project, silent=True, args=["build", target])
+        result.assert_success()
+
+        # Push the built artifacts to the remote cache
```
gtristan: We should not need an explicit `bst artifact push` command after `bst build`, as building should push the artifact all by itself.

Instead of this whole block, we can just add one more `assert_shared()` line above.

tests/frontend/show_artifact_cas_digest.py
```diff
@@ -0,0 +1,291 @@
...
```
gtristan: Same, we should not need an explicit `bst artifact push` here.

We can replace this block with an `assert_shared` statement in the build block success assertion above.

tests/frontend/show_artifact_cas_digest.py
```diff
@@ -0,0 +1,291 @@
...
+
+        cli.configure({"artifacts": {"servers": [{"url": share.repo}]}})
+
+        # Build the target locally
+        result = cli.run(project=project, silent=True, args=["build", target])
+        result.assert_success()
+
+        # Check the target has been built locally and is not existing in the remote cache
+        assert cli.get_element_state(project, target) == "cached"
+        assert_not_shared(cli, share, project, target)
```
gtristan: I don't understand why/how this can be true, did you do something explicitly to make sure that building would not push the artifact to the remote ?

tests/frontend/show_artifact_cas_digest.py
```diff
@@ -0,0 +1,291 @@
...
+)
+def test_show_artifact_cas_digest_uncached(cli, tmpdir, datafiles, target):
+    project = str(datafiles)
+    expected_no_digest = ""
+
+    # Configure a local cache
+    local_cache = os.path.join(str(tmpdir), "cache")
+    cli.configure({"cachedir": local_cache})
+
+    with create_artifact_share(os.path.join(str(tmpdir), "artifactshare")) as share:
```
gtristan: For this test, we are just testing something that is not cached locally and not cached remotely.

* We don't need an artifact share to simulate presence in a remote
* We don't need to test this on multiple, differing kinds of elements
  * The different kinds of elements are interesting to test only in a single test which checks all of the various digest values

tests/frontend/show_artifact_cas_digest.py
```diff
@@ -0,0 +1,291 @@
...
+    ],
+)
+def test_show_artifact_cas_digest_cached_locally(cli, tmpdir, datafiles, target, expected_digest):
+    project = str(datafiles)
+
+    # Configure a local cache
+    local_cache = os.path.join(str(tmpdir), "cache")
+    cli.configure({"cachedir": local_cache})
+
+    with create_artifact_share(os.path.join(str(tmpdir), "artifactshare")) as share:
```
gtristan: For testing things which *have not been cached remotely*, just don't use an artifact share.

tests/frontend/show_artifact_cas_digest.py
```diff
@@ -0,0 +1,291 @@
...
+        assert received_digest == "{target},{digest}".format(target=target, digest=expected_no_digest)
+
+
+# This tests that a target that has been built locally and
+# that isn't cached remotely has an artifact CAS digest.
+#
+# The test is performed with a remote cache.
+#
+@pytest.mark.datafiles(DATA_DIR)
+@pytest.mark.parametrize(
```
gtristan, line 109: Please specify the `ids` parameter for this parametrization, it will make the test output more legible.

E.g.:

```python
@pytest.mark.parametrize(

    "target, expected_digest",
    [
        ("import-basic-files.bst", "7093d3c89029932ce1518bd2192e1d3cf60fd88e356b39195d10b87b598c78f0/168"),
        ("import-executable-files.bst", "133a9ae2eda30945a363272ac14bb2c8a941770b5a37c2847c99934f2972ce4f/170"),
        ("import-symlinks.bst", "95947ea55021e26cec4fd4f9de90d2b7f4f7d803ccc91656b3e1f2c9923ddf19/131"),
    ],
    ids=["basic-files", "executable-files", "symlinks"]
)
```

tests/frontend/show_artifact_cas_digest.py
```diff
@@ -0,0 +1,291 @@
...
+
+# This tests that the target and its dependencies are built
+# as expected and that all their artifact CAS digests are
+# available.
+#
+# The test is performed without a remote cache.
+#
+@pytest.mark.datafiles(DATA_DIR)
+@pytest.mark.parametrize(
+    "target, expected_digests",
```
gtristan, line 110: Same comment regarding using the `ids` parameter in parametrized tests which have complicated/long parameters.

tests/frontend/show_artifact_cas_digest.py
```diff
@@ -0,0 +1,291 @@
...
+@pytest.mark.datafiles(DATA_DIR)
+@pytest.mark.parametrize(
+    "target, expected_digest",
+    [
+        ("import-basic-files.bst", "7093d3c89029932ce1518bd2192e1d3cf60fd88e356b39195d10b87b598c78f0/168"),
+        ("import-executable-files.bst", "133a9ae2eda30945a363272ac14bb2c8a941770b5a37c2847c99934f2972ce4f/170"),
+        ("import-symlinks.bst", "95947ea55021e26cec4fd4f9de90d2b7f4f7d803ccc91656b3e1f2c9923ddf19/131"),
+    ],
+)
+def test_show_artifact_cas_digest_cached_remotely(cli, tmpdir, datafiles, target, expected_digest):
```
gtristan: I think this is the same as `test_show_artifact_cas_digest_uncached`, we can drop this one.

tests/frontend/show.py
gtristan: Thanks !

src/buildstream/_frontend/widget.py
gtristan: Sure :)

tests/frontend/show.py
gtristan: This was addressed (and a bit too much :) ), closing this comment.

tests/frontend/show.py
gtristan: As mentioned in my new comments, let's take care to:

* Have one test which cares about what the digests are
* This test is parametrized (and uses the `ids` especially since it produces these super long test names)
* Not use parametrization for cases where the digest value itself doesn't matter.

This is just about only testing what is relevant to test, and keeping overall test time to a minimum.

tests/frontend/show.py
gtristan: I believe this is fixed, closing comment thread.

tests/frontend/show_artifact_cas_digest.py
gtristan: I feel ambivalent about this test, however I don't know that it is relevant, it seems irrelevant

* We are not testing the user's ability to parse `bst show` output here
* Are we testing that a cas digest of an element with dependencies has a specific digest ?
  * If so, we can add that single element  in the same test with the rest of the other elements which have potentially different cas digests
  * Note that the dependencies of an artifact does not affect its digest, because what we are interested in is *the digest of the **content** of the artifact*, not the digest of the overall artifact envelope.

Perhaps, it is relevant to test that artifacts which have the same *content* but differing *dependencies* have exactly the same cas digest ? if only as a means of validating that we are reporting the correct digest here ?

tests/frontend/show_artifact_cas_digest.py
gtristan: I've made a bunch of comments about what tests are overkill, a hand full of the artifact shares are not needed, and we are parametrizing tests in every tested case, even when we don't care about the value of the cas digest in the context of that test.

Please fix those and feel free to close this comment.

gtristan:
Overall, thanks for updating the tests, its a great update and sorry I didn't get back to it sooner.

There are some more comments which should be straight forward to address, and one comment which I find concerning, regarding whether we are actually reporting the *correct* digest of the content and not the digest of the artifact envelope: https://github.com/apache/buildstream/pull/1994#discussion_r2072625739

~~You will also need to apply https://github.com/apache/buildstream/pull/1997/commits/51227aca4085a0d8fcb28f6941f0c43ce2269de3 to this branch in order to pass CI, since upstream setuptools decided to break the whole world.~~ Scratch that, for this problem you will just need to rebase against buildstream master.

tests/frontend/show_artifact_cas_digest.py
Kekun: I'm not sure I understand how they are the same… I don't mind dropping it if you think it's better that way, but would you mind explaining your reasoning?

tests/frontend/show_artifact_cas_digest.py
Kekun: The goal of this test is to check the artifact is cached locally only, so I need a share to check the shared state of the cache. Unless I don't understand what you mean?

tests/frontend/show_artifact_cas_digest.py
Kekun: In my tests I don't remember the built artifacts to be automatically pushed to the cache, that's why I push it manually in other tests and why I check the artifacts aren't cached remotely in some other. Is there a parameter I should set to enable automatically sharing the artifacts?

tests/frontend/show_artifact_cas_digest.py
Kekun: If I don't explicitly push them, `assert_shared()` fails, that's why I'm explicitly pushing everywhere. Am I missing something?

tests/frontend/show_artifact_cas_digest.py
Kekun: Ditto, if I don't explicitly push, the artifacts aren't shared, at least when testing on my machine.

tests/frontend/show_artifact_cas_digest.py
Kekun: I wanted to check for different results when building building with *and* without a remote, do you think we shouldn't test that?

tests/frontend/show_artifact_cas_digest.py
Kekun: Fixed by adding `"push": True` as discussed in our meeting.

tests/frontend/show_artifact_cas_digest.py
Kekun: Fixed by adding `"push": True` as discussed in our meeting.

tests/frontend/show_artifact_cas_digest.py
Kekun: Fixed by adding `"push": True` as discussed in our meeting.

tests/frontend/show_artifact_cas_digest.py
gtristan: If we know the artifact is cached locally, then we are certain that we should be able to know the digest of the artifact, and whether or not artifact is also cached remotely is immaterial.

gtristan:
Interestingly, this is the first time `bst show` is requiring *artifact data* in order to work, the only tangentially similar thing is the `%{state}`, which has knowledge about *whether an artifact is cached* (or has build dependencies cached).

Asides from the testing we've already added in this PR around what happens with remote caches and locally deleted artifacts, this also raises the question of how this behaves with remote execution and options such as `storage-service` (https://docs.buildstream.build/master/using_config.html#cache-control).

It seems however fine that the digests will be empty unless a `bst artifact pull` command has completed for the artifact.

tests/frontend/show_artifact_cas_digest.py
Kekun: Fair enough, I'll remove it. :slightly_smiling_face:

Kekun: That being said, it's the only one where I actually test the expected digest for these elements, when you said "this one" do you mean I should drop `test_show_artifact_cas_digest_uncached` or `test_show_artifact_cas_digest_cached_remotely`?

gtristan: A test which does not bother with the extra time consuming task of setting up a remote cache, where we expect to observe the digests, is a test which makes sense to parametrize and test all of the different flavors of elements we're testing.

Kekun: Alright I got you, setting up a remote cache is expensive, we should minimize the number of tests setting one up and what we test with the remote.

Kekun:
I believe I addressed all your comments, please let me know if I missed something.

src/buildstream/_frontend/widget.py
```diff
@@ -461,6 +461,25 @@ def show_pipeline(self, dependencies, format_):
                 # Dump the SourceInfo provenance objects in yaml format
                 line = p.fmt_subst(line, "source-info", _yaml.roundtrip_dump_string(all_source_infos))
 
+            # Artifact CAS Digest
+            if "%{artifact-cas-digest" in format_:
+                artifact = element._get_artifact()
+                if not artifact.query_cache():
+                    artifact = None
+                if artifact is not None:
```
abderrahim: This is bad style, just check the actual condition directly

```suggestion
                if artifact.query_cache():
```

Better yet, since we have already queried the cache, we just need to check whether the artifact is cached
```suggestion
                if artifact.cached():
```

src/buildstream/_frontend/widget.py
```diff
@@ -461,6 +461,25 @@ def show_pipeline(self, dependencies, format_):
...
+                    artifact_files = artifact.get_files()
+                    # We call the private CasBasedDirectory._get_digest() for
+                    # the moment, we should make it public on Directory.
+                    artifact_digest = artifact_files._get_digest()
+                    formated_artifact_digest = "{}/{}".format(artifact_digest.hash, artifact_digest.size_bytes)
+                    line = p.fmt_subst(line, "artifact-cas-digest", formated_artifact_digest)
+                else:
+                    # FIXME
+                    # We could instead collect all of the elements that are not
+                    # cached and issue a single warning message.
```
abderrahim: I wouldn't worry about this. #1997 does the same thing of warning for every element.

tests/frontend/show_artifact_cas_digest.py
```diff
@@ -0,0 +1,225 @@
...
+# Pylint doesn't play well with fixtures and dependency injection from pytest
+# pylint: disable=redefined-outer-name
+
+import os
+import shutil
+
+import pytest
+
+from buildstream._testing import cli  # pylint: disable=unused-import
+from buildstream.exceptions import ErrorDomain
```
abderrahim: The linter is complaining about unused imports (shutils and ErrorDomain above) as well as trailing whitespace in this file.

gtristan:
Thanks for your relentless work on this !

PR #1997, opened 2025-03-04
Add new abstract methods for reporting source provenance

gtristan:
This is an initial implementation for the proposal up at: https://lists.apache.org/thread/q6gxjpld2vb1c9rqlsv24m12c087snc4

Some thoughts about this approach:

* This prioritizes machine readability and standardization of source provenance and version information

  Sources have a lot of freedom in how they implement things, and so we may very well need to expand on the types and constants added here, such as `SourceInfoMedium`, `SourceVersionType`, etc.

  The idea here is to have greater certainty about how sources are obtained, even if this cannot be covered by all currently existing source implementations (e.g. I didn't initially add a `bzr` medium for which we have a plugin, or a `cvs` medium for which we do not yet have a plugin).

  Aspirationally, forcing this data to be precise can allow adjacent tooling to do useful things.

* This drops the freeform "public data" mentioned in the proposal discussion

   My rationale for this choice in this branch, is that ultimately we want a data with a constant shape, and if for examlpe, we want the user to be able to override or assist a source with determining the reported *"version"*, then the `Source` implementation already has everything it needs to do so:
    * it can add additional configuration keys for the user to configure
    * it has the power to implement `collect_source_info()` however it wants.

* This does not yet attempt to cover the concept of *tracking* information

  I would like to consider this, but we should think carefully about how this can be useful. For instance, some git plugins have different interpretations of what their "tracking" strings mean, sometimes following a branch head, sometimes looking for the latest tag in history which matches a given regular expression.

  If we export this tracking information, it should probably be useful for external tooling to figure out how to do the tracking and come to the same conclusion, otherwise it is unclear what this is useful for.

* This does not cover the CVE information

   While the SourceInfo objects representing a source's provenance is a list, I believe that the CVE information continues to be a per-element concept.

  For example, when we have applied security patches to a module, those security patches are, themselves, sources, with provenance of being revisioned in the local project

### DIFF
--- a/src/buildstream/_frontend/widget.py
+++ b/src/buildstream/_frontend/widget.py
@@ -23,6 +23,7 @@ import click
 
 from .profile import Profile
 from ..types import _Scope
+from ..source import SourceImplError
 from .. import _yaml
 from .. import __version__ as bst_version
 from .. import FileType
@@ -436,6 +437,29 @@ class LogLine(Widget):
             if "%{runtime-deps" in format_:
                 runtime_deps = [e._get_full_name() for e in element._dependencies(_Scope.RUN, recurse=False)]
                 line = p.fmt_subst(line, "runtime-deps", _yaml.roundtrip_dump_string(runtime_deps).rstrip("\n"))
+
+            # Source Information
+            if "%{source-info" in format_:
+
+                # Get all the SourceInfo objects
+                #
+                all_source_infos = []
+                for source in element.sources():
+                    try:
+                        source_infos = source.collect_source_info()
+                    except SourceImplError as e:
+                        source.warn(str(e))
+                        continue
+
+                    serialized_sources = []
+                    for s in source_infos:
+                        serialized = s._serialize()
+                        serialized_sources.append(serialized)
+
+                    all_source_infos += serialized_sources
+
+                # Dump the SourceInfo provenance objects in yaml format
+                line = p.fmt_subst(line, "source-info", _yaml.roundtrip_dump_string(all_source_infos))
 
             report += line + "\n"
 
--- a/src/buildstream/downloadablefilesource.py
+++ b/src/buildstream/downloadablefilesource.py
@@ -25,6 +25,48 @@ Any derived classes must write their own stage() and get_unique_key()
 implementation.
 
 
+.. _core_downloadable_source_builtins:
+
+Built-in functionality
+----------------------
+The DownloadableFileSource class provides built in keys which can be set when
+intantiating any Source which derives from DownloadableFileSource
+
+* Guess pattern
+
+  The ``version-guess-pattern`` sets the regular expression which will be used to attempt
+  to guess the version of a source when parsing the source's URI.
+
+  The DownloadableFileSource provides a default implementation of
+  :func:`Source.collect_source_info() <buildstream.source.Source.collect_source_info>`,
+  which will use the ``version-guess-pattern`` to attempt to extract a human readable
+  version string from the specified URI, in order to fill out the reported
+  :attr:`~buildstream.source.SourceInfo.version_guess`.
+
+  The URI will be *searched* using this regular expression, and is allowed to
+  yield a number of *groups*. For example the value ``(\\d+)_(\\d+)_(\\d+)`` would
+  report 3 *groups* if 3 numerical values separated by underscores were found in
+  the URI.
+
+  The default value for ``version-guess-pattern`` is ``\\d+\\.\\d+(?:\\.\\d+)?``.
+
+  .. note:
+
+     The version guessing mechanism will not be observed if ``version`` is specified.
+
+  **Since: 2.5**.
+
+* Version
+
+  The ``version`` explicitly sets the :attr:`~buildstream.source.SourceInfo.version_guess`
+  attribute of the :class:`SourceInfo <buildstream.source.SourceInfo>` reported for this
+  source, overriding any guessing.
+
+  This is useful for remote files which do not express their version in their filenames.
+
+  **Since: 2.5**.
+
+
 SourceMirror extra data "http-auth"
 --------------------------------------------
 The DownloadableFileSource, and consequently any :class:`Source <buildstream.source.Source>`
@@ -81,17 +123,38 @@ DownloadableFileSource will add the following header to the ``GET`` request to d
 
    Authorization: Bearer 1234
 
+
+.. _core_downloadable_source_info:
+
+Default reporting of :class:`.SourceInfo`
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Source plugins which derive from the DownloadableFileSource, unless overridden
+and specified in the documentation of the specific source plugin, will behave
+as described here.
+
+The DownloadableFileSource reports the URL of the remote file as the *url*.
+
+Further, the DownloadableFileSourcebzr source reports the
+:attr:`SourceInfoMedium.REMOTE_FILE <buildstream.source.SourceInfoMedium.REMOTE_FILE>` *medium* and the
+:attr:`SourceVersionType.SHA256 <buildstream.source.SourceVersionType.SHA256>` *version_type*,
+for which it reports the sha256 checksum of the remote file content as the *version*.
+
+An attempt to guess the version based on the remote filename will be made
+for the reporting of the *guess_version*. Control over how the guess is made
+or overridden is explained above in the
+:ref:`built-in functionality documentation <core_downloadable_source_builtins>`.
 """
 
 
 import os
+import re
 import urllib.request
 import urllib.error
 import contextlib
 import shutil
 import netrc
 
-from .source import Source, SourceError
+from .source import Source, SourceError, SourceInfoMedium, SourceVersionType
 from . import utils
 
 
@@ -202,9 +265,10 @@ def _download_file(opener_creator, url, etag, directory, bearer_auth):
 class DownloadableFileSource(Source):
     # pylint: disable=attribute-defined-outside-init
 
-    COMMON_CONFIG_KEYS = Source.COMMON_CONFIG_KEYS + ["url", "ref"]
+    COMMON_CONFIG_KEYS = Source.COMMON_CONFIG_KEYS + ["url", "ref", "version-guess-pattern", "version"]
 
     __default_mirror_file = None
+    __default_guess_pattern = re.compile(r"\d+\.\d+(?:\.\d+)?")
 
     def configure(self, node):
         self.original_url = node.get_str("url")
@@ -216,11 +280,28 @@ class DownloadableFileSource(Source):
 
         self._mirror_dir = os.path.join(self.get_mirror_directory(), utils.url_directory_name(self.original_url))
 
+        self._guess_pattern_string = node.get_str("version-guess-pattern", None)
+        if self._guess_pattern_string is None:
+            self._guess_pattern = self.__default_guess_pattern
+        else:
+            self._guess_pattern = re.compile(self._guess_pattern_string)
+
+        self._version = node.get_str("version", None)
+
     def preflight(self):
         return
 
     def get_unique_key(self):
-        return [self.original_url, self.ref]
+        unique_key = [self.original_url, self.ref]
+
+        # Backwards compatible method of supporting configuration
+        # attributes which affect SourceInfo generation.
+        if self._version is not None:
+            unique_key.append(self._version)
+        elif self._guess_pattern is not self.__default_guess_pattern:
+            unique_key.append(self._guess_pattern_string)
+
+        return unique_key
 
     def is_cached(self) -> bool:
         return os.path.isfile(self._get_mirror_file())
@@ -269,6 +350,24 @@ class DownloadableFileSource(Source):
             raise SourceError(
                 "File downloaded from {} has sha256sum '{}', not '{}'!".format(self.url, sha256, self.ref)
             )
+
+    def collect_source_info(self):
+        if self._version is None:
+            version_match = self._guess_pattern.search(self.original_url)
+            if not version_match:
+                version_guess = None
+            elif self._guess_pattern.groups == 0:
+                version_guess = version_match.group(0)
+            else:
+                version_guess = ".".join(version_match.groups())
+        else:
+            version_guess = self._version
+
+        return [
+            self.create_source_info(
+                self.url, SourceInfoMedium.REMOTE_FILE, SourceVersionType.SHA256, self.ref, version_guess=version_guess
+            )
+        ]
 
     def _get_etag(self, ref):
         etagfilename = os.path.join(self._mirror_dir, "{}.etag".format(ref))
--- a/src/buildstream/plugins/sources/local.py
+++ b/src/buildstream/plugins/sources/local.py
@@ -31,10 +31,23 @@ local - stage local files and directories
 
 See :ref:`built-in functionality doumentation <core_source_builtins>` for
 details on common configuration options for sources.
+
+
+Reporting :class:`.SourceInfo`
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+The local source reports the project relative path of the file or directory as the *url*.
+
+Further, the local source reports the
+:attr:`SourceInfoMedium.LOCAL <buildstream.source.SourceInfoMedium.LOCAL>` *medium* and the
+:attr:`SourceVersionType.CAS_DIGEST <buildstream.source.SourceVersionType.CAS_DIGEST>` *version_type*,
+for which it reports the CAS digest of the local source as the *version*.
+
+The *guess_version* of a local source is meaningless, as it is tied instead to
+the BuildStream project in which it is contained.
 """
 
 import os
-from buildstream import Source, SourceError, Directory
+from buildstream import Source, SourceError, SourceInfoMedium, SourceVersionType, Directory
 
 
 class LocalSource(Source):
@@ -67,11 +80,7 @@ class LocalSource(Source):
         # * Do the regular staging activity into the Directory
         # * Use the hash of the cached digest as the unique key
         #
-        if not self.__digest:
-            with self._cache_directory() as directory:
-                self.__do_stage(directory)
-                self.__digest = directory._get_digest()
-
+        self.__ensure_digest()
         return self.__digest.hash
 
     # We dont have a ref, we're a local file...
@@ -110,11 +119,24 @@ class LocalSource(Source):
         #
         self.__do_stage(directory)
 
+    def collect_source_info(self):
+        self.__ensure_digest()
+        version = "{}/{}".format(self.__digest.hash, self.__digest.size_bytes)
+        return [self.create_source_info(self.path, SourceInfoMedium.LOCAL, SourceVersionType.CAS_DIGEST, version)]
+
     # As a core element, we speed up some scenarios when this is used for
     # a junction, by providing the local path to this content directly.
     #
     def _get_local_path(self):
         return self.fullpath
+
+    # Ensure that the digest is resolved
+    #
+    def __ensure_digest(self):
+        if not self.__digest:
+            with self._cache_directory() as directory:
+                self.__do_stage(directory)
+                self.__digest = directory._get_digest()
 
     # Staging is implemented internally, we preemptively put it in the CAS
     # as a side effect of resolving the cache key, at stage time we just
--- a/src/buildstream/plugins/sources/workspace.py
+++ b/src/buildstream/plugins/sources/workspace.py
@@ -21,20 +21,32 @@ workspace - stage an opened workspace directory
 
 The workspace plugin must not be directly used. This plugin is used as the
 kind for a synthetic node representing the sources of an element with an open
-workspace. The node constructed would be specified as follows:
+workspace.
 
-.. code:: yaml
 
-   # Specify the workspace source kind
-   kind: workspace
+Reporting :class:`.SourceInfo`
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+The workspace source reports the project relative or absolute path to the open
+workspace as the *url*.
 
-   # Specify the absolute path to the directory
-   path: /path/to/workspace
+Further, the workspace source reports the
+:attr:`SourceInfoMedium.WORKSPACE <buildstream.source.SourceInfoMedium.WORKSPACE>` *medium* and the
+:attr:`SourceVersionType.CAS_DIGEST <buildstream.source.SourceVersionType.CAS_DIGEST>` *version_type*,
+for which it reports the CAS digest of the workspace source as the *version*.
+
+The *guess_version* of a workspace source is meaningless and omitted.
+
+.. attention::
+
+   Observing a SourceInfo with the ``SourceInfoMedium.WORKSPACE`` in the output of
+   :ref:`bst show --format %{source-info} <invoking_show>` is most likely undesirable, given
+   that you are likely interested in observing the source provenance information of the
+   project in a clean state rather than in a state with open workspaces.
 """
 
 import os
 
-from buildstream import Source, SourceError, Directory, MappingNode
+from buildstream import Source, SourceError, SourceInfoMedium, SourceVersionType, Directory, MappingNode
 from buildstream.types import SourceRef
 
 
@@ -71,11 +83,7 @@ class WorkspaceSource(Source):
         # * Do the regular staging activity into the Directory
         # * Use the hash of the cached digest as the unique key
         #
-        if not self.__digest:
-            with self._cache_directory() as directory:
-                self.__do_stage(directory)
-                self.__digest = directory._get_digest()
-
+        self.__ensure_digest()
         return self.__digest.hash
 
     def get_ref(self) -> None:
@@ -108,11 +116,24 @@ class WorkspaceSource(Source):
         with self._cache_directory(digest=self.__digest) as cached_directory:
             directory._import_files_internal(cached_directory, collect_result=False)
 
+    def collect_source_info(self):
+        self.__ensure_digest()
+        version = "{}/{}".format(self.__digest.hash, self.__digest.size_bytes)
+        return [self.create_source_info(self.path, SourceInfoMedium.WORKSPACE, SourceVersionType.CAS_DIGEST, version)]
+
     # As a core element, we speed up some scenarios when this is used for
     # a junction, by providing the local path to this content directly.
     #
     def _get_local_path(self) -> str:
         return self.path
+
+    # Ensure that the digest is resolved
+    #
+    def __ensure_digest(self):
+        if not self.__digest:
+            with self._cache_directory() as directory:
+                self.__do_stage(directory)
+                self.__digest = directory._get_digest()
 
     # Staging is implemented internally, we preemptively put it in the CAS
     # as a side effect of resolving the cache key, at stage time we just
--- a/src/buildstream/source.py
+++ b/src/buildstream/source.py
@@ -262,6 +262,12 @@ class SourceError(BstError):
 
 @dataclass
 class AliasSubstitution:
+    """AliasSubstitution()
+    An opaque data structure which may be passed through
+    :func:`SourceFetcher.fetch() <buildstream.source.SourceFetcher.fetch>` and in such cases
+    must be provided to :func:`Source.translate_url() <buildstream.source.Source.translate_url>`.
+    """
+
     _effective_alias: str
     _mirror: Union[SourceMirror, str]
 
@@ -293,8 +299,9 @@ class SourceFetcher:
 
         Args:
            alias_override: The alias to use instead of the default one
-               defined by the :ref:`aliases <project_source_aliases>` field
-               in the project's config.
+               defined by the :ref:`aliases <project_source_aliases>` field in the
+               project's config. If provided, it must be used when calling
+               :func:`Source.translate_url() <buildstream.source.Source.translate_url>`.
 
         Raises:
            :class:`.SourceError`
--- a/src/buildstream/source.py
+++ b/src/buildstream/source.py
@@ -21,9 +21,8 @@ Source - Base source class
 
 Built-in functionality
 ----------------------
-
-The Source base class provides built in functionality that may be overridden
-by individual plugins.
+The Source base class provides built in keys which can be set when instantiating
+any Source.
 
 * Directory
 
@@ -101,6 +100,15 @@ these methods are mandatory to implement.
 
   **Optional**: This is completely optional and will do nothing if left unimplemented.
 
+* :func:`Source.collect_source_info() <buildstream.source.Source.collect_source_info>`
+
+  Collect SourceInfo objects to describe the provenance of sources.
+
+  **Optional**: BuildStream will function correctly if this is unimplemented, but the
+  ability to generate SBoMs will be impaired, it is highly recommented to implement this.
+
+  See: :ref:`documentation on generating SourceInfo <core_source_info>`.
+
 
 .. _core_source_ref:
 
@@ -163,6 +171,10 @@ configuration which may effect how the source is :func:`staged <buildstream.sour
 as well as any configuration which uniquely identifies the source, which of course
 includes the :attr:`~buildstream.types.SourceRef`.
 
+When plugins :ref:`generate SourceInfo <core_source_info>`, it is also
+important that any configuration attributes which contribute to the
+generation of SourceInfo also be included in the unique key.
+
 
 Accessing previous sources
 --------------------------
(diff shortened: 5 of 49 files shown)


src/buildstream/downloadablefilesource.py
```diff
@@ -270,6 +270,13 @@ def fetch(self):  # pylint: disable=arguments-differ
                 "File downloaded from {} has sha256sum '{}', not '{}'!".format(self.url, sha256, self.ref)
             )
 
+    def collect_source_info(self):
+        #
+        # XXX remote sources are not necessarily archives, perhaps we should
+        # allow downloadablefilesource imlementations to choose the SourceInfoMedium
+        #
+        return [SourceInfo(self.url, SourceInfoMedium.ARCHIVE, SourceVersionType.SHA256, self.ref)]
```
juergbi: `self.url` is a fully qualified URL after alias translation. The unique key of a source (and thus, the cache key of an element) only covers the aliased URL, though. I.e., `collect_source_info()` may return different fully qualified URLs (including URLs of internal mirrors) for builds with the same cache key. I'm not sure how to solve this but it seems like a potential issue. Or am I misreading the code?

This issue is not specific to `DownloadableFileSource`, of course.

src/buildstream/source.py
```diff
@@ -262,10 +262,123 @@ def __init__(
...
+        """
+        The url of the source input
+        """
+
+        self.medium: str = medium
+        """
+        The :class:`.SourceInfoMedium` of the source input
+        """
+
+        self.version_type: str = version_type
```
juergbi: A single source URL may have both an upstream version and e.g. a commit or hash. Do we want/need to support this? Or is the idea to handle this with multiple `SourceInfo` objects pointing to the same URL?

src/buildstream/source.py
```diff
@@ -262,10 +262,123 @@ def __init__(
...
+    must be provided to :func:`Source.translate_url() <buildstream.source.Source.translate_url>`.
+    """
+
     _effective_alias: str
     _mirror: Union[SourceMirror, str]
 
 
+class SourceInfoMedium(FastEnum):
+    """
+    Indicates the meduim in which the source is obtained
```
juergbi:
```suggestion
    Indicates the medium in which the source is obtained
```

src/buildstream/source.py
```diff
@@ -262,10 +262,123 @@ def __init__(
...
+
+    *Since: 2.5*
+    """
+
+    LOCAL = "local"
+    """
+    Files stored locally in the project
+    """
+
+    ARCHIVE = "archive"
```
juergbi: `LOCAL` vs. `ARCHIVE` seem like orthogonal attributes to me. You can have a local archive in your project and you can fetch a single e.g. config file from a remote server, at least theoretically. Maybe `LOCAL` should rather be a flag than part of this enum? (Or maybe even make it implicit as the corresponding URL will have no host component).

I'm also not sure whether we should have one enum entry for each archive file format (e.g. `TAR` and `ZIP`) or whether a generic `ARCHIVE` makes more sense.

src/buildstream/downloadablefilesource.py
gtristan: So the `ref` for a `DownloadableFileSource`, if not overridden, is the checksum of the downloaded archive (or "file"), so it is correct for `SourceVersionType.SHA256`.

The `self.url` is a fully qualified URL, my intention here, ideally, would be to guarantee that it is the original fully qualified URL after the initial project aliasing, and guarantee that no mirrors are involved.

The only time that a `self.url` would **not** be the original qualified URL of where the source *originally came from*, is if `Source.collect_source_info()` were to be called by the plugin inside `Source.fetch()` or `Source.track()`, possibly we can make this illegal with an assertion.

I would further like to protect against any *"original URL"* mangling would could result from junctioning and alias mappings, but I'm pretty sure that the alias mapping feature used for mirror plugins does not affect what the original upstream URLs would be.

If the project(s) are written in such a way to not ever write down the real provenance of upstream sources, of course we would get those URLs.

Does this make sense ?

src/buildstream/source.py
gtristan: This is a good question.

My initial thoughts were that some plugins, e.g. some of the git plugins out there in the wild, provide `git describe` format for their `ref`, and thus would provide both contexts in the same string, but always something useful and precise (e.g. merely a tag, or a tag + number of child commits + commit sha).

Another thought I have been playing with along these similar lines is that it may be possible for `Source` classes to require `Source.fetch()` be run and that they have access to the staged sources in order to implement `Source.collect_source_info()`.

Also, we can consider formatting of reports separately, as long as we are strict about what the version type *means*, so we could potentially have a version type which is a *"git sha or git describe"* output, and parse these strings into more fancy reports when generating html and such, if we want to *display* the version and sha separately.

That said, indeed I wonder how one could possibly determine the *version* of a tarball... and in that light I wonder how meaningful the version of a tarball is.

src/buildstream/source.py
gtristan: Interesting, I consider the *medium* type as the *means in which the data is stored*, and not what the downloaded data *actually is*.

I.e. the medium here explains that it comes from the project itself, while a git medium used a `git` protocol to download, and `archive` downloaded a file.

I rather lean the other way here and think that `ARCHIVE` may be renamed to `REMOTE_FILE` or such, and this would make sense for any `DownloadableFileSource`.

When considering this, I'm trying  to balance:
* What is useful for interoperability with external tooling
* What is needed to logically describe the provenance of the source, for reporting purposes

nanonyme:
We have been having the challenge with bst-to-lorry that we're missing a way for source to explicitly state relevant refspecs for git. Heuristically computed once are commonly incomplete or even wrong. So if there's API, it really should take this into account. The problem is that commonly in git repositories there are branches that need to be ignored (like pull request branches) since they are routinely force pushed to. Also with git_module it's common that upstream doesn't declare at all what the branch is from which commit originated.

src/buildstream/source.py
abderrahim: In general, I feel that it's way easier to determine the version of a tarball than it is to extract the version from a git tag. The tarball naming is just more consistent.

I understand the value of not having plugins care about these things and leave the processing to whatever uses this info, but at the same time it's useful to have plugins care about it because they sometimes have more info available.

> Or is the idea to handle this with multiple SourceInfo objects pointing to the same URL?

The issue with this is that you can't tell if they're referring to the same thing if e.g. you have a git repo and one SourceInfo refers to a tag and the other refers to a commit.

gtristan: > In general, I feel that it's way easier to determine the version of a tarball than it is to extract the version from a git tag. The tarball naming is just more consistent.

On the face of it, this *appears* to be true.

Here is my thoughts about trust, though - with a git repo, I have a higher level of *trust* that a given release tag means that the content I am getting *is that actual release*.

With a tarball on the other hand, I get an sha256sum and I get a filename. That filename has numbers in it.

Yes it's true that by convention, for a staggering majority of tarballs out there in the wild, the number in the filename is somehow related to the *version* of software you'll get in the tarball, for a CLI tool you might see that number when you type `cli --version`.

In order to *produce version information* that is reliable, I'm not sure it can be completely automated, perhaps at least a human being, when adding the `.bst` files to a project or when updating those, needs to at least sign off on the *way that we extract the version*.

What I'm getting at, is that overall, I'm not comfortable with any statement like:

> *"This plugin knows how to get the version of this upstream data blob"*

We just don't *really* know what the version is at any time, not without at least looking at the upstream project, reading about its release policy and figuring it out.

shymega:
In terms of implementing this into plugins that currently use the internal API, would someone be able to write up an example plugin that uses the new API - I'm a little lost on the implementation semantics...

src/buildstream/source.py
nanonyme: With how we operate now, the person who writes the element that has the source is able to write regexp to parse version from tarball name/ref. Is that possible with this new API?

nanonyme: This seems like a hard part of the problem to me. This is currently easily overridable through public data. If that cannot be done with this new API, then it feels like a step backwards.

nanonyme: Yes. Right. So I get the problem now. The API I was proposing *did not* specifically give version because that cannot be provided in any useful form. Instead it was providing git ref (in commit or git-describe format) or tar source and letting consumer provide a way to process this into version through out-of-band public data. This is *exactly* because, as you said, mechanisms to read version varies from project to project. IMHO us putting a default matcher to begin with was a mistake, we should have just forced element author to add an explicit matcher whenever there was intent to have version for a source. This would have allowed making it a fatal error always if the match fails.

nanonyme: The problem is that even though git release tags may have release version in it, the specific formattings vary wildly. Extracting version in a way that you manage to extract the wanted version and nothing else than the wanted version is harder than solving Halting problem.

src/buildstream/source.py
gtristan: I've now renamed this to be `REMOTE_FILE` which is used by default by `DownloadableFileSource`, and is, in my opinion factually correct for any remotely downloaded file, including possibly compressed archives.

src/buildstream/source.py
gtristan: I've updated the patches now to make a compromise here.

* We keep the *specific* versioning, which can be an sha256sum or a git hash or anything which can *accurately* describe a unique source version.

* We add a `SourceInfo.version_guess` member to the data type. This is implemented by plugins who implement it, for the `DownloadableFileSource`, there is a default implementation which does the regex thing found in the collect_manifest code, and an added `version-guess-pattern` configure key allowing the user to specify the regex which should be used in order to make a guess in determining the version.

I think essentially this gives us the same functionality, however I'm just more comfortable with calling this a *guess*, and avoid any  misinterpretation of the promises BuildStream is making here.

Essentially, really *knowing* a version requires some human understanding of the upstream project, what their release policy is, and what, if anything, their release tags or git branches mean in relation to that.

gtristan:
I've been testing this locally with freedesktop-sdk master and the sister merge requests applied:

* https://gitlab.com/BuildStream/buildstream-plugins-community/-/merge_requests/364
* https://github.com/apache/buildstream-plugins/pull/870

And running:

```sh
bst show --format $'%{name}\n%{source-info}'  platform.bst
```

There are still more plugins to implement, notably we can have version guessing on the `git_tag` or similar community plugins which encode more context than just the commit sha into the `ref`.

The `cargo2` plugin works, but *only* if you correct `gstreamer-plugins-rs.bst` to set the `git-mirrors` parameter (that whole thing looks like it needs support from buildstream to work with the project aliases automatically... there is probably an issue open about that).

gtristan:
> In terms of implementing this into plugins that currently use the internal API, would someone be able to write up an example plugin that uses the new API - I'm a little lost on the implementation semantics...

See the core plugins in this PR (separate commits), or the changes in the sister PRs
* https://gitlab.com/BuildStream/buildstream-plugins-community/-/merge_requests/364
* https://github.com/apache/buildstream-plugins/pull/87

src/buildstream/downloadablefilesource.py
juergbi: I forgot that `self.url` won't be a mirror URL outside `Source.fetch()` and `Source.track()`, so it should indeed be less of an issue. However, even the original URL is not part of the cache key, so changing the alias definition would still result in `collect_source_info()` without a cache key change. Or am I missing something else? Maybe this is acceptable but we should at least be aware of this.

> I would further like to protect against any "original URL" mangling would could result from junctioning and alias mappings, but I'm pretty sure that the alias mapping feature used for mirror plugins does not affect what the original upstream URLs would be.

I'm not sure whether that's the case. Won't the alias definition from the parent project be used if the alias from the subproject is mapped in the junction? That alias definition could potentially point to a different URL without affecting the cache key.

src/buildstream/downloadablefilesource.py
```diff
@@ -25,6 +25,46 @@
...
+  The URI will be *searched* using this regular expression, and is allowed to
+  yield a number of *groups*. For example the value ``(\\d+)_(\\d+)_(\\d+)`` would
+  report 3 *groups* if 3 numerical values separated by underscores were found in
+  the URI.
+
+  The default value for ``version-guess-pattern`` is ``\\d+\\.\\d+(?:\\.\\d+)?``.
+
+  **Since: 2.5**.
+
+Built-in functionality
```
abderrahim: I think this section doesn't belong here (or at least it should be merged with the previous, similarly named one)

src/buildstream/downloadablefilesource.py
```diff
@@ -270,6 +318,21 @@ def fetch(self):  # pylint: disable=arguments-differ
...
             )
 
+    def collect_source_info(self):
+        version_match = self._guess_pattern.search(self.original_url)
+        if not version_match:
+            version_guess = None
+        elif self._guess_pattern.groups == 0:
+            version_guess = version_match.group(0)
+        else:
+            version_guess = ".".join(version_match.groups())
```
abderrahim: Maybe we could have this version guessing logic in a function in utils?

src/buildstream/plugins/sources/local.py
```diff
@@ -110,6 +110,9 @@ def init_workspace_directory(self, directory):
         #
         self.__do_stage(directory)
 
+    def collect_source_info(self):
+        return [SourceInfo(self.path, SourceInfoMedium.LOCAL, SourceVersionType.DIGEST, self.get_unique_key())]
```
abderrahim: `get_unique_key()` only uses the hash of the digest, which isn't enough to get the content from CAS.

Buildbox uses the format `{hash}/{size}`, which is what we're planning to use in #1994 and #1991. We should use it here too. This would allow downloading the source from CAS using e.g. `casdownload` from buildbox

src/buildstream/plugins/sources/workspace.py
```diff
@@ -108,6 +108,9 @@ def stage_directory(self, directory):
         with self._cache_directory(digest=self.__digest) as cached_directory:
             directory._import_files_internal(cached_directory, collect_result=False)
 
+    def collect_source_info(self):
+        return [SourceInfo(self.path, SourceInfoMedium.WORKSPACE, SourceVersionType.DIGEST, self.get_unique_key())]
```
abderrahim: same comment as local

src/buildstream/source.py
```diff
@@ -262,10 +261,147 @@ def __init__(
...
+    """
+    Indicates the type of the version string
+
+    *Since: 2.5*
+    """
+
+    COMMIT = "commit"
+    """
+    A commit string which accurately represents a version in a source
+    code repository or VCS
```
abderrahim, line 476: Should we comment something about git specifically? In particular, our plugins can use either the commit sha1, or a the long version of the git describe.

I think the current description fits both, which I think is a good thing. But maybe it's better to give an example.

src/buildstream/source.py
```diff
@@ -262,10 +261,147 @@ def __init__(
...
+    """
+
+    SHA256 = "sha256"
+    """
+    An sha256 checksum
+    """
+
+    DIGEST = "digest"
+    """
+    A CAS digest representing the unique version of this source input
```
abderrahim: We could be more explicit about the `{hash}/{size}` format?

src/buildstream/source.py
```diff
@@ -684,6 +821,19 @@ def is_cached(self) -> bool:
...
+
+        This method is guaranteed to only be called whenever
+        :func:`Source.is_resolved() <buildstream.source.Source.is_resolved>`
+        returns `True`.
+
+        Returns: the :class:`.SourceInfo` objects describing this source
+
+        *Since: 2.5*
+        """
+        raise ImplError("Source plugin '{}' does not implement collect_source_info()".format(self.get_kind()))
```
abderrahim: Could we have this collect source info from source fetchers by default? Any plugin that uses source fetchers will definitely want to implement source info in source fetchers.

Of course, this is more for standardisation than anything, one can always implement it like you did in https://github.com/apache/buildstream-plugins/pull/87

src/buildstream/_frontend/widget.py
```diff
@@ -437,6 +437,23 @@ def show_pipeline(self, dependencies, format_):
                 runtime_deps = [e._get_full_name() for e in element._dependencies(_Scope.RUN, recurse=False)]
                 line = p.fmt_subst(line, "runtime-deps", _yaml.roundtrip_dump_string(runtime_deps).rstrip("\n"))
 
+            # Source Information
+            if "%{source-info" in format_:
+
+                # Get all the SourceInfo objects
+                #
+                all_source_infos = []
+                for source in element.sources():
```
abderrahim, line 447: A was playing with this a little, and came up with this:
```suggestion
                source_element = element._get_source_element()
                for source in source_element.sources():
```

given `Element._get_source_element()` gives the element that has the sources for this element, it really made sense. It gives a nice solution to the issue of actual sources of filter elements not being accessible.

I'm thinking we can push it further and move this whole loop into an `Element.collect_source_info()` method. This way, something like the `collect_manifest` plugin wouldn't need to call `Element.sources()` at all (and we can deprecate it). It would be able to call `_get_source_element()` as needed, and users won't have to worry about it.

Moving forward with this idea, we could even have the `compose` plugin collect sources from all build dependencies that contribute files to the final image. Though this might need some more thinking.

abderrahim (commented):
I'm starting to like this, I left a few comments on how I think we can improve it.

shymega:
> [apache/buildstream-plugins/pull/870](https://github.com/apache/buildstream-plugins/pull/870)

@gtristan This should be https://github.com/apache/buildstream-plugins/pull/87, not https://github.com/apache/buildstream-plugins/pull/870.

src/buildstream/source.py
gtristan: > Should we comment something about git specifically? In particular, our plugins can use either the commit sha1, or a the long version of the git describe.
> 
> I think the current description fits both, which I think is a good thing. But maybe it's better to give an example.

Interesting. I think that for the git plugins which do this, we can either include the whole git describe string or we can have the plugins parse out the commit sha only, while parsing out other things into the *guess_version*.

I think that *if* the git describe string cannot be directly used in every way a commit sha can be used, it would be preferable to be more strict here and have the plugins parse out the pure commit sha from the ref and report that for the version of type `SourceVersionType.COMMIT`.

src/buildstream/source.py
gtristan: > We could be more explicit about the {hash}/{size} format?

So I'm hoping to minimize how many types we put here, and I'm thinking along the lines of:

* If its a `SourceInfoMedium.LOCAL`, then a `SourceVersionType.DIGEST` is a CAS digest, which contains the necessary info (as you pointed out it should be `{hash}/{size}`.
* If it's a `SourceInfoMedium.OCI_REGISTRY`, it just has the digest hash

Otherwise, we need to have a lot of explicit types.

On the other hand, as you point out... this makes it difficult to clearly document the meaning of things... so perhaps I need a different `DIGEST` type for OCI source plugins like the `docker` source ?

src/buildstream/source.py
gtristan: This seems sensible, I'll add something to the `SourceFetcher` API for that.

src/buildstream/source.py
abderrahim: Wouldn't `SHA256` (which we also have) be more appropriate for an OCI registry? `DIGEST` explicitly says "A CAS digest", so it wouldn't be applicable to an OCI registry.

gtristan:
@abderrahim, @juergbi 

Some additional thoughts now that I get down to business...

## Abolish the whole freeform string entirely

In our meeting this week, we discussed that we would prefer to keep the flexibility and not require that `SourceInfoMedium` and `SourceVersionType` enumeration values be strictly followed (resulting in the requirement of many distinct values being introduced into the core)...

**And** we discussed that we would instead prefer freeform strings for these attributes of the `SourceInfo` object, and that we should:
* Automatically encode the plugin *"kind"* name into the SourceInfo reported
* Provide a machine readable way (e.g. via `bst show`) to provide the plugin provenance information for given plugin kinds
* Have the deciding information about the *meaning* of the `SourceInfo.version` string (e.g. is it a git sha or is it a sha256sum of a downloaded file or what), documented in the governing source plugin documentation individually.

Ok, this all makes sense, however now that I look at it again, I no longer see any need whatsoever of having the `SourceInfoMedium` or `SourceVersionType` information present **at all** in the `SourceInfo` object.

Instead we only need the following members:
* `kind`: Plugin "kind" string
* `url`: Unaliased upstream full URL (not mirrored URL)
* `version`: The machine readable *precise* version (e.g. git sha, CAS/docker digest, sha256sum, etc)
* `version_guess`: The plugin assisted *guess* at what the human readable version is

Given that we can obtain the precise plugin and version for a given `kind`, and that that governing plugin documentation **must** already document precisely *what it means* in the `version` member, the whole concept of the `medium` and `version_type` are obsolete and essentially meaningless.

## Automate the plugin specific documentation

As a separate thought, I feel like the need to associate the governing documentation of *what a `version` means* with a given *plugin `kind`* is overly onerous.

I think that while this can continue to be true, we could make this easier in the following way:

* Source plugins *must document the meaning of the `version`* in their `Source.collect_source_info()` implementation's docstring
* BuildStream can then inspect that docstring at runtime and provide a way to also report that docstring in some machine readable way.

This approach would allow, from both a `collect_manifest` plugin perspective, **and** from a CLI wrapper script perspective, to generate reports which both:

* Have a section including the governing plugin documentation about the *version* for each plugin "kind" in the context of that pipeline
* Have the SourceInfo members for each Source instance, including the plugin "kind" name in the context of that pipeline

abderrahim:
> Ok, this all makes sense, however now that I look at it again, I no longer see any need whatsoever of having the `SourceInfoMedium` or `SourceVersionType` information present **at all** in the `SourceInfo` object.
> 
> Instead we only need the following members:
> 
>     * `kind`: Plugin "kind" string
> 
>     * `url`: Unaliased upstream full URL (not mirrored URL)
> 
>     * `version`: The machine readable _precise_ version (e.g. git sha, CAS/docker digest, sha256sum, etc)
> 
>     * `version_guess`: The plugin assisted _guess_ at what the human readable version is

I feel this pushes it too far in the opposite direction. This assumes a single mapping between a plugin kind and a (medium, version_type) pair, which isn't always the case. For instance the cargo2 plugin can use both tarballs and git repositories. Also plugins like git, git_repo and git_module are very similar in what they provide and having a user of the API need to know all the different plugins seems unweildy.

The main goal behind having an API like this is to reduce the amount of knowledge that a user of the API needs about specific plugins.

> Given that we can obtain the precise plugin and version for a given `kind`, and that that governing plugin documentation **must** already document precisely _what it means_ in the `version` member, the whole concept of the `medium` and `version_type` are obsolete and essentially meaningless.

"We" as humans, sure. But as a script using the API, having a well defined meaning for the version field and a standardized way to define things is very important. Even if a new value comes up later, then depending on the use case the script could output a warning or outright fail. But having it fail on every new plugin isn't scalable.

gtristan:
[...]
> 
> I feel this pushes it too far in the opposite direction. This assumes a single mapping between a plugin kind and a (medium, version_type) pair, which isn't always the case. For instance the cargo2 plugin can use both tarballs and git repositories. Also plugins like git, git_repo and git_module are very similar in what they provide and having a user of the API need to know all the different plugins seems unweildy.

Yeah, I was just now thinking about this one source with multiple source infos... its unfortunate :-/

For the different flavors of git, I'm less concerned, they could all be made to behave the same way, and in any case, they would probably anyway need to document what they mean (even if they share a freeform string) - if not, they are just taking advantage of a documented *convention* in upstream BuildStream (although that is still nasty because it makes BuildStream know what git is).

> > Given that we can obtain the precise plugin and version for a given `kind`, and that that governing plugin documentation **must** already document precisely _what it means_ in the `version` member, the whole concept of the `medium` and `version_type` are obsolete and essentially meaningless.
> 
> "We" as humans, sure. But as a script using the API, having a well defined meaning for the version field and a standardized way to define things is very important. Even if a new value comes up later, then depending on the use case the script could output a warning or outright fail. But having it fail on every new plugin isn't scalable.

While this point is now moot due to the preceding point, I would have agrued that it is in fact humans which need to know *"What to do with this specific plugin kind"* when encountering one programatically.

This is the same kind of documentation that will be needed for representing what the meaning of these free form strings are.

Do you think there is any reason to have 2 different strings  (one for the medium and one for the version) ? Or would it be sufficient to have a single string to define how a given `SourceInfo` instance behaves ?

gtristan:
[...]
> 
> Do you think there is any reason to have 2 different strings (one for the medium and one for the version) ? Or would it be sufficient to have a single string to define how a given `SourceInfo` instance behaves ?

I  came to the following conclusions:
* I liked that we had strong types, it helps how things are documented
* I liked that we had this separation of "where it comes from" and "what the version means", as stated before on this issue, this reduces the need for a lot of redundant enum values
* We do still have a need for third party plugin developers to use freeform strings, as discussed last week

I've updated the branch to use `Union[SourceVersionType, str]` and `Union[SourceInfoMedium, str]` types and reverted to the original implementation with this change, this seems to be a happy compromise.

src/buildstream/plugins/sources/local.py
gtristan: This is interesting upon closer inspection.

It seems that the internal `Directory._get_size()` abstract method is completely unused, while the `_artifact.py` class does *observe* this in `Artifact.cache()`, it does to in a completely meaningless way, in order to satisfy a local `size` variable which is later unused.

This whole thing leads me down an irrelevant trail of unused deadcode, only to confirm that *indeed*, buildstream needs to download a directory digest and crawl the merkle tree without having any foreknowledge of the size of the entire merkle tree to be downloaded... finally observing the `remote_execution.proto` `message Digest` proto definition tells me which size would be appropriate to report in a `{hash}/{size}` version definition.

In conclusion, I will use the `size_bytes` member of the `Digest` proto, rather than asking the `CasBasedDirectory._get_size()` to tell me what "size" to put there.

gtristan: Fixed.

src/buildstream/plugins/sources/workspace.py
gtristan: Fixed.

src/buildstream/downloadablefilesource.py
gtristan: This was a copy/paste error, that part was erronously carried over from the Source base class, fixed now.

src/buildstream/downloadablefilesource.py
gtristan: I know this guessing is used by collect_manifest also for git plugins, but I'd rather not extend the API at this point, it's a bit unwieldy as an API and includes the definition of a default pattern, overridable pattern and some few lines for the matching - it's a lot of mental gymnastics just to share the few lines of code which only make up a portion of the functionality.

Maybe later.

src/buildstream/source.py
gtristan: I've updated the docstring to specify this format for cased *when* the `DIGEST` type refers to a CAS digest.

src/buildstream/_frontend/widget.py
gtristan: We agreed to postpone this.

src/buildstream/source.py
gtristan: I've added this and updated the plugins which use fetchers.

In addition, I've added a public facing `SourceImplError` and updated the `collect_source_info()` docstring to specify that callers of the API need to handle `SourceImplError` exceptions.

src/buildstream/downloadablefilesource.py
gtristan: I don't believe that we should worry about congruency of cache keys with the result of URLs found in SBoMs.

However I agree that we should be *aware* of that detail.

I am *more* concerned that the URLs reported in the SBoMs are the intentionally expressed *upstreams* rather than downstream mirrors, as this is a more accurate description of the *provenance* of the data which led, under some configuration, to an artifact with a cache key,

> I'm not sure whether that's the case. Won't the alias definition from the parent project be used if the alias from > the subproject is mapped in the junction? That alias definition could potentially point to a different URL without > affecting the cache key.

This sounds like a *you get what you pay for* scenario. I.e. if you have overridden the defaults for the default aliases of sources in a subproject, rather than just adding your own mirrors, then the SBoMs would still accurately describe what your toplevel project has expressed in terms of inputs.

It is however undesirable for toplevels to redirect the default aliases, and as with the above note about cache keys, I agree that it should be *known*.

I'll leave this comment open for now as a reminder that this indeed should be documented in the appropriate places.

src/buildstream/source.py
gtristan: I'm going to leave this open as a reminder that there is STILL a missing part from this patch, which is adding support for specifying explicit versions in existing plugins in yaml.

For instance, if I have a git tag or a tarball with a name like `pony-rider-fork-with-gallop-support.tgz` on some hosting, where they did not specify a meaningful version, but the version is expressed in some other way, e.g. in the project readme, or is listed on a main page and indexed with some sha like `https://flying-ponies.com/releases/5a9121c44e888f7c5cd7060dfb153e0e9f2f2e95/pony-release.tgz`, it cannot be matched or guessed with any pattern regex.

In this case we need `downloadablefilesource.tgz` and any plugins to allow some source level configuration to explicitly specify these user provided overrides, for that plugin to prioritize and use in their `Source.collect_source_info()` implementations.

gtristan: I've now rectified this last part in downloadablefilesource.

Will rectify this in the external plugins where this applies.

I think this is unnecessary for the automated plugins such as pip and cargo, since *those* plugins obtain their various second order URIs as a *consequence* of already knowing a version, as such there is no sense in having an explicit override.

I'll close this thread now.

src/buildstream/downloadablefilesource.py
gtristan: This is now accounted for in the branch.

src/buildstream/source.py
gtristan: I'm satisfied with this branch and the https://github.com/apache/buildstream-plugins/pull/87 now to an extent that I'm sure we can do something correct in the git_repo and git_tag plugins.

Preferably, they will report a git sha only as the *version* and some user comprehensible-ish string in the *version_guess*.

Closing out this comment.

gtristan:
This is now ready, documentation and testing and all, along with https://github.com/apache/buildstream-plugins/pull/87 which is also ready to the same degree.

And so is: https://gitlab.com/BuildStream/buildstream-plugins-community/-/merge_requests/364

src/buildstream/source.py
```diff
@@ -293,8 +299,9 @@ def fetch(self, alias_override: Optional[AliasSubstitution] = None, **kwargs) ->
 
         Args:
            alias_override: The alias to use instead of the default one
-               defined by the :ref:`aliases <project_source_aliases>` field
-               in the project's config.
+               defined by the :ref:`aliases <project_source_aliases>` field in the
+               project's config. If provided, it must be used in when calling
```
abderrahim: nit: I think you want either "used" or "passed in".

src/buildstream/source.py
```diff
@@ -272,6 +420,208 @@ class AliasSubstitution:
...
+    """
+
+    OCI_IMAGE = "oci-image"
+    """
+    An OCI image, such as docker or podman images.
+    """
+
+    PYTHON_PACKAGE_INDEX = "pypi"
+    """
+    A python package obtained from a python package index like https://pypi.org
```
abderrahim, line 462: Is there a reason to treat this differently from `remote-file`?

src/buildstream/source.py
```diff
@@ -272,6 +420,208 @@ class AliasSubstitution:
...
+    """
+
+    INDEXED_VERSION = "indexed-version"
+    """
+    This type of version is used in cases where we have repositories which
+    have an interface to index content by version, and that no additional validation
+    is performed to insure the uniqueness of the downloaded content (not recommended).
+
+    In the case of plugins which use this version type, it is probable that
+    ``SourceInfo.guess_version == SourceInfo.version``.
```
abderrahim: This should be `SourceInfo.version_guess`

src/buildstream/source.py
```diff
@@ -272,6 +420,208 @@ class AliasSubstitution:
...
+    and/or :func:`SourceFetcher.get_source_info() <buildstream.source.SourceFetcher.get_source_info>`
+
+    See: :ref:`documentation on generating SourceInfo <core_source_info>`.
+
+    .. attention::
+
+       A given SourceInfo for a given element is **not** guaranteed to be unique for
+       a given :ref:`cache key <cachekeys>`.
+
+       While given well behaved source plugins, it is expected to obtain the same *cache keys*
```
abderrahim: I don't think this is a reasonable expectation. There are other things that could influence cache keys besides the source infos. I don't think it's reasonable to expect all the source plugin configuration to end up in the source info.

src/buildstream/source.py
```diff
@@ -272,6 +420,208 @@ class AliasSubstitution:
...
+       for a project where all of the SourceInfo for all elements are exactly the same, the
+       inverse is not to be expected.
```
abderrahim: Is this only referring to the urls below? Maybe this could be reformulated to make it clearer? (or maybe dropping the first sentence would help)

src/buildstream/source.py
```diff
@@ -311,6 +661,25 @@ def fetch(self, alias_override: Optional[AliasSubstitution] = None, **kwargs) ->
...
 
+    def get_source_info(self) -> SourceInfo:
+        """Get the :class:`.SourceInfo` object describing this source
+
+        This method should only be called whenever
+        :func:`Source.is_resolved() <buildstream.source.Source.is_resolved>`
+        returns ``True``.
+
+        SourceInfo objects created by implementors should be created with
+        :func:`Source.is_resolved() <buildstream.source.Source.create_source_info>`.
```
abderrahim: copy-paste error? (s/is_resolved/create_source_info/)

tests/frontend/source-info/plugins/extradata.py
```diff
@@ -0,0 +1,44 @@
...
+                "1234567",
+                version_guess="12",
+                extra_data={"pony": "green"},
+            )
+        ]
+
+
+# Plugin entry point
+def setup():
+
```
abderrahim: nit: drop this newline? (here and in the other sources)

abderrahim (approved):
A few comments on the doucmentation, but otherwise looks good.

src/buildstream/source.py
gtristan: Yes.

I believe the `pypi` plugin uses `remote-file` (it should) and downloads tarballs. The `pip` plugin on the other hand requires a pypi instance in order to fetch things (it is actually using host pip to interact with a pypi instance to download specific package versions discovered at track time).

src/buildstream/source.py
gtristan: Hmmm ok that's interesting and true.

For example, even considering unchanging element configurations, the directory where a source is *staged* is a source configuration which will/must effect cache keys, but is unrelated to source *provenance* and is not necessary in the SourceInfo.

I'll think of a better way to reword this.

tests/frontend/source-info/plugins/extradata.py
gtristan: Sure, for absolutely no reason at all, except that you like it, I have updated that :)

src/buildstream/source.py
gtristan: I've updated this entire statement, changing especially the first part and leaving the rest almost completely the same.

I think this is now a more accurate description.

src/buildstream/source.py
gtristan: This whole block is now updated as per the comment before, and is updated to clarify that it is indeed *only* the URLs which are expected to possibly change for a given set of identical cache keys.

I think this is clear now and addresses the comment appropriately.

src/buildstream/source.py
abderrahim: Fair enough.

abderrahim (approved):
Let's land this.

pyproject.toml
```diff
@@ -14,7 +14,7 @@
 [build-system]
 requires = [
     # We need at least version 36.6.0 that introduced "build_meta"
-    "setuptools>=36.6.0",
+    "setuptools>=36.6.0,<80",
```
nanonyme: This is a pretty nasty thing to do to distro packagers. Is this really necessary or can it be worked around?

abderrahim: Yeah, I have a fix for this. I'll open a separate pull request.

abderrahim: https://github.com/apache/buildstream/pull/2007

gtristan: Yup thanks a lot for that !

Rebased and this commit is gone now.

PR #2001, opened 2025-03-14
Support directory mtime to fix tests with recent buildbox-casd

juergbi:
This fixes test failures with recent versions of buildbox-casd, see [commit 85f0578f](https://gitlab.com/BuildGrid/buildbox/buildbox/-/commit/85f0578f) ("common: Support capturing directory mtime and mode").

### DIFF
--- a/src/buildstream/_cas/cascache.py
+++ b/src/buildstream/_cas/cascache.py
@@ -226,6 +226,11 @@ class CASCache:
             fullpath = os.path.join(dest, symlinknode.name)
             os.symlink(symlinknode.target, fullpath)
 
+        node_properties = directory.node_properties
+        if node_properties.HasField("mtime"):
+            mtime = utils._parse_protobuf_timestamp(node_properties.mtime)
+            utils._set_file_mtime(dest, mtime)
+
     # ensure_tree():
     #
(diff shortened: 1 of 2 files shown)


gtristan:
This changes the artifact content correct ?

Building a given artifact before this fix and after this fix, for the same cache key, will give you a *different* cas digest correct ?

Or, does this *avoid* buildstream automatically consuming this new behaviour from buildbox-casd ? So users who have *not* yet applied this patch but happen to have a newer buildbox-casd, will be producing different artifacts by default ?

juergbi:
`mtime` capture is enabled only for elements with a workspace. Normal builds are neither affected by this change nor the buildbox-casd change.

For workspace builds, the buildbox-casd change will result in a different CAS digest as directory mtimes are now captured as well. However, CAS digests of workspace builds are not deterministic anyway due to the file timestamps.

What this change does is to apply the directory mtimes captured by recent casd to directories in `bst artifact checkout`. So the only noticeable difference should be when you checkout artifacts of workspace builds. It will restore the directory timestamps from the build, and I can't think of any negative side-effects of this in the context of workspace builds.

And it fixes a test to work with old and new versions of buildbox-casd.

gtristan (approved):
LGTM !

PR #2007, opened 2025-05-04
Stop monkey-patching ScriptWriter

abderrahim:
It's no longer used (for as long back as I could see) and was removed
in setuptools 80.

Also update the minimum version of setuptools.


gtristan:
Aha thanks for finding a better solution!

PR #2008, opened 2025-05-07
Prefer better version matches

gtristan:
This fixes the issue of parsing typically formulated release tarball URLs.

Consider the URL `https://flying-ponies.com/releases/1.2/pony-flight-1.2.3.tgz`.

Without this change, we only get the version `1.2` (the directory where we put all of the `1.2` releases).

After this change, we get `1.2.3`.

This is achieved by explicitly supporting *optional groups* in the regex (which we already did, but in an not very specified way), and *preferring* more qualified matches by preferring *matches which had more groups present in the match* while iterating over all non-overlapping matches.

By causing the user to explicitly specify *groups* in the regex, we avoid the pitfall of relying on silly things like string length (where both strings "1.200" and "1.2.3" would be considered *equally qualified*).

### DIFF
--- a/src/buildstream/_frontend/cli.py
+++ b/src/buildstream/_frontend/cli.py
@@ -409,7 +409,7 @@ def help_command(ctx, command):
 @click.option(
     "--min-version",
     type=click.STRING,
-    default="2.4",
+    default="2.5",
     show_default=True,
     help="The required format version",
 )
--- a/src/buildstream/downloadablefilesource.py
+++ b/src/buildstream/downloadablefilesource.py
@@ -43,12 +43,9 @@ intantiating any Source which derives from DownloadableFileSource
   version string from the specified URI, in order to fill out the reported
   :attr:`~buildstream.source.SourceInfo.version_guess`.
 
-  The URI will be *searched* using this regular expression, and is allowed to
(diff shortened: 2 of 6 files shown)


gtristan:
FWIW with the increased complexity, and increased *clarity* on how the regex string works - I would be more happy now to make a well defined `utils` function to do the guess work and share that code with the plugin repos.

abderrahim (approved):
This works fine, but I still think it's a bit too complicated.

I feel that extracting the basename from the URL and using just that for guessing the version is good enough. I went through all the buildstream projects I work on and couldn't find any instance where the version isn't in the basename.

That said, I won't block this.

gtristan:
> FWIW with the increased complexity, and increased _clarity_ on how the regex string works - I would be more happy now to make a well defined `utils` function to do the guess work and share that code with the plugin repos.

I have now:
* Understood the regular expression nature a bit better
  * While it is possible to match *repeating groups* with a '*' wildcard, it is impossible to capture an *arbitrary number* of groups
* I have now created `utils.guess_version()` which reduces the implementer load by:
  * Doing the match iteration and preferring the match with the greatest amount of captured *groups*
  * Providing a central location to define the API contract for specifying `version-guess-pattern`

PR #2009, opened 2025-05-14
Additional source provenance in SourceInfo

gtristan:
With the recent addition of SourceInfo etc, we are closer to generating automated SBoMs,
and to a reasonable degree, everything that should be delegated to plugins has been delegated
via the `Source.collect_source_info()` and `SourceFetcher.get_source_info()` abstract methods, and
the utility function `utils.guess_version()` for version guessing.

This leaves some information which only users can explicitly provide in their buildstream projects,
notably, it is relevant for SBoM generation and compliance purposes, to include information such
as the project *homepage* URL, and an existing *issue tracking* URL.

It is presumed that these cannot be realistically guessed by plugin implementations, and that
project authors would have to provide these in some form.

This patch adds some source level provenance information which users can use to contribute to
the SourceInfo.

### DIFF
--- a/src/buildstream/_loader/metasource.py
+++ b/src/buildstream/_loader/metasource.py
@@ -26,14 +26,17 @@ class MetaSource:
     #    element_index: The index of the source in the owning element's source list
     #    element_kind: The kind of the owning element
     #    kind: The kind of the source
+    #    directory: A subdirectory where to stage the source
+    #    provenance: The user provided provenance information (e.g. homepage, issue tracking, etc).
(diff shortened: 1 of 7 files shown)


gtristan:
It occurs to me that this patch has issues with multi-url sources like `pip` and `cargo`.

To address this, I would suggest extending `Source.create_source_info()` to add optional, keyword-only `homepage` and `issue_tracker` arguments, which can be used to provide this information differently using their own user configuration attributes.

PR #2014, opened 2025-05-23
Add support for REAPI execution from the sandbox

### LINKED ISSUE #1985: Add option to expose remote execution services to the sandbox
Context: https://lists.apache.org/thread/scsz1ybc25d3x1oy33onfmm8rhr0kddg

We want to add an option to allow exposing subset of the remote execution services on a unix socket in the sandbox (similar to https://github.com/apache/buildstream/pull/1772 / https://github.com/apache/buildstream/pull/1945). This would allow REAPI aware clients (such as bazel or RECC) to take advantage of it.

As discussed on the mailing list, we only want to allow CAS, Remote Execution and Action Cache. We don't want to expose Remote Asset or Local CAS in the sandbox.

Ideally this would be implemented in buildbox-run (filed https://gitlab.com/BuildGrid/buildbox/buildbox/-/issues/183 for this) and buildstream only needs to pass the correct parameters to enable it.

Either way, this also needs sandbox configuration in buildstream to enable it.

juergbi:
Fixes #1985.

### DIFF
--- a/src/buildstream/_cas/casdprocessmanager.py
+++ b/src/buildstream/_cas/casdprocessmanager.py
@@ -45,9 +45,9 @@ _CASD_TIMEOUT = 300  # in seconds
 #
 # Minimum required version of buildbox-casd
 #
-_REQUIRED_CASD_MAJOR = 0
-_REQUIRED_CASD_MINOR = 0
-_REQUIRED_CASD_MICRO = 58
+_REQUIRED_CASD_MAJOR = 1
+_REQUIRED_CASD_MINOR = 2
+_REQUIRED_CASD_MICRO = 0
 
 
 # CASDProcessManager
@@ -76,7 +76,8 @@ class CASDProcessManager:
         messenger,
         *,
         reserved=None,
-        low_watermark=None
+        low_watermark=None,
+        local_jobs=None
     ):
         os.makedirs(path, exist_ok=True)
 
@@ -103,6 +104,17 @@ class CASDProcessManager:
 
         if protect_session_blobs:
             casd_args.append("--protect-session-blobs")
+
+        if local_jobs is not None:
+            try:
+                buildbox_run = utils._get_host_tool_internal("buildbox-run", search_subprojects_dir="buildbox")
+                casd_args.append("--buildbox-run={}".format(buildbox_run))
+                casd_args.append("--jobs={}".format(local_jobs))
+            except utils.ProgramNotFoundError:
+                # Not fatal as buildbox-run is not needed for remote execution
+                # and buildbox-casd local execution will never be used if
+                # buildbox-run is not available.
+                pass
 
         if remote_cache_spec:
             casd_args.append("--cas-remote={}".format(remote_cache_spec.url))
--- a/src/buildstream/sandbox/_config.py
+++ b/src/buildstream/sandbox/_config.py
@@ -34,6 +34,7 @@ if TYPE_CHECKING:
 #    build_arch: A canonical machine architecture name, as defined by Platform.canonicalize_arch()
 #    build_uid: The UID for the sandbox process
 #    build_gid: The GID for the sandbox process
+#    remote_apis_socket_path: The path to a UNIX socket providing REAPI access for nested remote execution
 #
 # If the build_uid or build_gid is unspecified, then the underlying sandbox implementation
 # does not guarantee what UID/GID will be used, but generally UID/GID 0 will be used in a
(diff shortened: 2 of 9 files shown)


abderrahim:
I tested this, and recc inside the sandbox is complaining
```
2025-05-27T08:22:40.382+0000 [218:140519756722176] [buildboxcommon_grpcretrier.cpp:134] [ERROR] ActionCache.GetActionResult() failed with: 12: Action Cache service is not enabled for instance ""
2025-05-27T08:22:40.383+0000 [218:140519756722176] [executioncontext.cpp:467] [ERROR] Error while querying action cache at "unix:/tmp/casd.sock": 12: Action Cache service is not enabled for instance ""
2025-05-27T08:22:40.385+0000 [218:140519756722176] [buildboxcommon_grpcretrier.cpp:134] [ERROR] Execution.Execute() failed with: 12: Execute: 12: Action Cache service is not enabled for instance ""
2025-05-27T08:22:40.386+0000 [218:140519756722176] [executioncontext.cpp:832] [ERROR] Error while calling `Execute()` on "unix:/tmp/casd.sock": 12: Execute: 12: Action Cache service is not enabled for instance ""
```

I dug a bit and it is due to the fact that I'm using a cache storage service. Disabling it works fine.

src/buildstream/sandbox/_config.py
```diff
@@ -66,15 +74,15 @@ def __init__(
...
+    def to_dict(self) -> Dict[str, Union[str, int, bool]]:
 
         # Assign mandatory portions of the sandbox configuration
         #
         # /!\ No additional mandatory members can ever be added to
         #     the sandbox configuration, as that would result in
         #     breaking cache key stability.
         #
-        sandbox_dict: Dict[str, Union[str, int]] = {"build-os": self.build_os, "build-arch": self.build_arch}
+        sandbox_dict: Dict[str, Union[str, int, bool]] = {"build-os": self.build_os, "build-arch": self.build_arch}
```
gtristan: Is this outdated perhaps ? Maybe there was an added boolean here in a previous iteration of the patch ?

I think here there is only the possibility of a new string being serialized in the dict.

gtristan:
Generally like this patch overall... although it is still in draft, do we want this in 2.5 ?

Should we add an integration test for exercising `recc` ? Would be pretty cool if we had an example for this in the canonical format too ?

I'd be happy to add a tested example based on the [recc demo](https://gitlab.com/tristanvb/buildstream-recc-demo/) to sweeten the pot if we think we're ready to go ahead with this...

abderrahim:
I feel it's still not ready. While it's good enough, ~~it depends on unreleased buildbox~~ (buildbox was released earlier today, my bad) and there are a few corner cases such as https://github.com/apache/buildstream/pull/2014#issuecomment-2911646294 above that need to be handled.

I'd say let's leave it till 2.6. I think that's also @juergbi's opinion as this is still marked as draft.

src/buildstream/sandbox/_config.py
juergbi: Yes, there was indeed an additional boolean config option in a previous iteration. Will remove this.

juergbi:
Leaving it till 2.6 is fine with me. I anyway expect typical use cases to require #1991 and that's also still a draft.

juergbi:
> I dug a bit and it is due to the fact that I'm using a cache storage service. Disabling it works fine.

I've added a check for this to fail the build with a clear error message. In https://github.com/apache/buildstream/tree/jbilleter/action-cache-service, I'm adding support for `action-cache-service`, which then enables use of the REAPI socket with `storage-service`. That branch is not yet tested (or documented), though, and I think it's ok to first merge this and then the `action-cache-service` addition.

I've also added a minimal test to this branch.

CI is currently failing due to BuildBox being too old for the added test. https://github.com/apache/buildstream/pull/2023 updates the CI images based on https://gitlab.com/BuildStream/buildstream-docker-images/-/merge_requests/230.

tests/integration/project/elements/sandbox/remote-apis-socket.bst
```diff
@@ -0,0 +1,13 @@
...
+  - filename: base.bst
+    type: build
+
+sandbox:
+  remote-apis-socket:
+    path: /tmp/reapi.sock
+
+config:
+  build-commands:
+    - test -e /tmp/reapi.sock
```
abderrahim: Can't we test that it's a socket instead?

```suggestion
    - test -S /tmp/reapi.sock
```

abderrahim:
> I'm adding support for action-cache-service, which then enables use of the REAPI socket with storage-service. That branch is not yet tested (or documented), though, and I think it's ok to first merge this and then the action-cache-service addition.

Yeah, this is fine. With the clear error message here, we can land this even if it doesn't support the cache storage-service.

I'll test the action-cache-service branch in the mean time

tests/integration/project/elements/sandbox/remote-apis-socket.bst
juergbi: Good suggestion, changed.

src/buildstream/_cas/casdprocessmanager.py
```diff
@@ -45,9 +45,9 @@
 #
 # Minimum required version of buildbox-casd
 #
-_REQUIRED_CASD_MAJOR = 0
-_REQUIRED_CASD_MINOR = 0
-_REQUIRED_CASD_MICRO = 58
+_REQUIRED_CASD_MAJOR = 1
+_REQUIRED_CASD_MINOR = 2
```
abderrahim, line 49: I think we can have some cleanups of fallback code paths now that we require buildbox 1.2. That shouldn't block this pull request, but it should be nice to have it done before the next release.

abderrahim (approved):

PR #2015, opened 2025-05-24
cascache: fix pushing with remote cache

abderrahim:
CASCache.send_blobs() expects to be able to iterate twice on the passed digests (though it only does so when using a remote cache). Make sure to pass a list rather than a generator.


gtristan:
Ok this looks totally sensible !

PR #2016, opened 2025-05-26
source.py: Make SourceInfo.serialize() a public API.

gtristan:
This serializes the `SourceInfo` into a regular simple dictionary object that can be easily serialized into formats like json or yaml.

While it was requested to make this public in some way, it is not particularly useful to make this serialize directly into a yaml string, since normally one will probably want to serialize a *list of SourceInfo* objects reported for a given *element*, in whatever data structure one would want to serialize this into.

For this reason, we export the current serialization which just converts the `SourceInfo` object into something that other serialization python libraries can reliably consume (simple dictionaries).

### DIFF
--- a/src/buildstream/_frontend/widget.py
+++ b/src/buildstream/_frontend/widget.py
@@ -453,7 +453,7 @@ class LogLine(Widget):
 
                     serialized_sources = []
                     for s in source_infos:
-                        serialized = s._serialize()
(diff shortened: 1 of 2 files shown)


shymega (approved):
LGTM - I think we're good with this one.

abderrahim (approved):

PR #2018, opened 2025-06-02
core warnings for missing Source.collect_source_info() implementations

gtristan:
This allows projects to have a policy for only using sources which provide the new source provenance introspection routines.


src/buildstream/source.py
```diff
@@ -1078,11 +1063,8 @@ def collect_source_info(self) -> Iterable[SourceInfo]:
         SourceInfo objects created by implementors should be created with
         :func:`Source.create_source_info() <buildstream.source.Source.create_source_info>`.
 
-        Returns: the :class:`.SourceInfo` objects describing this source
-
-        Raises:
-           :class:`.SourceImplError`: if the source class does not implement this method and does not implement
-                                      :func:`SourceFether.get_source_info() <buildstream.source.SourceFetcher.get_source_info>`
+        Returns: the :class:`.SourceInfo` objects describing this source, or ``None`` if the
+                 SourceFetcher does not implement this method.
```
abderrahim: I think you meant Source rather than SourceFetcher.

tests/frontend/source-info-unimplemented/fatal/elements/unimplemented.bst
```diff
@@ -0,0 +1,4 @@
+kind: import
+
+sources:
+- kind: fatal_unimplemented
```
abderrahim, line 4: Is there a reason to call the plugins "fatal_unimplemented" and "non_fatal_unimplemented"? AFAICT, they are identical and the difference is in the project rather than the sources.

abderrahim (approved):
A couple minor comments.

tests/frontend/source-info-unimplemented/fatal/elements/unimplemented.bst
gtristan: Yes, it was originally just called "unimplemented.py".

The problem is that tox does this weird stuff where it doesn't like to have multiple modules with the same name that are not somehow joined by a whole tree of directories that all contain `__init__.py` files.

gtristan: I should say I first considered generating the project.conf differently in the same directory but weighed the extra code against just having some separate files and decided on the latter, it's fairly arbitrary, while I have a slight preference for static test data over generated files.

src/buildstream/source.py
gtristan: Ah good catch, copy/paste error.

Fixed.

PR #2022, opened 2025-06-12
_protos: regenerate with grpcio-tools 1.73

abderrahim:
This makes it compatible with protobuf 30+, a.k.a. 6, a.k.a. 6.30+


juergbi:
According to https://protobuf.dev/support/cross-version-runtime-guarantee/:

> As of 6.32.0, all generated code since 3.20.0 will be supported until at least 8.x.y.

I.e., there should be long forward compatibility of generated Python code. Assuming this is correct, we should not upgrade grpcio-tools for code generation. We should instead loosen the protobuf requirement to allow use with protobuf 6 and later.

Runtime warnings have been reported, though, in #2042. We should investigate the reason for these warnings before taking any action.

juergbi:
The protobuf warning has been removed upstream: https://github.com/protocolbuffers/protobuf/commit/1af7fd407c85c9c1f34b6ca8441cfb3a72d4f5ea

juergbi:
Based on the above, I've opened #2043 as alternative to this PR.

juergbi:
Closing this as #2043 has been merged instead.

PR #2024, opened 2025-06-14
Add `action-cache-service` support

juergbi:
This enables use of `remote-apis-socket` (#2014) with `storage-service` also for local execution.


juergbi:
Replaced by #2031 to support a remote action cache without having to switch to a cache `storage-service` configuration.

PR #2025, opened 2025-06-24
tests: use importlib.metadata instad of pkg_resources

abderrahim:
Follow up to #1948


tests/testutils/setuptools.py
```diff
@@ -1,55 +0,0 @@
...
-        dist, package = lookup_string.split(":")
-
-        def mock_entry(pdist, pentry_point, ppackage):
-            assert pdist == dist
-            assert pentry_point == entry_point
-            assert ppackage == package
-
-            return MockEntry(datafiles, package)
-
-        monkeypatch.setattr(pkg_resources, "get_entry_info", mock_entry)
```
abderrahim, line 53: I don't know what's the purpose of this file, but `pkg_resources.get_entry_point()` is no longer used by buildstream (since #1948). I'm deleting it because I think it's unused now, but we might want to replace it if it was doing something useful.

abderrahim: @juergbi pointed it to me that it was used in a test `tests/plugins/third_party.py`. The test in question was removed in c7f69bcc2d772157cde3a932c0e93f09a329019f, so this is safe to delete.

juergbi (approved):
I assume `types-setuptools` can be removed from tox.ini as well.

LGTM overall, assuming CI passes on all Python versions.

gtristan (approved):
LGTM.

PR #2026, opened 2025-06-24
Migrate Python formatting to ruff from black

kieran-ryan:
Migrates Python formatting to [ruff](https://docs.astral.sh/ruff/) from [black](https://black.readthedocs.io/en/stable/) to speed up formatting by multiples (see below benchmarks); and for robustness improvements; such as those applied by the formatting of this merge request to string formatting, trailing commas rather than a mixture with and without; improved formatting of arguments, whitespace line handling, etc. - though all of which are included with later releases of various Python formatter tools. 

The [ruff formatter](https://docs.astral.sh/ruff/formatter/) is "designed as a drop-in replacement for [Black](https://pypi.org/project/black/)". The caching in particular improves development workflows with tools such as IDEs and hooks.

[Ruff's default configuration](https://docs.astral.sh/ruff/configuration/) is a superset of those specified by the _black_ configuration within the `pyproject.toml`; and thus only required extension to the custom path `src/buildstream/_protos`.

Raising based on observation and in case of interest.

#### Cold Runs

  | Black | Ruff | Delta | % Change |
  |----------|-----|-------|----------|
  | 13.86s | 2.07s | 11.79s | -85% |

<details>
<summary>Screenshots</summary>
<h5>Black</h5>

<h5>Ruff</h5>

</details>

#### Warm Runs

  | Black | Ruff | Delta | % Change |
  |----------|-----|-------|----------|
  | 12.31s | 0.54s | 11.77s | -95.6% |

<details>
<summary>Screenshots</summary>
<h5>Black</h5>

<h5>Ruff</h5>

</details>

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -95,7 +95,7 @@ confidence=
 #    be nice, but too much work atm.
 #
 # 3) Messages related to code formatting
-#    Since we use Black to format code automatically, there's no need for
+#    Since we use Ruff to format code automatically, there's no need for
 #    pylint to also check for those things.
 #
 
@@ -146,7 +146,7 @@ disable=,
 	unused-private-member,
 
         ##################################################
-        # Formatting-related messages, enforced by Black #
+        # Formatting-related messages, enforced by Ruff #
         ##################################################
 
         line-too-long,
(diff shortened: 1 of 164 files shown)


cs-shadow:
hi @kieran-ryan , I'd be in favor of this change in principle.

Although it may be good to float this on the [mailing list](https://lists.apache.org/list.html?dev@buildstream.apache.org) first as it may be a little bit of disruptive change for people who may be in the habit of running Black.

Especially because it seems like Ruff would do more reformatting than "just" a drop-in replacement.

PR #2030, opened 2025-07-03
Draft: Feature: Add support for generating json from bst show command

kevinschoon:
This adds a new flag called --json to the `bst show` command which when specified will make bst output a machine readable JSON dump of the dependency state within the current buildstream project. Having JSON support makes it easier to write tools that need to inspect the configuration of a buildstream enabled project.

### DIFF
--- a/src/buildstream/_frontend/cli.py
+++ b/src/buildstream/_frontend/cli.py
@@ -14,6 +14,7 @@
 import os
 import re
 import sys
+import json
 from functools import partial
 
 import shutil
@@ -21,7 +22,7 @@ import click
 from .. import _yaml
 from .._exceptions import BstError, LoadError, AppError, RemoteError
 from .complete import main_bashcomplete, complete_path, CompleteUnhandled
-from ..types import _CacheBuildTrees, _SchedulerErrorAction, _PipelineSelection, _HostMount, _Scope
+from ..types import _CacheBuildTrees, _ElementKind, _SchedulerErrorAction, _PipelineSelection, _HostMount, _Scope
 from .._remotespec import RemoteSpec, RemoteSpecPurpose
 from ..utils import UtilError
 
@@ -73,6 +74,7 @@ class RemoteSpecType(click.ParamType):
 ##################################################################
 #            Override of click's main entry point                #
 ##################################################################
+
 
 # search_command()
 #
@@ -538,6 +540,7 @@ def build(
 @click.option(
     "--except", "except_", multiple=True, type=click.Path(readable=False), help="Except certain dependencies"
 )
+@click.option("--json", "as_json", is_flag=True, help="Output the information as machine readable JSON")
 @click.option(
     "--deps",
     "-d",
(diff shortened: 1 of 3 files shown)


src/buildstream/_frontend/cli.py
```diff
@@ -538,6 +540,7 @@ def build(
 @click.option(
     "--except", "except_", multiple=True, type=click.Path(readable=False), help="Except certain dependencies"
 )
+@click.option("--json", "as_json", is_flag=True, help="Output the information as machine readable JSON")
```
abderrahim, line 543: It would be nice to have a similar option that outputs YAML, as that's what buildstream uses.

src/buildstream/_frontend/cli.py
```diff
@@ -552,7 +555,29 @@ def build(
             _PipelineSelection.ALL,
         ],
     ),
-    help="The dependencies to show",
+    help="Types of Elements to Show",
```
abderrahim, line 558: Please keep the old description. "Types of elements" doesn't make sense here.

src/buildstream/_frontend/cli.py
```diff
@@ -552,7 +555,29 @@ def build(
...
+)
+@click.option(
+    "--kind",
+    "-k",
+    default=[],
+    show_default=True,
+    multiple=True,
+    type=FastEnumType(
+        _ElementKind,
```
abderrahim, line 567: "Element Kind" is definitely not how you should name this. The element kind is already a concept in buildstream, and it's not this.

I don't know what would be the best way to name this option though. info? keys?

Let's wait for other maintainer's opinions, they might have a better idea.

src/buildstream/_frontend/cli.py
```diff
@@ -629,19 +654,44 @@ def show(app, elements, deps, except_, order, format_):
...
+    \b
+        bst show --json --kind all
     """
     with app.initialized():
 
+        if as_json and format_:
+            raise AppError("--format and --json are mutually exclusive")
+
+        if format_ and kind:
+            raise AppError("--format does not support --kind")
```
abderrahim, line 674: Not sure I like this error message, as I see `--format` and `--kind` (in your current implementation) to be essentially the same thing except `--format` is used in the default mode and `--kind` is used in the json mode.

Maybe we could change it such that `--format` could take the special value `json`, and that would unlock the (replacement for the) `--kind` option.

Let's wait for other opinions.

abderrahim (commented):
Overall, I like this. It's more easily parsed, and doesn't suffer from the pitfalls of trying to format yaml using the current `--format` option.

I think adding yaml output with the same semantics would be good, especially as buildstream is already yaml-centric.

The implementation itself uses names that are already taken, but that shouldn't be hard to fix.

abderrahim:
Oh, I forgot. This also needs tests. You can probably just add a test or two to `tests/frontend/show.py`

kevinschoon:
Closing this in favor of https://github.com/apache/buildstream/pull/2035

Issue #2020, opened 2025-06-11
Cached failed action is used when retrying a failed build


abderrahim:
When using remote execution, if a build fails the failure is saved in the action cache. However builds can fail for a variety of reasons (e.g. running out of resources) that are not necessarily reproducible.

Buildstream offers to retry at the command line and also offers the `--retry-failed` command line argument for this purpose. However, it will still get a cached failure from the action cache.

Here is an example log (I extracted just the things relevant to the element in question):
```
[11:08:35][--:--:--][8524a2f2][    pull:freedesktop-sdk.bst:components/fontconfig.bst] START   freedesktop-sdk/components-fontconfig/8524a2f2-pull.20250611-110835.log
[11:08:35][--:--:--][8524a2f2][    pull:freedesktop-sdk.bst:components/fontconfig.bst] STATUS  Pulling artifact 8524a2f2 <- https://gbm.gnome.org:11004
[11:08:35][--:--:--][8524a2f2][    pull:freedesktop-sdk.bst:components/fontconfig.bst] STATUS  Pulling data for artifact 8524a2f2 <- https://gbm.gnome.org:11004
[11:08:35][--:--:--][8524a2f2][    pull:freedesktop-sdk.bst:components/fontconfig.bst] INFO    Pulled artifact 8524a2f2 <- https://gbm.gnome.org:11004
[11:08:35][--:--:--][8524a2f2][    pull:freedesktop-sdk.bst:components/fontconfig.bst] INFO    Discarded failed build
[11:08:35][00:00:00][8524a2f2][    pull:freedesktop-sdk.bst:components/fontconfig.bst] SKIPPED Pull
[11:08:35][--:--:--][8524a2f2][   fetch:freedesktop-sdk.bst:components/fontconfig.bst] START   freedesktop-sdk/components-fontconfig/8524a2f2-fetch.20250611-110835.log
[11:08:35][--:--:--][8524a2f2][   fetch:freedesktop-sdk.bst:components/fontconfig.bst] STATUS  Pulling source 031032b8 <- https://gbm.gnome.org:11004
[11:08:36][--:--:--][8524a2f2][   fetch:freedesktop-sdk.bst:components/fontconfig.bst] STATUS  Pulling data for source 031032b8 <- https://gbm.gnome.org:11004
[...]

PR #2033, opened 2025-07-08
_sandboxremote.py: avoid reusing failed actions

abderrahim:
This is a proposal towards https://github.com/apache/buildstream/issues/2020. It's probably not the most efficient way to do it, but at least it works.

What this does is:
* ignore failures found in the action cache
* ask the remote execution service to not look in the cache

### DIFF
--- a/src/buildstream/sandbox/_sandboxremote.py
+++ b/src/buildstream/sandbox/_sandboxremote.py
@@ -154,7 +154,7 @@ class SandboxRemote(SandboxREAPI):
 
         stub = self.exec_remote.exec_service
         request = remote_execution_pb2.ExecuteRequest(
-            instance_name=self.exec_remote.instance_name, action_digest=action_digest, skip_cache_lookup=False
+            instance_name=self.exec_remote.instance_name, action_digest=action_digest, skip_cache_lookup=True
         )
 
         def __run_remote_command(stub, execute_request=None, running_operation=None):
@@ -346,10 +346,14 @@ class SandboxRemote(SandboxREAPI):
             if e.code() != grpc.StatusCode.NOT_FOUND:
                 raise SandboxError("Failed to query action cache: {} ({})".format(e.code(), e.details()))
             return None
-        else:
-            context = self._get_context()
-            context.messenger.info("Action result found in action cache", element_name=self._get_element_name())
-            return result
+
+        context = self._get_context()
+        if result.exit_code:
+            context.messenger.info("Ignoring failed action result found in action cache", element_name=self._get_element_name())
+            return None
+
+        context.messenger.info("Action result found in action cache", element_name=self._get_element_name())
+        return result
 
     @staticmethod
     def _extract_action_result(operation):


src/buildstream/sandbox/_sandboxremote.py
```diff
@@ -154,7 +154,7 @@ def run_remote_command(self, action_digest):
 
         stub = self.exec_remote.exec_service
         request = remote_execution_pb2.ExecuteRequest(
-            instance_name=self.exec_remote.instance_name, action_digest=action_digest, skip_cache_lookup=False
+            instance_name=self.exec_remote.instance_name, action_digest=action_digest, skip_cache_lookup=True
```
juergbi, line 157: We don't want to always skip cache lookup. If no `action-service-cache` is declared, internal action cache lookup by the remote execution server should still be used.

juergbi:
I think the main issue is that failed actions are cached in the action cache in the first place. While there may be circumstances where caching failed actions is useful, I don't think we need this for BuildStream at all (given that we have a higher level caching mechanism where we already support caching failures) and most remote execution servers default to not caching failed actions, mainly because failures may be spurious, e.g., due to a worker running out of RAM.

BuildGrid caches failures by default but it can be disabled with `cache-failed-actions: false`. buildbox-casd currently unconditionally caches failures but I think we should change that or at least add an option to disable it.

On the BuildStream side, might it be sufficient to skip action cache lookup (direct action cache query as well as indirectly via Execute()) if `context.build_retry_failed` is set?

src/buildstream/sandbox/_sandboxremote.py
abderrahim: Yeah, even when the `action-cache-service` is defined, there is still value in having the execution re-do cache lookup (in case the action is requested and built by someone else before our action reaches the top of the queue).

You're right that this is probably working around a "broken" remote execution server. I've only tested it with buildbox-casd, and wasn't aware it was different with other servers.

abderrahim:
> I think the main issue is that failed actions are cached in the action cache in the first place. While there may be circumstances where caching failed actions is useful, I don't think we need this for BuildStream at all (given that we have a higher level caching mechanism where we already support caching failures) and most remote execution servers default to not caching failed actions, mainly because failures may be spurious, e.g., due to a worker running out of RAM.
> 
> BuildGrid caches failures by default but it can be disabled with `cache-failed-actions: false`. buildbox-casd currently unconditionally caches failures but I think we should change that or at least add an option to disable it.

Yeah, I was using this with buildbox-casd as a server. The failures in question were due to a bug in buildbox-fuse.
 
> On the BuildStream side, might it be sufficient to skip action cache lookup (direct action cache query as well as indirectly via Execute()) if `context.build_retry_failed` is set?

The thing is `context.build_retry_failed` is only set when passing the option on the command line or the config file. It doesn't work when you choose retry on the prompt.

juergbi:
> The thing is `context.build_retry_failed` is only set when passing the option on the command line or the config file. It doesn't work when you choose retry on the prompt.

Good point but maybe we can find a way to forward this information also in the interactive retry case.

PR #2035, opened 2025-07-08
Implement bst inspect subcommand

kevinschoon:
This adds a new subcommand to bst called `inspect` which dumps structured data to stdout of a given project. The data sent to stdout can be either JSON or YAML with JSON being the default.

Having this command available will make writing external tools which need to inspect the state of a buildstream project more easy since JSON and YAML encoding are widely supported. This command may easily be extended in the future to support other inspectable elements related to bst project which isn't present in this first commit.

### DIFF
--- a/src/buildstream/_frontend/app.py
+++ b/src/buildstream/_frontend/app.py
@@ -305,7 +305,7 @@ class App:
             self.stream.set_project(self.project)
 
             # Initialize the inspector
-            self.inspector = Inspector(self.stream, self.project)
+            self.inspector = Inspector(self.stream, self.project, self.context)
 
             # Run the body of the session here, once everything is loaded
             try:
--- a/src/buildstream/_frontend/app.py
+++ b/src/buildstream/_frontend/app.py
@@ -40,7 +40,6 @@ from ..utils import UtilError
 from .profile import Profile
 from .status import Status
 from .widget import LogLine
-from .inspect import Inspector
 
 # Intendation for all logging
 INDENT = 4
@@ -66,7 +65,6 @@ class App:
         self.logger = None  # The LogLine object
         self.interactive = None  # Whether we are running in interactive mode
         self.colors = None  # Whether to use colors in logging
-        self.inspector = None # If inspection is required
 
         #
         # Private members
@@ -303,9 +301,6 @@ class App:
             # Set the project on the Stream, this can be None if there is no project.
             #
             self.stream.set_project(self.project)
-
-            # Initialize the inspector
-            self.inspector = Inspector(self.stream, self.project, self.context)
 
             # Run the body of the session here, once everything is loaded
             try:
--- a/src/buildstream/_frontend/app.py
+++ b/src/buildstream/_frontend/app.py
@@ -40,6 +40,7 @@ from ..utils import UtilError
 from .profile import Profile
 from .status import Status
 from .widget import LogLine
+from .inspect import Inspector
 
 # Intendation for all logging
 INDENT = 4
@@ -65,6 +66,7 @@ class App:
         self.logger = None  # The LogLine object
         self.interactive = None  # Whether we are running in interactive mode
         self.colors = None  # Whether to use colors in logging
+        self.inspector = None # If inspection is required
 
         #
         # Private members
@@ -301,6 +303,10 @@ class App:
             # Set the project on the Stream, this can be None if there is no project.
             #
             self.stream.set_project(self.project)
+
+            # Initialize the inspector
+            if self._session_name == "Inspect":
+                self.inspector = Inspector(self.stream, self.project)
 
             # Run the body of the session here, once everything is loaded
             try:
--- a/src/buildstream/_frontend/app.py
+++ b/src/buildstream/_frontend/app.py
@@ -305,8 +305,7 @@ class App:
             self.stream.set_project(self.project)
 
             # Initialize the inspector
-            if self._session_name == "Inspect":
-                self.inspector = Inspector(self.stream, self.project)
+            self.inspector = Inspector(self.stream, self.project)
 
             # Run the body of the session here, once everything is loaded
             try:
--- a/src/buildstream/_frontend/cli.py
+++ b/src/buildstream/_frontend/cli.py
@@ -21,7 +21,7 @@ import click
 from .. import _yaml
 from .._exceptions import BstError, LoadError, AppError, RemoteError
 from .complete import main_bashcomplete, complete_path, CompleteUnhandled
-from ..types import _CacheBuildTrees, _SchedulerErrorAction, _PipelineSelection, _HostMount, _Scope, _Encoding
+from ..types import _CacheBuildTrees, _SchedulerErrorAction, _PipelineSelection, _HostMount, _Scope
 from .._remotespec import RemoteSpec, RemoteSpecPurpose
 from ..utils import UtilError
 
@@ -536,7 +536,6 @@ def build(
 ##################################################################
 @cli.command(name="inspect", short_help="Inspect Project Information")
 @click.option("-s", "--state", default=False, show_default=True, is_flag=True, help="Show information that requires inspecting remote state")
-@click.option("-e", "--encoding", default=_Encoding.JSON, show_default=True, type=FastEnumType(_Encoding, [_Encoding.JSON, _Encoding.YAML]))
 @click.option(
     "--deps",
     "-d",
@@ -555,7 +554,7 @@ def build(
 )
 @click.argument("elements", nargs=-1, type=click.Path(readable=False))
 @click.pass_obj
-def inspect(app, elements, state, encoding, deps):
+def inspect(app, elements, state, deps):
     """Access structured data about a given buildstream project and it's computed elements.
 
     Specifying no elements will result in showing the default targets
@@ -575,8 +574,6 @@ def inspect(app, elements, state, encoding, deps):
         run:   Runtime dependencies, including the element itself
         build: Build time dependencies, excluding the element itself
         all:   All dependencies
-
-    Use ``--encoding JSON|YAML`` to control the type of encoding written to stdout.
 
     If ``--state`` is toggled then pipeline elements which require remote state will be
     shown in addition to information that is available on the local system.
@@ -604,7 +601,7 @@ def inspect(app, elements, state, encoding, deps):
 
     """
     with app.initialized(session_name="Inspect"):
-        app.inspector.dump_to_stdout(elements, selection=deps, encoding=encoding, with_state=state)
+        app.inspector.dump_to_stdout(elements, selection=deps, with_state=state)
 
 
 ##################################################################
--- a/src/buildstream/_frontend/cli.py
+++ b/src/buildstream/_frontend/cli.py
@@ -21,7 +21,7 @@ import click
 from .. import _yaml
 from .._exceptions import BstError, LoadError, AppError, RemoteError
 from .complete import main_bashcomplete, complete_path, CompleteUnhandled
-from ..types import _CacheBuildTrees, _SchedulerErrorAction, _PipelineSelection, _HostMount, _Scope
+from ..types import _CacheBuildTrees, _SchedulerErrorAction, _PipelineSelection, _HostMount, _Scope, _Encoding
 from .._remotespec import RemoteSpec, RemoteSpecPurpose
 from ..utils import UtilError
 
@@ -529,6 +529,82 @@ def build(
             ignore_project_source_remotes=ignore_project_source_remotes,
             retry_failed=retry_failed,
         )
+
+
+##################################################################
+#                           Inspect Command                      #
+##################################################################
+@cli.command(name="inspect", short_help="Inspect Project Information")
+@click.option("-s", "--state", default=False, show_default=True, is_flag=True, help="Show information that requires inspecting remote state")
+@click.option("-e", "--encoding", default=_Encoding.JSON, show_default=True, type=FastEnumType(_Encoding, [_Encoding.JSON, _Encoding.YAML]))
+@click.option(
+    "--deps",
+    "-d",
+    default=_PipelineSelection.ALL,
+    show_default=True,
+    type=FastEnumType(
+        _PipelineSelection,
+        [
+            _PipelineSelection.NONE,
+            _PipelineSelection.RUN,
+            _PipelineSelection.BUILD,
+            _PipelineSelection.ALL,
+        ],
+    ),
+    help="The dependencies to show",
+)
+@click.argument("elements", nargs=-1, type=click.Path(readable=False))
+@click.pass_obj
+def inspect(app, elements, state, encoding, deps):
+    """Access structured data about a given buildstream project and it's computed elements.
+
+    Specifying no elements will result in showing the default targets
+    of the project. If no default targets are configured, all project
+    elements will be shown.
+
+    When this command is executed from a workspace directory, the default
+    is to show the workspace element.
+
+    By default this will show all of the dependencies of the
+    specified target element.
+
+    Specify ``--deps`` to control which elements to show:
+
+    \b
+        none:  No dependencies, just the element itself
+        run:   Runtime dependencies, including the element itself
+        build: Build time dependencies, excluding the element itself
+        all:   All dependencies
+
+    Use ``--encoding JSON|YAML`` to control the type of encoding written to stdout.
+
+    If ``--state`` is toggled then pipeline elements which require remote state will be
+    shown in addition to information that is available on the local system.
+
+    Examples:
+
+        # Show all default elements with remote information
+        \n
+            bst inspect --state
+
+
+        # A specific target (glob pattern)
+        \n
+            bst inspect -s public/*.bst
+
+
+        # With a dependency target
+        \n
+            bst inspect -d run --state
+
+
+        # Show each remote file source (with the help of jq)
+        \n
+            bst inspect -d all | jq '.elements.[].sources | select( . != null ) | .[] | select( .medium == "remote-file")
+
+    """
+    with app.initialized(session_name="Inspect"):
+        app.inspector.dump_to_stdout(elements, selection=deps, encoding=encoding, with_state=state)
 
 
 ##################################################################
--- a/src/buildstream/_frontend/cli.py
+++ b/src/buildstream/_frontend/cli.py
@@ -535,7 +535,6 @@ def build(
 #                           Inspect Command                      #
 ##################################################################
 @cli.command(name="inspect", short_help="Inspect Project Information")
-@click.option("-s", "--state", default=False, show_default=True, is_flag=True, help="Show information that requires inspecting remote state")
 @click.option(
     "--deps",
     "-d",
@@ -554,7 +553,7 @@ def build(
 )
 @click.argument("elements", nargs=-1, type=click.Path(readable=False))
 @click.pass_obj
-def inspect(app, elements, state, deps):
+def inspect(app, elements, deps):
     """Access structured data about a given buildstream project and it's computed elements.
 
     Specifying no elements will result in showing the default targets
@@ -575,24 +574,16 @@ def inspect(app, elements, state, deps):
         build: Build time dependencies, excluding the element itself
         all:   All dependencies
 
-    If ``--state`` is toggled then pipeline elements which require remote state will be
-    shown in addition to information that is available on the local system.
-
     Examples:
-
-        # Show all default elements with remote information
-        \n
-            bst inspect --state
-
 
         # A specific target (glob pattern)
         \n
-            bst inspect -s public/*.bst
+            bst inspect public/*.bst
 
 
         # With a dependency target
         \n
-            bst inspect -d run --state
+            bst inspect -d run
 
 
         # Show each remote file source (with the help of jq)
@@ -601,7 +592,7 @@ def inspect(app, elements, state, deps):
 
     """
     with app.initialized():
-        app.inspector.dump_to_stdout(elements, selection=deps, with_state=state)
+        app.inspector.dump_to_stdout(elements, selection=deps)
 
 
 ##################################################################
--- a/src/buildstream/_frontend/cli.py
+++ b/src/buildstream/_frontend/cli.py
@@ -600,7 +600,7 @@ def inspect(app, elements, state, deps):
             bst inspect -d all | jq '.elements.[].sources | select( . != null ) | .[] | select( .medium == "remote-file")
 
     """
-    with app.initialized(session_name="Inspect"):
+    with app.initialized():
         app.inspector.dump_to_stdout(elements, selection=deps, with_state=state)
 
 
--- a/src/buildstream/_frontend/cli.py
+++ b/src/buildstream/_frontend/cli.py
@@ -24,6 +24,7 @@ from .complete import main_bashcomplete, complete_path, CompleteUnhandled
 from ..types import _CacheBuildTrees, _SchedulerErrorAction, _PipelineSelection, _HostMount, _Scope
 from .._remotespec import RemoteSpec, RemoteSpecPurpose
 from ..utils import UtilError
+from .inspect import Inspector
 
 
 ##################################################################
@@ -592,7 +593,8 @@ def inspect(app, elements, deps):
 
     """
     with app.initialized():
-        app.inspector.dump_to_stdout(elements, selection=deps)
+        inspector = Inspector(app.stream, app.project, app.context)
+        inspector.dump_to_stdout(elements, selection=deps)
 
 
 ##################################################################
--- a/src/buildstream/_frontend/cli.py
+++ b/src/buildstream/_frontend/cli.py
@@ -537,6 +537,9 @@ def build(
 ##################################################################
 @cli.command(name="inspect", short_help="Inspect Project Information")
 @click.option(
+    "--except", "except_", multiple=True, type=click.Path(readable=False), help="Except certain dependencies"
+)
+@click.option(
     "--deps",
     "-d",
     default=_PipelineSelection.ALL,
@@ -554,7 +557,7 @@ def build(
 )
 @click.argument("elements", nargs=-1, type=click.Path(readable=False))
 @click.pass_obj
-def inspect(app, elements, deps):
+def inspect(app, except_, elements, deps):
     """Access structured data about a given buildstream project and it's computed elements.
 
     Specifying no elements will result in showing the default targets
@@ -594,7 +597,7 @@ def inspect(app, elements, deps):
     """
     with app.initialized():
         inspector = Inspector(app.stream, app.project, app.context)
-        inspector.dump_to_stdout(elements, selection=deps)
+        inspector.dump_to_stdout(elements, except_=except_, selection=deps)
 
 
 ##################################################################
--- a/src/buildstream/_frontend/inspect.py
+++ b/src/buildstream/_frontend/inspect.py
@@ -1,16 +1,14 @@
 import json
 import sys
-
+from dataclasses import dataclass, fields, is_dataclass
 from enum import StrEnum
-from dataclasses import dataclass
-
-from ruamel.yaml import YAML
 
 from ..types import _PipelineSelection, _Scope
 
+
 # Inspectable Elements as serialized to the terminal
 @dataclass
-class _InspectElement:
+class _Element:
     name: str
     description: str
     workspace: any
@@ -25,15 +23,76 @@ class _InspectElement:
     runtime_dependencies: list[str]
     sources: list[dict[str, str]]
 
-@dataclass
-class _ProjectOutput:
-    name: str
-    directory: str
 
+# Representation of a cache server
+@dataclass
+class _CacheServer:
+    url: str
+    instance: str
+
+    def __init__(self, spec):
+        self.url = spec.url
+        self.instance = spec.instance_name
+
+
+# User configuration
+@dataclass
+class _UserConfig:
+    configuration: str
+    cache_directory: str
+    log_directory: str
+    source_directory: str
+    build_directory: str
+    source_mirrors: str
+    build_area: str
+    strict_build_plan: bool
+    cache_directory: str
+    maximum_fetch_tasks: int
+    maximum_build_tasks: int
+    maximum_push_tasks: int
+    maximum_network_retries: int
+    cache_storage_service: _CacheServer | None
+    # remote specs
+    remote_execution_service: _CacheServer | None
+    remote_storage_service: _CacheServer | None
+    remote_action_cache_service: _CacheServer | None
+
+
+# String representation of loaded plugins
+@dataclass
+class _Plugin:
+    name: str
+    full: str  # class str
+
+
+# Configuration of a given project
+@dataclass
+class _ProjectConfig:
+    name: str
+    directory: str | None
+    junction: str | None
+    variables: [(str, str)]
+    element_plugins: [_Plugin]
+    source_plugins: [_Plugin]
+
+
+# A single project loaded from the current configuration
+@dataclass
+class _Project:
+    provenance: str
+    duplicates: [str]
+    declarations: [str]
+    config: _ProjectConfig
+
+
+# Wrapper object ecapsulating the entire output of `bst inspect`
 @dataclass
 class _InspectOutput:
-    project: _ProjectOutput
-    elements: list[_InspectElement]
+    project: [_Project]
+    # user configuration
+    user_config: _UserConfig
+    elements: list[_Element]
+
 
 # Used to indicate the state of a given element
 class _ElementState(StrEnum):
@@ -59,11 +118,91 @@ class _ElementState(StrEnum):
     BUILDABLE = "buildable"
 
 
+# _make_dataclass()
+#
+# This is a helper class for extracting values from different objects used
+# across Buildstream into JSON serializable output.
+#
+# If keys is a list of str then each attribute is copied directly to the
+# dataclass.
+# If keys is a tuple of str then the first value is extracted from the object
+# and renamed to the second value.
+#
+# The field of kwarg is mapped directly onto the dataclass. If the value is
+# callable then that function is called passing the object to it.
+#
+# Args:
+#       obj: Whichever object you are serializing
+#       _cls: The dataclass you are constructing
+#       keys: attributes to include directly from the obj
+#       kwargs: key values passed into the dataclass
+def _make_dataclass(obj, _cls, keys: list[(str, str)] | list[str], **kwargs):
+    params = dict()
+    for key in keys:
+        name = None
+        rename = None
+        if isinstance(key, tuple):
+            name = key[0]
+            rename = key[1]
+        elif isinstance(key, str):
+            name = key
+            rename = None
+        else:
+            raise Exception("BUG: Keys may only be (str, str) or str")
+        value = None
+        if isinstance(obj, dict):
+            value = obj.get(name)
+        elif isinstance(obj, object):
+            try:
+                value = getattr(obj, name)
+            except AttributeError:
+                pass
+        else:
+            raise Exception("BUG: obj must be a dict or object")
+        if rename:
+            params[rename] = value
+        else:
+            params[name] = value
+    for key, helper in kwargs.items():
+        if callable(helper):
+            params[key] = helper(obj)
+        else:
+            params[key] = helper
+    return _cls(**params)
+
+
+# Recursively dump the dataclass into a serializable dictionary. Null values
+# are dropped from the output.
+def _dump_dataclass(_cls):
+    d = dict()
+    if not is_dataclass(_cls):
+        raise Exception("BUG: obj must be a dataclass")
+    for field in fields(_cls):
+        value = getattr(_cls, field.name)
+        if value is None:  # hide null values
+            continue
+        if is_dataclass(value):
+            d[field.name] = _dump_dataclass(value)
+        elif isinstance(value, list):
+            items = []
+            for item in value:
+                if is_dataclass(item):
+                    # check if it's a list of dataclasses
+                    items.append(_dump_dataclass(item))
+                else:
+                    items.append(item)
+            d[field.name] = items
+        else:
+            d[field.name] = value
+    return d
+
+
 # Inspect elements from a given Buildstream project
 class Inspector:
-    def __init__(self, stream, project):
+    def __init__(self, stream, project, context):
         self.stream = stream
         self.project = project
+        self.context = context
 
     def _read_state(self, element):
         try:
@@ -94,33 +233,6 @@ class Inspector:
     def _elements(self, dependencies, with_state=False):
         for element in dependencies:
 
-            name = element._get_full_name()
-            description = " ".join(element._description.splitlines())
-            workspace = element._get_workspace()
-            variables = dict(element._Element__variables)
-            environment = dict(element._Element__environment)
-
-            sources = []
-            for source in element.sources():
-                source_infos = source.collect_source_info()
-
-                if source_infos is not None:
-                    serialized_sources = []
-                    for s in source_infos:
-                        serialized = s.serialize()
-                        serialized_sources.append(serialized)
-
-                    sources += serialized_sources
-
-            # Show dependencies
-            dependencies = [e._get_full_name() for e in element._dependencies(_Scope.ALL, recurse=False)]
-
-            # Show build dependencies
-            build_dependencies = [e._get_full_name() for e in element._dependencies(_Scope.BUILD, recurse=False)]
-
-            # Show runtime dependencies
-            runtime_dependencies = runtime_dependencies = [e._get_full_name() for e in element._dependencies(_Scope.RUN, recurse=False)]
-
             # These operations require state and are only shown if requested
             key = None
             key_full = None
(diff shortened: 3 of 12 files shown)


src/buildstream/_frontend/inspect.py
```diff
@@ -0,0 +1,184 @@
...
+                artifact=artifact,
+                dependencies=dependencies,
+                build_dependencies=build_dependencies,
+                runtime_dependencies=runtime_dependencies,
+                sources=sources,
+            )
+
+
+    def _dump_project(self):
+        # TODO: What else do we want here?
```
kevinschoon: Please take note of this, I'm unsure what other items we should include in the project output.

tests/frontend/inspect.py
```diff
@@ -0,0 +1,61 @@
...
+    assert(source["kind"] == "local")
+    assert(source["url"] == "files/bin-files")
+
+
+@pytest.mark.datafiles(os.path.join(DATA_DIR, "simple"))
+def test_inspect_element_glob(cli, datafiles):
+    project = str(datafiles)
+    result = cli.run(project=project, silent=True, args=["inspect", "*.bst"])
+    result.assert_success()
+    json.loads(result.output)
```
shymega-ct: I would perform an assert between the result, and an 'expected' JSON result stored in the project directory.

tests/frontend/inspect.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+@pytest.mark.datafiles(os.path.join(DATA_DIR, "source-fetch"))
+def test_inspect_with_state(cli, datafiles):
+    project = str(datafiles)
+    result = cli.run(project=project, silent=True, args=["inspect", "--state", "--deps", "all"])
+    result.assert_success()
+    json.loads(result.output)
```
shymega-ct: Same comment as above.

tests/frontend/inspect.py
```diff
@@ -0,0 +1,61 @@
...
+def test_inspect_basic(cli, datafiles):
+    project = str(datafiles)
+    result = cli.run(project=project, silent=True, args=["inspect"])
+    result.assert_success()
+    output = json.loads(result.output)
+    assert(output["project"]["name"] == "test")
+    element = _element_by_name(output["elements"], "import-bin.bst")
+    source = element["sources"][0]
+    assert(source["kind"] == "local")
+    assert(source["url"] == "files/bin-files")
```
shymega-ct: suggestion: `assert` should report the reason if the assert fails.

src/buildstream/types.py
```diff
@@ -389,6 +389,40 @@ def new_from_node(cls, node: MappingNode) -> "_SourceMirror":
 
         return cls(name, aliases)
 
+# Used to indicate the state of a given element
+class _ElementState(FastEnum):
```
shymega-ct: I'm not sure about introducing another Enum just for this new functionality. Can we re-use any existing Enum for state of Elements?

harrysarson:
This PR allows you to see the version of a nested junction element. Currently this is only possible by hacking around with `.bst/staged-junctions`.

With this PR you can do something like

```
bst inspect xxxx.bst:freedesktop-sdk.bst | jq '.elements[0].sources[0]["version-guess"]'
```

Which is really nice and will be very useful.

abderrahim:
> This PR allows you to see the version of a nested junction element. Currently this is only possible by hacking around with `.bst/staged-junctions`.
> 
> With this PR you can do something like
> 
> ```
> bst inspect xxxx.bst:freedesktop-sdk.bst | jq '.elements[0].sources[0]["version-guess"]'
> ```

You should be able to get the source information already using `bst show`

```
bst show --format %{source-info} xxxx.bst:freedesktop-sdk.bst
```

Unless you're referring to a bug with the above command?

tests/frontend/inspect.py
kevinschoon: Thanks, I have cleaned up and parameterized the tests a bit in my last commit.

tests/frontend/inspect.py
kevinschoon: I looked into doing this but because the inspect format appears to contain some localized information to the caller, for example:

```json
      "sources": [
        {
          "kind": "remote",
          "url": "file:///home/kevin/repos/external/github.com/apache/buildstream/tests/frontend/source-fetch/files/bananas",
          "medium": "remote-file",
          "version-type": "sha256",
          "version": "e49295702f7da8670778e9b95a281b72b41b31cb16afa376034b45f59a18ea3f"
        }
      ]
```

It would not be possible to store the entire expected output of the inspect command. Instead of doing that I instead just tested for the existence of a few keys. If you think we should still test more though I'm open to trying a different approach.

src/buildstream/types.py
kevinschoon: The only place AFAIK that introduces the concept of "states" is in the `bst show` command as seen here: https://github.com/apache/buildstream/blob/master/src/buildstream/_frontend/widget.py#L363-L381. I think that the `_ElementState` enum could move privately into `inspect.py` though.

harrysarson:
Ah `--format %{source-info}` is new. And also works!

src/buildstream/_frontend/app.py
```diff
@@ -302,6 +304,10 @@ def initialized(self, *, session_name=None):
             #
             self.stream.set_project(self.project)
 
+            # Initialize the inspector
+            if self._session_name == "Inspect":
```
gtristan: I don't think `bst inspect` should have a session name.

The session name is a UI detail used to determine whether the print an elaborate heading about buildstream version and project details, and whether to time the overall session encapsulated in a START / SUCCESS / FAILURE block, these are more appropriate for work related commands like `bst build`, `bst source fetch`, `bst artifact push`, `bst source track` etc (i.e. *sessions*).

src/buildstream/types.py
gtristan: I also feel skeptical about this.

If we do add this, I would prefer for it to be properly integrated, and come in the form of a preceding commit which adds the functionality:

* Adds the new `_ElementState` (or perhaps `_ElementUIState` in order to emphasize that this is not to be used in the core for any decision making, but only for reporting purposes to the UI).
* Adds `Element._compute_ui_state() -> _ElementState`, such that the reporting of this state is done by the `Element` class
  * This *internal method* (i.e. one leading underscore, not class private, and using comments for documentation rather than docstrings) should document the conditions under which states will be reported. I.e. it should be clear that *cached* state will not be reported unless the cache has been initialized.
* Updates the existing code in `_frontend/widget.py` to use this function to obtain the `_ElementUIState` and do the `p.fmt_subst(line, "state", ui_state.value, color)`

Noticing that we have colors associated to states, these could probably be encoded into `_ElementUIState` in some way in order to avoid having a separate switch case around this.

tests/frontend/inspect.py
```diff
@@ -0,0 +1,96 @@
...
+    "flags,elements",
+    [
+        ([], ["import-dev.bst", "import-bin.bst", "compose-all.bst", "target.bst", "subdir/target.bst"]),
+        (["*.bst", "**/*.bst"], ["import-dev.bst", "import-bin.bst", "compose-all.bst", "target.bst", "subdir/target.bst"]),
+        (["--state"], ["import-dev.bst", "import-bin.bst", "compose-all.bst", "target.bst", "subdir/target.bst"]),
+        (["--state", "--deps", "all"], ["import-dev.bst", "import-bin.bst", "compose-all.bst", "target.bst", "subdir/target.bst"]),
+        (["subdir/*.bst"], ["import-dev.bst", "subdir/target.bst"])
+    ],
+)
+@pytest.mark.datafiles(os.path.join(DATA_DIR, "simple"))
```
gtristan: Please do not reuse data from other tests.

Instead, create your own subdirectory with example data to practice/test with.

src/buildstream/_frontend/inspect.py
```diff
@@ -0,0 +1,184 @@
...
+    environment: dict[str, str]
+    variables: dict[str, str]
+    artifact: any
+    dependencies: list[str]
+    build_dependencies: list[str]
+    runtime_dependencies: list[str]
+    sources: list[dict[str, str]]
+
+@dataclass
+class _ProjectOutput:
```
gtristan: A lot of project related data missing here, such as which plugins have been loaded and their provenances, which cache servers were accessed, the provenance string of the project itself (i.e. which junction it was loaded from and where).

All of which might not be reported by default.

I think we should cover everything that is reported in `LogLine.print_heading()` - even if some of this data may be conditional (i.e. user config and access to caches may dictate which artifact/source remotes are active).

src/buildstream/types.py
```diff
@@ -389,6 +389,40 @@ def new_from_node(cls, node: MappingNode) -> "_SourceMirror":
...
+
+    # The element my be built
+    BUILDABLE = "buildable"
+
+    def __str__(self):
+        return str(self.value)
+
+
+# The type of encoding used when outputing machine readable information
+class _Encoding(FastEnum):
```
gtristan: Given this is a completely new command, I'm not sure we really need YAML formatting support here.

src/buildstream/types.py
kevinschoon: I've simply placed `_ElementState` as a private class in `inspect.py` such that it is an implementation detail. Since there is nothing else using this concept now I think it's better to finish the command and then later we can decide if it can be used elsewhere.

src/buildstream/_frontend/app.py
kevinschoon: I've eliminated the use of  `_session_name`.

src/buildstream/types.py
kevinschoon: I've eliminated the `_Encoding` class entirely and now we only output JSON.

tests/frontend/inspect.py
kevinschoon: I've included a new sub-directory with some basic test data.

src/buildstream/types.py
gtristan: Thats fine with me.

While it would be good to code share this switch case with the other one in `_frontend/widget.py`, I'm happy that this at least lives exclusively in the frontend (view/controller) rather than in the core (data model).

src/buildstream/types.py
gtristan: worksforme :)

src/buildstream/_frontend/app.py
```diff
@@ -302,6 +304,9 @@ def initialized(self, *, session_name=None):
             #
             self.stream.set_project(self.project)
 
+            # Initialize the inspector
+            self.inspector = Inspector(self.stream, self.project, self.context)
```
gtristan: This is not needed for anything outside of the `bst inspect` command, as such I'd rather this be created directly in `_frontend/cli.py` within the `app.initialized()` block.

src/buildstream/_frontend/inspect.py
```diff
@@ -0,0 +1,397 @@
...
+                # BUG: Due to the assersion within .get_artifact this will
+                # error but there is no other way to determine if an artifact
+                # exists and we only want to show this value for informational
+                # purposes.
+                try:
+                    _artifact = element._get_artifact()
+                    if _artifact.cached():
+                        artifact = {
+                            "files": artifact.get_files(),
+                            "digest": artifact_files._get_digest(),
```
gtristan: Looks untested, the `artifact_files` variable appears to be undeclared here.

Also `artifact.get_files()` reports a `CasBasedDirectory`, perhaps this patch is expecting the `__iter__` implementation to serialize into a list of paths ?

Not sure what's going on overall in this code block, but if we want to report files, we need more structured data here, such that we can report file types, symlink targets, file sizes, permissions, etc.

That said; I think it is acceptable to not support reporting artifact files for an initial version of `bst inspect`.

src/buildstream/_frontend/cli.py
```diff
@@ -531,6 +531,79 @@ def build(
...
+            _PipelineSelection.RUN,
+            _PipelineSelection.BUILD,
+            _PipelineSelection.ALL,
+        ],
+    ),
+    help="The dependencies to show",
+)
+@click.argument("elements", nargs=-1, type=click.Path(readable=False))
+@click.pass_obj
+def inspect(app, elements, state, deps):
```
gtristan: As with `bst show`, we should also support support the `--except` options here.

src/buildstream/_frontend/inspect.py
```diff
@@ -0,0 +1,397 @@
...
+            artifact = None
+
+            if with_state:
+                key = element._get_display_key().brief
+
+                key_full = element._get_display_key().full
+
+                state = self._read_state(element).value
+
+                # BUG: Due to the assersion within .get_artifact this will
```
gtristan: I believe the correct way to go about this is `Element._load_artifact()`.

I think we can avoid `pull` for now and consider whether `bst inspect` pulls data from remotes as a possible followup - and ` strict` comes from the `Context` object here which will be resolved by the toplevel options (defaults to `True` unless `bst --no-strict inspect ....` is specified).

gtristan:
Overall this is coming along nicely... some things which come to mind:

* We need to consider how this will eventually evolve to support project-less use
  * I.e. there is a certain amount of information which is serialized into artifact metadata, that is extensible and can grow, and `bst inspect` should (at least eventually) allow inspection of artifact dependency chains loaded from a remote cache, without needing to have a local checkout of the project
  * This means we need to consider whether we intend to support this directly in `bst inspect`, or if we intend to have that as a separate addition as `bst artifact inspect`... perhaps the latter seems more sensible
* We still do not have anything reporting of public data here as far as I can see
  * Public data is special as it is *mutable*
  * We need to consider what is sensible to report here
  * Initially, my feeling is that we should support reporting both original public data and mutated public data separately
    * This might require some added features to the core
    * Displaying original public data is only possible if we have a local project (not loading from a remote artifact)
    * Displaying mutated public data is only possible if the element has a locally cached artifact
* Looks like this is still missing the ability to show artifact build logs
  * Perhaps not necessary for an initial implementation but shouldn't be too hard to include
  * Definitely should not be included in output by default (only if specified on the command line *--fields*)

src/buildstream/_frontend/inspect.py
kevinschoon: Sorry I'm not sure I fully understand this. Are you saying we should avoid pulling the state entirely with `bst inspect` (for this PR anyway)?

gtristan: I think what you are misunderstanding is that there is a difference between the local and remote caches.

loading artifacts from the local CAS is expensive, so is checking the cached state of artifacts. That is why we avoid loading cached state where possible.

In this comment, we’re talking about loading the artifact to check its files and CAS digest, we can do that if the artifact is in the local cache.

Some bst commands also allow pulling from remotes, in case the artifacts were built in CI or pushed by another user, it can be useful, but I don’t think having `bst inspect` support downloading artifacts from remotes is important for an initial implementation.

For this case, you should follow the code paths for `bst artifact list-contents` in order to see how artifact loading is done… and follow the `bst show` code path for displaying `%{artifact-cas-digest}` to observe how to get the digest.

It is acceptable to just omit the data for these in the case that the local artifact is not cached and is thus unavailable.

gtristan:
Suggestion: I think we should have a whole `”artifact”: { … }` block which contains only things which are loaded from the artifact, and omit it entirely in the case that the artifact is not present.

This can include:

* the public data as found in the artifact, which may not be the same as the public data loaded from the bst files before building
* The cas digest(s)
* The file metadata 
* The build logs
* The failure state (whether it is a cached failed build)
* whether the artifact has a cached build tree

I think this nested structure disambiguates things, solves the public data concern nicely and is all around nicely structured.

Also… I think that given we present a lot of data in the output json, I think maybe we should not do the `%{state}` thing at all, and leave that only to `bst show`. Rationale being that the “state” concept is really very user facing, and it can anyway be easily computed based on the json reported by `bst inspect`

kevinschoon:
> Also… I think that given we present a lot of data in the output json, I think maybe we should not do the `%{state}` thing at all, and leave that only to `bst show`. Rationale being that the “state” concept is really very user facing, and it can anyway be easily computed based on the json reported by `bst inspect`

I have eliminated state entirely from the output.

src/buildstream/_frontend/inspect.py
kevinschoon: > A lot of project related data missing here, 


> such as which plugins have been loaded and their provenances, 


Plugins are now exposed, example:

```json
...
 "plugins": [
	{
	  "name": "stack",
	  "description": "core plugin",
	  "plugin_type": "element"
	},
	{
	  "name": "import",
	  "description": "core plugin",
	  "plugin_type": "element"
	},
  ...
]
```
> which cache servers were accessed

All remote state including cache servers it not included in inspect output currently.

> the provenance string of the project itself (i.e. which junction it was loaded from and where).

Each project loaded shows the junction from where it comes, example:

```json
...
 "junction": "freedesktop-sdk.bst:plugins/buildstream-plugins.bst",
...
```

src/buildstream/_frontend/app.py
kevinschoon: This has been moved from `app.py -> cli.py`.

src/buildstream/_frontend/inspect.py
kevinschoon: I've eliminated all of the artifact inspection for this first version.

src/buildstream/_frontend/cli.py
kevinschoon: I've added support for `--except` in the same way `bst show` works.

kevinschoon:
Here is an example of the current state of the inspect output against a test project.

Any insight into specifically what types of fields we would like to include in this output as well as general feedback would be appreciated.

```json
{
  "project": [
    {
      "duplicates": [],
      "declarations": [],
      "config": {
        "name": "test",
        "directory": "/home/kevin/repos/external/github.com/apache/buildstream/tests/frontend/inspect",
        "aliases": {
          "example": "https://example.org/"
        },
        "element_overrides": {},
        "source_overrides": {},
        "plugins": [
          {
            "name": "stack",
            "description": "core plugin",
            "plugin_type": "element"
          },
          {
            "name": "import",
            "description": "core plugin",
            "plugin_type": "element"
          },
          {
            "name": "local",
            "description": "core plugin",
            "plugin_type": "source"
          },
          {
            "name": "remote",
            "description": "core plugin",
            "plugin_type": "source"
          },
          {
            "name": "tar",
            "description": "core plugin",
            "plugin_type": "source"
          },
          {
            "name": "local",
            "description": "core plugin",
            "plugin_type": "source-mirror"
          },
          {
            "name": "remote",
            "description": "core plugin",
            "plugin_type": "source-mirror"
          },
          {
            "name": "tar",
            "description": "core plugin",
            "plugin_type": "source-mirror"
          }
        ]
      }
    }
  ],
  "user_config": {
    "configuration": "/home/kevin/.config/buildstream.conf",
    "cache_directory": "/home/kevin/.cache/buildstream",
    "log_directory": "/home/kevin/.cache/buildstream/logs",
    "source_directory": "/home/kevin/.cache/buildstream/sources",
    "build_directory": "/home/kevin/.cache/buildstream/build",
    "source_mirrors": "/home/kevin/.cache/buildstream/sources",
    "build_area": "/home/kevin/.cache/buildstream/build",
    "strict_build_plan": "/home/kevin/.config/buildstream.conf",
    "maximum_fetch_tasks": 10,
    "maximum_build_tasks": 4,
    "maximum_push_tasks": 4,
    "maximum_network_retries": 2
  },
  "elements": [
    {
      "name": "import-local-files.bst",
      "description": "",
      "environment": {
        "PATH": "/usr/bin:/bin:/usr/sbin:/sbin",
        "SHELL": "/bin/sh",
        "TERM": "dumb",
        "USER": "tomjon",
        "USERNAME": "tomjon",
        "LOGNAME": "tomjon",
        "LC_ALL": "C",
        "HOME": "/tmp",
        "TZ": "UTC",
        "SOURCE_DATE_EPOCH": "1321009871"
      },
      "variables": {
        "prefix": "/usr",
        "exec_prefix": "/usr",
        "bindir": "/usr/bin",
        "sbindir": "/usr/sbin",
        "libexecdir": "/usr/libexec",
        "datadir": "/usr/share",
        "sysconfdir": "/etc",
        "sharedstatedir": "/usr/com",
        "localstatedir": "/var",
        "lib": "lib",
        "libdir": "/usr/lib",
        "debugdir": "/usr/lib/debug",
        "includedir": "/usr/include",
        "docdir": "/usr/share/doc",
        "infodir": "/usr/share/info",
        "mandir": "/usr/share/man",
        "build-root": "/buildstream/test/import-local-files.bst",
        "conf-root": ".",
        "install-root": "/buildstream-install",
        "strip-binaries": "",
        "schema": "https",
        "project-name": "test",
        "max-jobs": "8",
        "element-name": "import-local-files.bst"
      },
      "dependencies": [],
      "build_dependencies": [],
      "runtime_dependencies": [],
      "sources": [
        {
          "kind": "local",
          "url": "files",
          "medium": "local",
          "version-type": "cas-digest",
          "version": "d8c20623d7160ffe2fd69bd03b3ad7b24a6d1dfd7c96f3ed6c8deb3f268d2d64/85"
        }
      ]
    },
    {
      "name": "import-remote-files.bst",
      "description": "",
      "environment": {
        "PATH": "/usr/bin:/bin:/usr/sbin:/sbin",
        "SHELL": "/bin/sh",
        "TERM": "dumb",
        "USER": "tomjon",
        "USERNAME": "tomjon",
        "LOGNAME": "tomjon",
        "LC_ALL": "C",
        "HOME": "/tmp",
        "TZ": "UTC",
        "SOURCE_DATE_EPOCH": "1321009871"
      },
      "variables": {
        "prefix": "/usr",
        "exec_prefix": "/usr",
        "bindir": "/usr/bin",
        "sbindir": "/usr/sbin",
        "libexecdir": "/usr/libexec",
        "datadir": "/usr/share",
        "sysconfdir": "/etc",
        "sharedstatedir": "/usr/com",
        "localstatedir": "/var",
        "lib": "lib",
        "libdir": "/usr/lib",
        "debugdir": "/usr/lib/debug",
        "includedir": "/usr/include",
        "docdir": "/usr/share/doc",
        "infodir": "/usr/share/info",
        "mandir": "/usr/share/man",
        "build-root": "/buildstream/test/import-remote-files.bst",
        "conf-root": ".",
        "install-root": "/buildstream-install",
        "strip-binaries": "",
        "schema": "https",
        "project-name": "test",
        "max-jobs": "8",
        "element-name": "import-remote-files.bst"
      },
      "dependencies": [],
      "build_dependencies": [],
      "runtime_dependencies": [],
      "sources": [
        {
          "kind": "remote",
          "url": "https://example.org/foo.bar.bin",
          "medium": "remote-file",
          "version-type": "sha256",
          "version": "d1bc8d3ba4afc7e109612cb73acbdddac052c93025aa1f82942edabb7deb82a1"
        },
        {
          "kind": "tar",
          "url": "https://example.org/baz.qux.tar.gz",
          "medium": "remote-file",
          "version-type": "sha256",
          "version": "d1bc8d3ba4afc7e109612cb73acbdddac052c93025aa1f82942edabb7deb82a1"
        }
      ]
    },
    {
      "name": "target.bst",
      "description": "Main stack target for the bst build test",
      "environment": {
        "PATH": "/usr/bin:/bin:/usr/sbin:/sbin",
        "SHELL": "/bin/sh",
        "TERM": "dumb",
        "USER": "tomjon",
        "USERNAME": "tomjon",
        "LOGNAME": "tomjon",
        "LC_ALL": "C",
        "HOME": "/tmp",
        "TZ": "UTC",
        "SOURCE_DATE_EPOCH": "1321009871"
      },
      "variables": {
        "prefix": "/usr",
        "exec_prefix": "/usr",
        "bindir": "/usr/bin",
        "sbindir": "/usr/sbin",
        "libexecdir": "/usr/libexec",
        "datadir": "/usr/share",
        "sysconfdir": "/etc",
        "sharedstatedir": "/usr/com",
        "localstatedir": "/var",
        "lib": "lib",
        "libdir": "/usr/lib",
        "debugdir": "/usr/lib/debug",
        "includedir": "/usr/include",
        "docdir": "/usr/share/doc",
        "infodir": "/usr/share/info",
        "mandir": "/usr/share/man",
        "build-root": "/buildstream/test/target.bst",
        "conf-root": ".",
        "install-root": "/buildstream-install",
        "strip-binaries": "",
        "schema": "https",
        "project-name": "test",
        "max-jobs": "8",
        "element-name": "target.bst"
      },
      "dependencies": [
        "import-local-files.bst",
        "import-remote-files.bst"
      ],
      "build_dependencies": [
        "import-local-files.bst",
        "import-remote-files.bst"
      ],
      "runtime_dependencies": [
        "import-local-files.bst",
        "import-remote-files.bst"
      ],
      "sources": []
    }
  ]
}

```

gtristan:
> Here is an example of the current state of the inspect output against a test project.
> 
> Any insight into specifically what types of fields we would like to include in this output as well as general feedback would be appreciated.
> 

Here are my thoughts for today...

## Connecting the dots

Something I that we'll absolutely need is a way for scripts to piece together
the data in interesting ways, this means we need reliable ways to make references
to other data in the JSON output.

For instance, when printing a dependency - we need a reliable way to uniquely
address that dependency in the accompanying data.


### Element references (dependencies)

How to uniquely address an element dependency in the output may depend on other
things.

Currently with `bst show` output, we print the full element paths (including
the junction prefixes) which would allow one to make the connection.

In your comment you are proposing:

```yaml
      "dependencies": [
        "import-local-files.bst",
        "import-remote-files.bst"
      ],
      "build_dependencies": [
        "import-local-files.bst",
        "import-remote-files.bst"
      ],
      "runtime_dependencies": [
        "import-local-files.bst",
        "import-remote-files.bst"
      ],
```

I think it would be better to have one `"dependencies"` list following the
[full dependency dictionary](https://docs.buildstream.build/2.5/format_declaring.html#expressing-dependencies)

E.g.:

```
      "dependencies" : [
        {
          "name": "foo.bst",
          "junction": "junction.bst"
          "type" : "dependency type"
        },
      ]
```

Where `"type"` can be any of the [valid dependency types](https://docs.buildstream.build/2.5/format_declaring.html#format-dependencies-types).

This would mean that for a dependency in the toplevel project, the "junction" entry
would be empty, if the element is in a sub-subproject, the junction would be a junction
path, such as `"subproject.bst:subsubproject.bst"`.


### Project junction

In order to piece together the data, we should have a special `"junction"` parameter
which is serialized as a regular element, serializing data from the junction element
directly.

In this way, the junction's `"name"` will be a junction element present in the parent
project which the project was loaded from.

Also, this is very helpful as it adds the ability to provide a
[SourceInfo](https://docs.buildstream.build/2.5/buildstream.source.html#buildstream.source.SourceInfo)
list for the junction itself.


### Project elements

I think that the elements which were loaded from a given project, should be found
in a nested `"elements"` list defined within the `"project"`.

E.g.:

```yaml
{
  "projects" : [
    {
      "name": "freedesktop-sdk",
      "junction" : {
        "freedesktop-sdk.bst",
        "source-info" : [
           ...
        ]
      },

      # Serialize the elements from this project, which are loaded in this pipeline
      "elements" : [
         ...
      ]
    }
  ],
  ...
}
```

## Lets start smaller and build on that

There is a bunch of stuff here which we don't know why it might be useful, lets
reduce this dramatically and iteravely add aspects to the serialized data with
careful thought, depending on what it might be useful for.


### User config

I think we should just drop the `"user_config"` stuff, it is unclear whether
any of this will be useful.

Also, it doesn't seem to make sense regarding the BuildStream data
model to dump data about "project config" and "user config" separately.

The finalized data model is an amalgamation of defaults, user configuration,
project configuration and element specific information, and I am skeptical
about serializing/framing the loaded data model in terms of the input
which BuildStream uses to load the data model.

In any case, I suggest we just drop all of this for now.


### Project level things


* `"duplicates"` and `"declarations"`

  I don't think we need `"duplicates"` or `"declarations"` here, it is not clear
  what you propose `"declarations"` to actually mean, either.

  About duplicates: Given my previous comment about *Connecting the dots* above, I do not
  think it is relevant to serialize how a project configures this; instead we focus on how
  the data model was constructed.

  In the case that the same project was loaded twice in the pipeline, they will be
  distinguished by their `"junction"`, and any dependencies on an element that is loaded
  twice in the pipeline, will be able to distinguish *which* `"foo.bst"` was being
  referred to, by way of providing strong references to elements, as described
  in my previous comment about *Element references*.

* `"element_overrides"` and `"source_overrides"`

  I don't know what you intended here, but in either of my interpretations, lets
  drop these.

  If we are talking about [overriding elements](https://docs.buildstream.build/2.5/elements/junction.html#overriding-elements)
  in subprojects via junctions, then lets ignore this detail.

  The resolved data model will show which element was used from which project, even in the
  case that a subproject ends up depending on an element in it's parent project, which
  depends on elements in the parent and subproject.

  I think instead you are referring to
  [overriding plugin configuration](https://docs.buildstream.build/2.5/format_project.html#overriding-plugin-defaults)

  There is no reason to include this in the serialized json, as it is not a part of the
  resolved data model. These dictionaries are a part of the
  [element composition](https://docs.buildstream.build/2.5/format_intro.html#format-composition),
  and we are only concerned with providing the end resulting `variables`, `environment` etc in
  the reported elements.

* `"config"` dictionary

  I don't think we need any nesting here, I think we can assume that the toplevel
  `"project"` dictionary is *stuff about the project*.


* `"directory"`

  I don't think we need this, because it is only contextual to where the project itself
  is checked out and built - I don't see how this is useful for a scripting interface.

* `"aliases"`

  Again this is input data to the data model, but not relevant I think.

  Ultimately relevant URL results will be reported in the source URLs.

  Later we may consider doing something fancy to report possible mirror URLs for
  a given laoded project/user config - allowing us to print which URLs will be
  traversed in order for a given source - but nobody is asking for this now
  and we don't need it in an initial revision.

## Plugin serialization

We need to consider some different useful things for each plugin origin.

We have local plugins, pip loaded plugins, and plugins loaded through junctions.

I think it will make sense to have separate lists here within a `"project"`
dictionary, specifying more precisely where a plugin came from.

Similarly to my suggestion about the project junction, we should also
serialize the junction from whence the plugins from that origin were
loaded in the same way (so we can have the SourceInfo describing where
plugins came from).

While I feel strongly that fully descriptive data needs to be provided
about loaded plugins, I do not think that serializing the project's plugins
in the `bst inspect` output is a hard blocker for an initial version.


## What we need and don't have

* We definitely want the cache keys here.

  For each element, at least the equivalent of "%{full-key}" from
  `bst show`, or empty if the element or it's dependencies are in
  an *inconsistent state* (inconsistent state means that the element
  or it's dependencies are missing a resolved, specific source reference).

* Element description

  This is easy, and can be useful for generating nice reports

* Artifact cas digest

  I'm on the fence about whether we need to have this in an initial
  revision, not having it is a step down from `bst show`.

  If we add this, we should have it in a nested `"artifact"` block
  found on the element dictionary which is enumerated on a per
  project basis.

  E.g.:

  ```yaml
  {
    "projects" : [
      ...
      "elements" : [
        {
          "artifact" : {
            "cas-digest" : "<digest value>"
          }
        }
      ]
    ]
  }

  This will allow extensibility for other things from the artifact
  which we'll want to add later, like files metadata, build logs, etc.

juergbi:
> ## Connecting the dots
> 
> Something I that we'll absolutely need is a way for scripts to piece together the data in interesting ways, this means we need reliable ways to make references to other data in the JSON output.

I fully agree. In addition to making sure that dependencies are fully qualified, also the element definitions need to be fully qualified.

And I think we want to make sure that they use the same structure. As far as I can tell, this branch currently uses `_get_full_name()` in both places, which results in qualified strings such as `foo.bst:bar.bst` when junctions are involved. This seems perfectly reasonable to me.

If we want to extract the junction part to a separate JSON member, we should do it also in element definitions for consistency. However, I suspect that the current approach with a single colon-separated string will be more convenient in practice, as there will be no need to do any string manipulation just for 'connecting the dots'. And it also matches `bst show`.

kevinschoon:
> I think we should just drop the "user_config" stuff, it is unclear whether any of this will be useful.

I've dropped all of the user config.

kevinschoon:
I've re-written the `dependencies` section of the elements to now output their names and junctions as you described.

Example:

```json
   ....
    {
      "name": "bootstrap/base-sdk/automake.bst",
      "junction": "freedesktop-sdk.bst",
      "kind": "all"
    },
    {
      "name": "bootstrap/build/rsync.bst",
      "junction": "freedesktop-sdk.bst",
      "kind": "all"
    },
  ...
```

Non-junction output:

```json
  "dependencies": [
    {
      "name": "import-local-files.bst",
      "kind": "all"
    },
    {
      "name": "import-remote-files.bst",
      "kind": "all"
    },
    {
      "name": "target.bst",
      "kind": "all"
    }
  ]
```

kevinschoon:
I've continued to make updates based on your feedback although this is still a WIP please see an updated project output:

```json
{
  "projects": [
    {
      "name": "test",
      "plugins": [
        {
          "name": "stack",
          "description": "core plugin",
          "plugin_type": "element"
        },
        {
          "name": "import",
          "description": "core plugin",
          "plugin_type": "element"
        },
        {
          "name": "local",
          "description": "core plugin",
          "plugin_type": "source"
        },
        {
          "name": "remote",
          "description": "core plugin",
          "plugin_type": "source"
        },
        {
          "name": "tar",
          "description": "core plugin",
          "plugin_type": "source"
        }
      ],
      "elements": [
        {
          "name": "import-local-files.bst",
          "description": "",
          "environment": {},
          "variables": {
            "prefix": "/usr",
            "exec_prefix": "/usr",
            "bindir": "/usr/bin",
            "sbindir": "/usr/sbin",
            "libexecdir": "/usr/libexec",
            "datadir": "/usr/share",
            "sysconfdir": "/etc",
            "sharedstatedir": "/usr/com",
            "localstatedir": "/var",
            "lib": "lib",
            "libdir": "/usr/lib",
            "debugdir": "/usr/lib/debug",
            "includedir": "/usr/include",
            "docdir": "/usr/share/doc",
            "infodir": "/usr/share/info",
            "mandir": "/usr/share/man",
            "build-root": "/buildstream/test/import-local-files.bst",
            "conf-root": ".",
            "install-root": "/buildstream-install",
            "strip-binaries": "",
            "schema": "https",
            "project-name": "test",
            "max-jobs": "8",
            "element-name": "import-local-files.bst"
          },
          "dependencies": [
            {
              "name": "import-local-files.bst",
              "kind": "all"
            }
          ],
          "sources": [
            {
              "kind": "local",
              "url": "files",
              "medium": "local",
              "version-type": "cas-digest",
              "version": "d8c20623d7160ffe2fd69bd03b3ad7b24a6d1dfd7c96f3ed6c8deb3f268d2d64/85"
            }
          ]
        },
        {
          "name": "import-remote-files.bst",
          "description": "",
          "environment": {},
          "variables": {
            "prefix": "/usr",
            "exec_prefix": "/usr",
            "bindir": "/usr/bin",
            "sbindir": "/usr/sbin",
            "libexecdir": "/usr/libexec",
            "datadir": "/usr/share",
            "sysconfdir": "/etc",
            "sharedstatedir": "/usr/com",
            "localstatedir": "/var",
            "lib": "lib",
            "libdir": "/usr/lib",
            "debugdir": "/usr/lib/debug",
            "includedir": "/usr/include",
            "docdir": "/usr/share/doc",
            "infodir": "/usr/share/info",
            "mandir": "/usr/share/man",
            "build-root": "/buildstream/test/import-remote-files.bst",
            "conf-root": ".",
            "install-root": "/buildstream-install",
            "strip-binaries": "",
            "schema": "https",
            "project-name": "test",
            "max-jobs": "8",
            "element-name": "import-remote-files.bst"
          },
          "dependencies": [
            {
              "name": "import-remote-files.bst",
              "kind": "all"
            }
          ],
          "sources": [
            {
              "kind": "remote",
              "url": "https://example.org/foo.bar.bin",
              "medium": "remote-file",
              "version-type": "sha256",
              "version": "d1bc8d3ba4afc7e109612cb73acbdddac052c93025aa1f82942edabb7deb82a1"
            },
            {
              "kind": "tar",
              "url": "https://example.org/baz.qux.tar.gz",
              "medium": "remote-file",
              "version-type": "sha256",
              "version": "d1bc8d3ba4afc7e109612cb73acbdddac052c93025aa1f82942edabb7deb82a1"
            }
          ]
        },
        {
          "name": "target.bst",
          "description": "Main stack target for the bst build test",
          "environment": {},
          "variables": {
            "prefix": "/usr",
            "exec_prefix": "/usr",
            "bindir": "/usr/bin",
            "sbindir": "/usr/sbin",
            "libexecdir": "/usr/libexec",
            "datadir": "/usr/share",
            "sysconfdir": "/etc",
            "sharedstatedir": "/usr/com",
            "localstatedir": "/var",
            "lib": "lib",
            "libdir": "/usr/lib",
            "debugdir": "/usr/lib/debug",
            "includedir": "/usr/include",
            "docdir": "/usr/share/doc",
            "infodir": "/usr/share/info",
            "mandir": "/usr/share/man",
            "build-root": "/buildstream/test/target.bst",
            "conf-root": ".",
            "install-root": "/buildstream-install",
            "strip-binaries": "",
            "schema": "https",
            "project-name": "test",
            "max-jobs": "8",
            "element-name": "target.bst"
          },
          "dependencies": [
            {
              "name": "import-local-files.bst",
              "kind": "all"
            },
            {
              "name": "import-remote-files.bst",
              "kind": "all"
            },
            {
              "name": "target.bst",
              "kind": "all"
            }
          ],
          "sources": []
        }
      ]
    }
  ],
  "defaults": {
    "environment": {
      "PATH": "/usr/bin:/bin:/usr/sbin:/sbin",
      "SHELL": "/bin/sh",
      "TERM": "dumb",
      "USER": "tomjon",
      "USERNAME": "tomjon",
      "LOGNAME": "tomjon",
      "LC_ALL": "C",
      "HOME": "/tmp",
      "TZ": "UTC",
      "SOURCE_DATE_EPOCH": "1321009871"
    }
  }
}
```

kevinschoon:
One thing to note: I added a top level `defaults` field which contains the default environment loaded per the default configuration so we don't have to duplicate redundant information within. Happy to remove this if you think the previous duplicated way is still better though.

gtristan:
> One thing to note: I added a top level `defaults` field which contains the default environment loaded per the default configuration so we don't have to duplicate redundant information within. Happy to remove this if you think the previous duplicated way is still better though.

It is interesting to consider the optimization this might provide for external tooling which wants to consider the env vars of many elements yes.

I wouldn’t want to frame such an optimization in terms of project defaults though, as this would imply the user repeats the composition algorithm, which is a bit more complex than just elements overriding project settings.

Also, we need to construct data which can equally be obtained by loading a local buildstream project, or, by downloading artifacts, so we really want to be dealing with only resolved element data which can potentially be encoded into artifact metadata (not all data is currently included in artifact metadata, but that is an interesting thing to enrich).

Perhaps there is some semantic we could have to reduce redundant data in the output, especially when the requested output fields include things which typically repeat (like env vars), I think it would be also fine to punt that optimization down the road, though.

gtristan:
> I've continued to make updates based on your feedback although this is still a WIP please see an updated project output:
> 

This is great progress and already looking much better !

> ```json
> {
>   "projects": [
>     {
>       "name": "test",
>       "plugins": [
>         {
>           "name": "stack",
>           "description": "core plugin",
>           "plugin_type": "element"
>         },
> ```

For the plugins, we will need to nest this into individual origins.

This way we can see an origin defined with its junction or its pip package and version, and see which plugins were loaded from that origin.

Also I think it will be better to list the different plugin types as lists, I.e separate lists for elements, sources, and source mirrors plugins, within a plugin origin block. This might be more a matter of taste, but would seem to avoid many redundant “plugin_type” members.

Finally, I think we can drop “description” from plugins, as I think that is inspired by an unstructured string intended for user comprehension coming from `_frontend/widget.py`, defining the origins in a more structured way will anyway be more useful for introspection and interoperability purposes.

abderrahim:
I gave this a try and it crashed with the following traceback. (Yes, I do have an open workspace).

```
[--:--:--][        ][    main:core activity                 ] BUG     Object of type Workspace is not JSON serializable

      File "/home/abderrahimkitouni/.local/bin/bst", line 10, in <module>
        sys.exit(cli())
                 ~~~^^
      File "/home/abderrahimkitouni/.local/share/uv/tools/buildstream/lib/python3.13/site-packages/click/core.py", line 1128, in __call__
        return self.main(*args, **kwargs)
               ~~~~~~~~~^^^^^^^^^^^^^^^^^
      File "/home/abderrahimkitouni/.local/share/uv/tools/buildstream/lib/python3.13/site-packages/buildstream/_frontend/cli.py", line 273, in override_main
        original_main(self, args=args, prog_name=prog_name, complete_var=None, standalone_mode=standalone_mode, **extra)
        ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/home/abderrahimkitouni/.local/share/uv/tools/buildstream/lib/python3.13/site-packages/click/core.py", line 1053, in main
        rv = self.invoke(ctx)
      File "/home/abderrahimkitouni/.local/share/uv/tools/buildstream/lib/python3.13/site-packages/click/core.py", line 1659, in invoke
        return _process_result(sub_ctx.command.invoke(sub_ctx))
                               ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
      File "/home/abderrahimkitouni/.local/share/uv/tools/buildstream/lib/python3.13/site-packages/click/core.py", line 1395, in invoke
        return ctx.invoke(self.callback, **ctx.params)
               ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/home/abderrahimkitouni/.local/share/uv/tools/buildstream/lib/python3.13/site-packages/click/core.py", line 754, in invoke
        return __callback(*args, **kwargs)
      File "/home/abderrahimkitouni/.local/share/uv/tools/buildstream/lib/python3.13/site-packages/click/decorators.py", line 38, in new_func
        return f(get_current_context().obj, *args, **kwargs)
      File "/home/abderrahimkitouni/.local/share/uv/tools/buildstream/lib/python3.13/site-packages/buildstream/_frontend/cli.py", line 600, in inspect
        inspector.dump_to_stdout(elements, except_=except_, selection=deps)
        ~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/home/abderrahimkitouni/.local/share/uv/tools/buildstream/lib/python3.13/site-packages/buildstream/_frontend/inspect.py", line 251, in dump_to_stdout
        json.dump(_dump_dataclass(output), sys.stdout)
        ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/usr/lib/python3.13/json/__init__.py", line 179, in dump
        for chunk in iterable:
                     ^^^^^^^^
      File "/usr/lib/python3.13/json/encoder.py", line 433, in _iterencode
        yield from _iterencode_dict(o, _current_indent_level)
      File "/usr/lib/python3.13/json/encoder.py", line 407, in _iterencode_dict
        yield from chunks
      File "/usr/lib/python3.13/json/encoder.py", line 326, in _iterencode_list
        yield from chunks
      File "/usr/lib/python3.13/json/encoder.py", line 407, in _iterencode_dict
        yield from chunks
      File "/usr/lib/python3.13/json/encoder.py", line 326, in _iterencode_list
        yield from chunks
      File "/usr/lib/python3.13/json/encoder.py", line 407, in _iterencode_dict
        yield from chunks
      File "/usr/lib/python3.13/json/encoder.py", line 440, in _iterencode
        o = _default(o)
      File "/usr/lib/python3.13/json/encoder.py", line 180, in default
        raise TypeError(f'Object of type {o.__class__.__name__} '
                        f'is not JSON serializable')
    TypeError: Object of type Workspace is not JSON serializable
```

abderrahim:
Another thing I noticed is that all elements are included under all projects, even the plugins project that has no elements of its own.

PR #2036, opened 2025-07-10
Add the digest-environment dependency configuration

abderrahim:
This allows setting an environment variable inside the sandbox to the CAS digest of one or more dependencies.

This us useful when communicating with the nested remote execution endpoint, but could also be used independently.

This PR replaced #1991.

### DIFF
--- a/src/buildstream/_artifact.py
+++ b/src/buildstream/_artifact.py
@@ -200,6 +200,7 @@ class Artifact:
     #    variables (Variables): The element's Variables
     #    environment (dict): dict of the element's environment variables
     #    sandboxconfig (SandboxConfig): The element's SandboxConfig
+    #    buildsandbox (Sandbox): The element's configured build sandbox
     #
     def cache(
         self,
@@ -213,6 +214,7 @@ class Artifact:
         variables,
         environment,
         sandboxconfig,
+        buildsandbox,
     ):
 
         context = self._context
@@ -316,6 +318,19 @@ class Artifact:
             rootvdir = CasBasedDirectory(cas_cache=self._cas)
             rootvdir._import_files_internal(buildrootvdir, properties=properties, collect_result=False)
             artifact.buildroot.CopyFrom(rootvdir._get_digest())
+
+        if buildsandbox is not None:
+            sandbox_env = buildsandbox._get_configured_environment()
+            if sandbox_env:
+                for key, value in sorted(sandbox_env.items()):
+                    artifact.buildsandbox.environment.add(name=key, value=value)
+
+            artifact.buildsandbox.working_directory = buildsandbox._get_work_directory()
+
+            for subsandbox in buildsandbox._get_subsandboxes():
+                vdir = subsandbox.get_virtual_directory()
+                digest = artifact.buildsandbox.subsandbox_digests.add()
+                digest.CopyFrom(vdir._get_digest())
 
         os.makedirs(os.path.dirname(os.path.join(self._artifactdir, element.get_artifact_name())), exist_ok=True)
         keys = utils._deduplicate([self._cache_key, self._weak_cache_key])
@@ -680,6 +695,21 @@ class Artifact:
             artifacts.link_key(self._element, pull_key, key)
 
         return True
+
+    def configure_sandbox(self, sandbox):
+        artifact = self._get_proto()
+
+        if artifact.buildsandbox and artifact.buildsandbox.environment:
+            env = {}
+            for env_var in artifact.buildsandbox.environment:
+                env[env_var.name] = env_var.value
+        else:
+            env = self.load_environment()
+
+        sandbox.set_environment(env)
+
+        if artifact.buildsandbox and artifact.buildsandbox.working_directory:
+            sandbox.set_work_directory(artifact.buildsandbox.working_directory)
 
     #  load_proto()
     #
--- a/src/buildstream/_artifactcache.py
+++ b/src/buildstream/_artifactcache.py
@@ -303,6 +303,10 @@ class ArtifactCache(AssetCache):
                 except FileNotFoundError:
                     pass
 
+            if artifact_proto.buildsandbox:
+                for subsandbox_digest in artifact_proto.buildsandbox.subsandbox_digests:
+                    self.cas._send_directory(remote, subsandbox_digest)
+
             digests = [artifact_digest, artifact_proto.low_diversity_meta, artifact_proto.high_diversity_meta]
 
             if str(artifact_proto.public_data):
@@ -361,6 +365,9 @@ class ArtifactCache(AssetCache):
             referenced_directories.append(artifact_proto.sources)
         if artifact_proto.buildroot:
             referenced_directories.append(artifact_proto.buildroot)
+        if artifact_proto.buildsandbox:
+            for subsandbox_digest in artifact_proto.buildsandbox.subsandbox_digests:
+                referenced_directories.append(subsandbox_digest)
 
         referenced_blobs = [artifact_proto.low_diversity_meta, artifact_proto.high_diversity_meta] + [
             log_file.digest for log_file in artifact_proto.logs
@@ -419,6 +426,9 @@ class ArtifactCache(AssetCache):
                     self.cas.fetch_directory(remote, artifact.buildtree)
                 if str(artifact.buildroot):
                     self.cas.fetch_directory(remote, artifact.buildroot)
+                if artifact.buildsandbox:
+                    for subsandbox_digest in artifact.buildsandbox.subsandbox_digests:
+                        self.cas.fetch_directory(remote, subsandbox_digest)
 
             digests = [artifact.low_diversity_meta, artifact.high_diversity_meta]
             if str(artifact.public_data):
--- a/src/buildstream/buildelement.py
+++ b/src/buildstream/buildelement.py
@@ -219,6 +219,7 @@ class BuildElement(Element):
     def configure_dependencies(self, dependencies):
 
         self.__layout = {}  # pylint: disable=attribute-defined-outside-init
+        self.__digest_environment = {}  # pylint: disable=attribute-defined-outside-init
 
         # FIXME: Currently this forcefully validates configurations
         #        for all BuildElement subclasses so they are unable to
@@ -227,9 +228,18 @@ class BuildElement(Element):
         for dep in dependencies:
             # Determine the location to stage each element, default is "/"
             location = "/"
+
             if dep.config:
-                dep.config.validate_keys(["location"])
-                location = dep.config.get_str("location")
+                dep.config.validate_keys(["digest-environment", "location"])
+
+                location = dep.config.get_str("location", "/")
+
+                digest_var_name = dep.config.get_str("digest-environment", None)
+
+                if digest_var_name is not None:
+                    element_list = self.__digest_environment.setdefault(digest_var_name, [])
+                    element_list.append((dep.element, dep.path))
+
             try:
                 element_list = self.__layout[location]
             except KeyError:
@@ -268,6 +278,16 @@ class BuildElement(Element):
             }
             dictionary["layout"] = layout_key
 
+        # Specify the layout in the key, if buildstream is to generate an environment
+        # variable with the digest
+        #
+        if self.__digest_environment:
+            sorted_envs = sorted(self.__digest_environment)
+            digest_key = {
+                env: [dependency_path for _, dependency_path in self.__digest_environment[env]] for env in sorted_envs
+            }
+            dictionary["digest-enviornment"] = digest_key
+
         return dictionary
 
     def configure_sandbox(self, sandbox):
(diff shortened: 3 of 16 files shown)


src/buildstream/buildelement.py
```diff
@@ -285,10 +305,20 @@ def configure_sandbox(self, sandbox):
             command_dir = build_root
         sandbox.set_work_directory(command_dir)
 
-        # Setup environment
-        sandbox.set_environment(self.get_environment())
-
     def stage(self, sandbox):
+        # Setup environment
```
abderrahim, line 333: I've moved this here because buildstream wouldn't let me stage outside of `stage()`.

It might be better to try to work around that limitation rather than doing this virtual staging here.

src/buildstream/buildelement.py
```diff
@@ -285,10 +305,20 @@ def configure_sandbox(self, sandbox):
...
+        env = self.get_environment()
+
+        # Add "CAS digest" environment variables
+        sorted_envs = sorted(self.__digest_environment)
+        for digest_variable in sorted_envs:
+            element_list = [element for element, _ in self.__digest_environment[digest_variable]]
+            subsandbox = sandbox.create_subsandbox()
+            self.stage_dependency_artifacts(subsandbox, element_list)
+            digest = subsandbox.get_virtual_directory()._get_digest()
+            env[digest_variable] = "{}/{}".format(digest.hash, digest.size_bytes)
```
abderrahim, line 345: I also noticed that these environment variables don't end up in the build logs.

src/buildstream/sandbox/_sandboxremote.py
```diff
@@ -263,9 +263,17 @@ def _execute_action(self, action, flags):
                 "Uploading input root", element_name=self._get_element_name()
             ):
                 # Determine blobs missing on remote
+                root_digests = [action.input_root_digest]
+
+                # Add virtual directories for subsandboxes
+                for subsandbox in self._get_subsandboxes():
+                    vdir = subsandbox.get_virtual_directory()
+                    root_digests.append(vdir._get_digest())
```
abderrahim, line 182: I wonder if we should have an option to only upload these conditionally. It think it's not worth the trouble, but wanted to have a second opinion.

src/buildstream/buildelement.py
```diff
@@ -285,10 +305,20 @@ def configure_sandbox(self, sandbox):
...
     def stage(self, sandbox):
+        # Setup environment
+        env = self.get_environment()
+
+        # Add "CAS digest" environment variables
+        sorted_envs = sorted(self.__digest_environment)
+        for digest_variable in sorted_envs:
+            element_list = [element for element, _ in self.__digest_environment[digest_variable]]
+            subsandbox = sandbox.create_subsandbox()
+            self.stage_dependency_artifacts(subsandbox, element_list)
```
abderrahim: @gtristan was worried that the CAS digest might depend on the order the elements are declared in the .bst file

See https://github.com/apache/buildstream/pull/1991#discussion_r2153808770

src/buildstream/buildelement.py
abderrahim: This is due to the overlap collector, btw. The staging functions will refuse to work if there is no active overlap collector.

juergbi: Setting up the environment in `stage()` is a problem at least for `bst shell --build --use-buildtree`, which does not invoke `Element.stage()` but still requires environment variables to be set.

Draft fix for the overlap collector issue and moving environment setup back to `configure_sandbox()`: https://github.com/apache/buildstream/commits/jbilleter/digest-environment/

juergbi: I've pushed the fix to this branch and it seems to work well for normal builds. However, it still doesn't work for `bst shell --build --use-buildtree` (see failing test case) as that skips the loading of dependency artifacts. Always requiring build dependencies to be available for `bst shell --build --use-buildtree` would fix the test case but I suspect it would break buildtree shells for artifacts. Any suggestions?

src/buildstream/buildelement.py
juergbi: I've pushed an improvement to build environment logging.

src/buildstream/buildelement.py
gtristan: It looks to me that we do actually cache the environment variables in the artifact metadata, it seems to me that we should be relying on that in `bst shell --build --use-buildtree`.

juergbi: Right, it's part of the low diversity metadata. I'll have to check whether this would allow us to completely skip `configure_sandbox()` for the buildtree shell.

However, even if that works, there is still one potential issue, which is that this would not guarantee that the directory tree referenced by the digest environment variable is available in the local cache, as the artifact doesn't reference that digest (except via that environment variable but BuildStream core is not aware of that). Maybe we can add that to the Artifact proto as well, but this may require additional code and API in the core for use by `BuildElement`. Or do you have another suggestion?

gtristan: > However, even if that works, there is still one potential issue, which is that this would not guarantee that the directory tree referenced by the digest environment variable is available in the local cache, as the artifact doesn't reference that digest (except via that environment variable but BuildStream core is not aware of that). Maybe we can add that to the Artifact proto as well, but this may require additional code and API in the core for use by BuildElement. Or do you have another suggestion?

We are interestingly dancing on the line of sanity here.

* The assumption with `bst shell --build --use-buildtree` is that you have everything you need to reproduce the (probably failed) build locally in order to debug it.
* The assumption with exporting the local casd socket to the build environment, and moreso, remoting out the builds to a different sandbox (whether it is on the same machine or not)... is that you have your hands on some sensitive things, you know what you are doing, and you risk corrupting your build and cache with some non-determinism

The above two are not necessarily at odds with eachother, but it could be fair to say that if you are doing this wild stuff, you are out of bounds for having support for debugging remotely failed builds.

I think that I would be satisfied with either of the following:

* The officially supported and documented usage of `recc` is to use it locally in your sandbox, exercise the local compiler in the same sandbox, and use only the caching nature of `recc` to store/lookup objects in the CAS
  * This would "just work" given that we use the same environment variables from cached artifacts
* The officially supported and documented usage of `recc` is to selectively stage the compiler, source code and required header files to an adjacent sandbox on the same buildhost via the local `buildbox-casd` feature, and the artifact proto is extended to capture a list of digests which might be used for weird features like this one

I.e., lets support something, and if you are doing something outside of what is supported, you might be on your own and your build shells for debugging remotely failed artifacts might not work.

juergbi: The intention is to recommend the latter, i.e., use the remote execution protocol even for local execution, as the former (cache-only) is much easier to get wrong and get incorrect cache hits due to incomplete cache keys.

src/buildstream/sandbox/_sandboxremote.py
juergbi: I also think it's not worth it. The files in subsandboxes should typically be a subset of the main input root, so the overhead should just be a few `Directory` protobuf messages.

src/buildstream/buildelement.py
juergbi: The loader takes care of sorting dependencies via `loadelement.sort_dependencies()` early on, so `Element` and its subclasses never even have access to the declaration order and thus, there is no such issue, as far as I can tell.

gtristan: While that is true, it is unclear to me whether the `Element` constructor interprets the `config:` of each dependency in dependency order, or in declaration order.

juergbi: `Element._new_from_load_element()` iterates over the already sorted dependencies, creates a `DependencyConfiguration` object for each dependency with configuration data, and then hands that to `configure_dependencies()`, still in the same order. The declaration order can't even be observed by `Element`, as far as I can tell.

Does this clarify things or am I misunderstanding your comment?

src/buildstream/buildelement.py
juergbi: I've implemented the Artifact proto enhancement and all tests are now passing locally.

src/buildstream/buildelement.py
gtristan: Yes this clarifies things, this was precisely my concern :)

src/buildstream/_artifactcache.py
```diff
@@ -361,6 +361,9 @@ def _push_artifact_proto(self, element, artifact, artifact_digest, remote):
             referenced_directories.append(artifact_proto.sources)
         if artifact_proto.buildroot:
             referenced_directories.append(artifact_proto.buildroot)
+        if artifact_proto.buildsandbox:
+            for subsandbox_digest in artifact_proto.buildsandbox.subsandbox_digests:
```
abderrahim, line 369: I think we need to push these in `_push_artifact_blobs()` as well.

src/buildstream/_artifact.py
```diff
@@ -681,6 +696,21 @@ def pull(self, *, pull_buildtrees):
...
 
+    def configure_sandbox(self, sandbox):
+        artifact = self._get_proto()
+
+        if artifact.buildsandbox and artifact.buildsandbox.environment:
+            env = {}
+            for env_var in artifact.buildsandbox.environment:
+                env[env_var.name] = env_var.value
+        else:
+            env = self.load_environment()
```
abderrahim, line 707: I am not sure what this brings us. Why are we storing two different sets of environment variables in the artifact? Are they supposed to be different?

A bit of digging later: It seems the only relevance of the "static" environment is to compute the cache key, but we don't recompute the cache key from a loaded artifact. Am I missing something?

src/buildstream/buildelement.py
```diff
@@ -286,10 +306,20 @@ def configure_sandbox(self, sandbox):
         sandbox.set_work_directory(command_dir)
 
         # Setup environment
-        sandbox.set_environment(self.get_environment())
+        env = self.get_environment()
 
-    def stage(self, sandbox):
```
abderrahim, line 291: nit: removing the empty line after `def stage(self, sandbox):` results in a confusing diff.

tests/integration/project/elements/digest-environment/base-buildtree.bst
```diff
@@ -0,0 +1,12 @@
+kind: manual
```
abderrahim, line 1: This file is identical to `base.bst`, any reason to have a separate element?

abderrahim (commented):
A few comments but otherwise looks good.

(I can't approve on Github because I created the pull request)

src/buildstream/_artifactcache.py
juergbi: Good catch. I've added that incl. test coverage.

tests/integration/project/elements/digest-environment/base-buildtree.bst
juergbi: If I remember correctly, this is required for isolation with parallel runs of integration tests with the shared integration cache directory.

src/buildstream/_artifact.py
juergbi: Loading the (new) sandbox environment in `load_environment()` would be wrong as it is used to set `Element.__environment`, which should not contain e.g. digest environment variables. It's possible that `Element.__environment` loaded from the artifact is not actually used right now, however, it may be used in the future e.g. by `bst inspect`. I didn't want to change anything in this regard to keep the risk of breakage or compatibility problems to a minimum.

tests/integration/project/elements/digest-environment/base-buildtree.bst
abderrahim: Ack, I suspected as much. But I saw that there are two tests that share the same `base.bst` element. I guess the shared cache doesn't make a difference in that case?

src/buildstream/_artifact.py
abderrahim: Alright, I don't have a strong opinion either way. Happy to keep this if you believe it reduces risk.

src/buildstream/buildelement.py
```diff
@@ -286,7 +306,18 @@ def configure_sandbox(self, sandbox):
...
         # Setup environment
-        sandbox.set_environment(self.get_environment())
+        env = self.get_environment()
+
+        # Add "CAS digest" environment variables
+        sorted_envs = sorted(self.__digest_environment)
+        for digest_variable in sorted_envs:
+            element_list = [element for element, _ in self.__digest_environment[digest_variable]]
+            with self.subsandbox(sandbox) as subsandbox:
+                self.stage_dependency_artifacts(subsandbox, element_list)
```
abderrahim, line 343: We should probably put this inside a timed activity with `silent_nested=True`. Logging is a bit too verbose (and confusing) right now.

tests/integration/project/elements/digest-environment/base-buildtree.bst
juergbi: Yes, `test_build_shell` doesn't actually build `base.bst`, so a reused cachedir shouldn't cause any issues.

src/buildstream/buildelement.py
juergbi: Good point. I've updated the branch for this.

src/buildstream/buildelement.py
juergbi: Working on additional tests, I've noticed that `bst artifact checkout` may be broken for elements that use `digest-environment`. Need to look into this further.

juergbi: I've added the commit `element.py: Don't call configure_sandbox() for non-build sandboxes` and expanded the tests to cover `bst artifact checkout`.

src/buildstream/element.py
```diff
@@ -1453,7 +1453,11 @@ def _prepare_sandbox(self, scope, shell=False, integrate=True, usebuildtree=Fals
                 # Stage what we need
                 self.__stage(sandbox)
             else:
-                self.__configure_sandbox(sandbox)
+                # Runtime shell or `bst artifact checkout`
+
+                # Don't call `configure_sandbox()` as that may attempt to construct subsandboxes
+                # with build dependencies and we're not setting up a build sandbox.
+                sandbox.set_environment(self.get_environment())
```
abderrahim, line 1460: I think it makes sense to squash this commit into the one before it.

src/buildstream/buildelement.py
```diff
@@ -268,6 +278,16 @@ def get_unique_key(self):
             }
             dictionary["layout"] = layout_key
 
+        # Specify the layout in the key, if buildstream is to generate an environmenta
```
abderrahim: typo: "environmenta"

abderrahim (commented):
LGTM (with a couple nits)

PR #2039, opened 2025-07-18
tar.py: Fix hardlink extraction with latest Python

### LINKED ISSUE #2029: `test_out_of_basedir_hardlinks` test fails on Python 3.13.x
My Nixpkgs derivation for Buildstream has been failing recently with the test `test_out_of_basedir_hardlinks` on Python 3.13.x.

We disabled the test for now, but would like to get this fixed in 2.5.1, as test coverage is now reduced on Nixpkgs.

I've attached a log from the Nix build environment, but it's not clear what the issue is.

Nixpkgs does not have Buildstream 2.5.0 currently, due to protobuf dependencies - we suspect https://github.com/apache/buildstream/pull/2022 will land a fix.

[buildstream-hardlinks.log](https://github.com/user-attachments/files/21007656/buildstream-hardlinks.log)

juergbi:
Recent Python security fixes to the tarfile module (included in 3.13.4 but also backported to older branches) require link targets to pass the specified filter function as well. This can result in hardlinked files to not be extracted when a `base-dir` is set.

This commit separates the filtering from the extraction to avoid this issue. Pass `filter="tar"` as Python 3.14+ will default to the too restrictive `filter="data"`.

https://github.com/python/cpython/pull/135037

Fixes #2029.

### DIFF
--- a/src/buildstream/plugins/sources/tar.py
+++ b/src/buildstream/plugins/sources/tar.py
@@ -145,14 +145,14 @@ class TarSource(DownloadableFileSource):
                         base_dir = base_dir + os.sep
 
                 filter_function = functools.partial(self._extract_filter, base_dir)
+                filtered_members = []
+                for member in tar.getmembers():
+                    member = filter_function(member, directory)
+                    if member is not None:
+                        filtered_members.append(member)
                 if sys.version_info >= (3, 12):
-                    tar.extractall(path=directory, filter=filter_function)
+                    tar.extractall(path=directory, members=filtered_members, filter="tar")
                 else:
-                    filtered_members = []
-                    for member in tar.getmembers():
-                        member = filter_function(member, directory)
-                        if member is not None:
-                            filtered_members.append(member)
                     tar.extractall(path=directory, members=filtered_members)
 
         except (tarfile.TarError, OSError) as e:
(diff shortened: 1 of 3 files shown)


gtristan:
Looking at https://github.com/python/cpython/pull/135037, it looks like cpython is messing with extracted links, calling `os.path.realpath()` (with a new *"allow missing"* flag)... for some purpose... perhaps for avoiding creating hardlinked files outside of the destination directory...

Your commit itself looks sensible, and addresses the mentioned test case where we are testing tarball behavior - however the behavioral change in python is deep and a bit worrying.

Now that I'm seeing these `os.path.realpath()` calls popping up in the CPython code around "links", I wonder if this might effect symbolic links ? Surely it would be an absurd python bug if relative symbolic link targets ended up becoming absolute paths when extracted by `TarFile`, but it also looks like we don't have much coverage around symbolic links in tarball extraction in `tests/sources/tar.py`, so I'm not sure we would notice if such an absurd bug occurred.

@juergbi I think it is unlikely that python broke symlinks with this, I'll leave it up to you to decide whether you think we need to add more symlink coverage on our side, otherwise lets just go ahead with this merge.

juergbi:
> Now that I'm seeing these `os.path.realpath()` calls popping up in the CPython code around "links", I wonder if this might effect symbolic links ? Surely it would be an absurd python bug if relative symbolic link targets ended up becoming absolute paths when extracted by `TarFile`, but it also looks like we don't have much coverage around symbolic links in tarball extraction in `tests/sources/tar.py`, so I'm not sure we would notice if such an absurd bug occurred.

I've pushed a new test to verify that symlinks are supported and targets are not mangled.

The new test passes without code changes. It would fail on Python 3.14 without `filter="tar"` (as `filter="data"` will be the new default), so it's certainly nice to have that covered in the test suite.

PR #2043, opened 2025-08-15
requirements: Relax protobuf version constraint

juergbi:
All generated code since 3.20.0 will be supported until at least 8.x.y.

https://protobuf.dev/support/cross-version-runtime-guarantee/
https://github.com/protocolbuffers/protobuf/commit/1af7fd407c85c9c1f34b6ca8441cfb3a72d4f5ea

### DIFF
--- a/doc/source/hacking/grpc_protocols.rst
+++ b/doc/source/hacking/grpc_protocols.rst
@@ -36,6 +36,3 @@ To actually regenerate the code::
 
   ./setup.py build_grpc
 
-The ``requirements/requirements.in`` file needs to be updated to match the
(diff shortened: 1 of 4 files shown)


abderrahim:
I would revert https://github.com/apache/buildstream/pull/2017 wholesale, but given this new info it would be nice to have your changes on top of that.

juergbi:
I've reverted the remaining part of #2017. And I've added a tox environment `build-grpc` that regenerates the code with the desired version of `grpcio-tools` with a small documentation update.

abderrahim (approved):

PR #2045, opened 2025-08-24
_sandboxreapi: check stdout and stderr exist when fetching action result

abderrahim:
Starting with b8e6876d760641c2b6ccc3e2578eba48d0a33332, action results generated by the buildbox-run sandbox may be pushed to the action cache, and they don't have stderr and stdout stored.

### DIFF
--- a/src/buildstream/sandbox/_sandboxreapi.py
+++ b/src/buildstream/sandbox/_sandboxreapi.py
@@ -184,8 +184,13 @@ class SandboxREAPI(Sandbox):
 
             raise SandboxError("Output directory structure had no digest attached.")
 
-        # Fetch stdout and stderr blobs
-        cascache.fetch_blobs(casremote, [action_result.stdout_digest, action_result.stderr_digest])
+        # Fetch stdout and stderr blobs, if they exist
+        blobs = []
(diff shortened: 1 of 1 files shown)


src/buildstream/sandbox/_sandboxreapi.py
```diff
@@ -184,8 +184,12 @@ def _fetch_action_result_outputs(self, casremote, action_result):
...
             raise SandboxError("Output directory structure had no digest attached.")
 
-        # Fetch stdout and stderr blobs
-        cascache.fetch_blobs(casremote, [action_result.stdout_digest, action_result.stderr_digest])
+        # Fetch stdout and stderr blobs, if they exist
+        blobs = []
+        for digest in [action_result.stdout_digest, action_result.stderr_digest]:
+            if digest.hash:
+                blobs.append(digest)
+        cascache.fetch_blobs(casremote, blobs)
```
abderrahim: I think we can short-circuit this even further

```suggestion
        if blobs:
            cascache.fetch_blobs(casremote, blobs)
```

juergbi: This PR looks reasonable incl. the above suggestion.

juergbi (approved):

PR #2046, opened 2025-08-25
_yaml: set a large 'width' value in the roundtrip yaml config

abderrahim:
At some point, ruamel-yaml started line wrapping at 80 columns, breaking our 'roundtrip' promise

### DIFF
--- a/src/buildstream/_yaml.pyx
+++ b/src/buildstream/_yaml.pyx
@@ -389,6 +389,8 @@ yaml.RoundTripConstructor.add_constructor(u'tag:yaml.org,2002:timestamp',
 def prepare_roundtrip_yaml():
     yml = yaml.YAML()
     yml.preserve_quotes=True
+    # defaults to 80 if we don't set it
+    yml.width=9999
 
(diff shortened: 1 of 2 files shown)


src/buildstream/_yaml.pyx
```diff
@@ -389,6 +389,8 @@ yaml.RoundTripConstructor.add_constructor(u'tag:yaml.org,2002:timestamp',
 def prepare_roundtrip_yaml():
     yml = yaml.YAML()
     yml.preserve_quotes=True
+    # defaults to 80 if we don't set it
+    yml.width=9999
```
gtristan, line 393: Have we tried a negative value here ? It would seem a value of `-1` may be unlimited.

Otherwise `9999` seems like a perfectly sensible value.

abderrahim: This is the actual code:
```
        self.best_width = 80
        if width and width > self.best_sequence_indent * 2:
            self.best_width = width
```

There is no provision for negative values.

gtristan (approved):
LGTM :)

PR #2048, opened 2025-08-29
_stream: speed up `bst artifact show`

abderrahim:
We only need to check the remotes for cached artifacts if we don't have them cached locally.


gtristan (approved):
LGTM

PR #2050, opened 2025-08-30
ci: Add Debian 12 and 13 and update images

juergbi:
Add Debian 12 and 13 and update BuildBox to 1.3.31.

Debian 12 extends test coverage to Python 3.11. It is supported until June 10, 2028, which covers the full upstream Python 3.11 support period.

Debian 13 provides long-term test coverage of Python 3.13. It is supported until June 30, 2030, which covers the full upstream Python 3.13 support period.

### DIFF
--- a/.github/common.env
+++ b/.github/common.env
@@ -1,6 +1,6 @@
 # Shared common variables
 
-CI_IMAGE_VERSION=master-1938898146
+CI_IMAGE_VERSION=master-2011999263
 CI_TOXENV_MAIN=py39,py310,py311,py312,py313
 CI_TOXENV_PLUGINS=py39-plugins,py310-plugins,py311-plugins,py312-plugins,py313-plugins
 CI_TOXENV_ALL="${CI_TOXENV_MAIN},${CI_TOXENV_PLUGINS}"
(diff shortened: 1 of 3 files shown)


abderrahim (approved):
Looks good.

Now the fedoras look redundant since they both provide 3.13. We should probably keep only the latest fedora, which will be the first to bring us 3.14 when that's released.

We still have a gap for python 3.12, which we could fill by using ubuntu 24.04 (latest LTS).

juergbi:
> Now the fedoras look redundant since they both provide 3.13. We should probably keep only the latest fedora, which will be the first to bring us 3.14 when that's released.

Yes, I think it's fine to already drop Fedora 41.

> We still have a gap for python 3.12, which we could fill by using ubuntu 24.04 (latest LTS).

Indeed, Ubuntu 24.04 would be a suitable choice to cover Python 3.12. RHEL 8.10 or RHEL 10 (or derivatives) would be possible alternatives as they are also supported beyond upstream EOL of Python 3.12..

juergbi:
https://github.com/apache/buildstream/pull/2051 drops Fedora 41.

PR #2052, opened 2025-09-02
requirements: Add version exclusion qualifier for `Click` on 8.2.2

shymega-ct:
Click ==8.2.2 causes a regression that affects Buildstream's argument parsing. Currently, one known case is where `--no-strict` is set when not explicitly configured to by the command-line.

This change to requirements.in adds a exclusion for Click 8.2.2, but allows any other version higher than Click 7.0.

### DIFF
--- a/requirements/requirements.in
+++ b/requirements/requirements.in
@@ -1,4 +1,4 @@
-Click >= 7.0
+Click >= 7.0, != 8.2.2
 grpcio
 Jinja2 >= 2.10
 importlib_metadata >= 3.6; python_version < "3.10"


juergbi:
click 8.2.2 got yanked from PyPI due to this unintended change in behavior, so this should no longer be needed: https://pypi.org/project/click/#history

jjardon:
@juergbi Can we please still merge this patch? Not everyone uses pip to install dependencies: we might make it more flexible to only restrict 8.2.2 being used? ie allow anything < 8.2.2 and > 8.2.2

shymega:
@jjardon I think we could achieve something like that with 'version specifiers', which is implemented by `pip` - specifically, a version exclusion clause.

I could edit this PR to implement something like:

`Click >= 7.0; != 8.2.2`, if that syntax is correct - off the top of my head.

juergbi:
> `Click >= 7.0; != 8.2.2`, if that syntax is correct - off the top of my head.

I think multiple version specifiers need to be separated by a comma, not a semicolon. I'm happy to merge that if that helps integrations where the yanking is not sufficient. (It obviously still won't help when installing the already released BuildStream 2.5.0).

abderrahim:
> Not everyone uses pip to install dependencies: we might make it more flexible to only restrict 8.2.2 being used? ie allow anything < 8.2.2 and > 8.2.2

I think that it won't help those who don't use pip to install dependencies. For instance, even if we had this in 2.5.0, it wouldn't have prevented freedesktop-sdk from picking up the update. I don't know how other distros work (would e.g. the Fedora packaging scripts pick up on incompatible versions requirements? I don't think they would unless they're calling to pip under the hood)

jjardon:
btw, I know is not going to help much with 2.5.0, I'm thinking more in 2.5.1 and future releases

shymega:
@abderrahim @juergbi Could you approve the CI on this PR so we can check the change works?

Thanks.

juergbi (approved):

abderrahim:
Please update the pull request title

shymega-ct:
@abderrahim That's now done, as well as typos fixed, and PR description updated in-line with commit message.

PR #2057, opened 2025-09-08
docs: Add 'Execution of BuildStream Elements' page

shymega-ct:
WIP!

### DIFF
--- a/doc/source/using_building_elements.rst
+++ b/doc/source/using_building_elements.rst
@@ -1,0 +1,121 @@
+Building of BuildStream Elements
+=================================
+
+This page provides an high-level overview of the processes involved in which
+Elements are built as part of the build+integration pipeline. 
+
+When `bst build` is running on a given Element, it can be - roughly - broken
+down into individual steps:
+
+- Sources are fetched
+- The sandbox is prepared
+- Configure and run commands
+   These can include:
+      - `configure-commands`
+      - `build-commands`
+      - `install-commands`
+      - `strip-commands`.
+- Caching artifacts
+
+Sources are fetched
+-------------------
+
+The hidden first step is actually validating the ``yaml``. This includes
+resolving includes, options, appends, which are denoted by ``(@)``,
+``(?)`` and ``(>)`` respectively.
+
+The first step once the ``yaml`` has been validated is that BuildStream
+will fetch the sources. This is dependent on the source ``kind`` as to
+how these are fetched. After the sources are fetched they may need to
+checked. For example a ``kind: tar`` would need to check the ``sha256``
+matches, and a ``git_repo`` source would need to switch to the specified
+commit sha.
+
+The Sandbox is prepared
+-----------------------
+
+The Sandbox is prepared as a temporary filesystem, where build
+dependencies (``build-depends:``) and general dependencies (``depends``)
+are staged, along with their own runtime dependencies. This happens in
+an abstract state, which can quickly spot repeated files and overlaps.
+
+When a dependency is being staged, the produced artifact is added from the root
+(``/``). This can sometimes be changed using the ``location:`` attribute.
+However, this applies to `BuildElements`, and not all `Elements` (where
+`BuildElement` is a subclass of `Element`) support the `location` argument. 
+
+In most cases dependencies are marked by BuildStream as immutable. 
+
+After the dependencies are staged BuildStream stages the sources in the
+``build-root`` location. The actual location of this differs slightly
+depending on the project file structure, which is why it is common to
+see elements use the variable ``%{build-root}`` which resolves to the
+correct location.
+
+Configure and Build commands
+----------------------------
+
+Now that all dependencies and sources are staged in a temporary filesystem,
+this filesystem is mounted inside BuildBox as a sandbox.
+
+The first commands to be run are configure commands and it is
+recommended to include things like moving the sources about, generating
+config files and other “configure” related actions into this section.
+These should be the commands that can only be run once (for example a
+folder can only be moved once), this is due to BuildStream
+workspaces.
+
+!!! tip “BuildStream Workspaces and Configure Commands”
+
+.. toctree::
+   :numbered:
+   :maxdepth: 1
+
+   developing/workspaces.rst
+
+::
+
+   When a workspace is opened, it stages all the sources for the indicated
+   element locally, then when doing a build of that element it uses these local
+   sources instead of pulling in fresh sources. Builds using workspaces only
+   run configure commands once, and any subsequent builds using the same
+   workspace will skip the configure commands step, therefore steps of the
+   build that aren't reproducible (without re-staging sources) should be added
+   to configure commands.
+
+After Configure commands are run, then Build commands are next. Build
+commands are intended to contain the actual build process, for example
+if the build uses ``make`` then this stage should run the
+``make target`` command.
+
+Install Commands and Caching Artifacts
+--------------------------------------
+
+Install commands are the final commands that are run before BuildStream
+collects the artifacts and closes the build sandbox. Install commands
+should mainly be comprised of moving the built artifacts from the
+``${build-root}`` to the ``${install-root}``.
+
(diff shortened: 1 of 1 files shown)


doc/source/using_execution_of_elements.rst
```diff
@@ -0,0 +1,112 @@
+Execution of BuildStream Elements
+=================================
+
+This page provides a guide on how BuildStream elements are converted
+into executable build steps, including how element types may affect
```
abderrahim: I feel this goes against the terminology used by buildstream. You don't *execute* an element, you probably *build* it. Also talking about "converting a buildstream element into executable steps" feels wrong as well.

doc/source/using_execution_of_elements.rst
```diff
@@ -0,0 +1,112 @@
...
+this.
+
+When a BuildStream element is being built, it can roughly be broken down
+into these individual stages:
+
+- Sources are fetched
+- The sandbox is prepared
+- Configure and Run commands
+- Install Commands and Caching Artifacts
```
abderrahim: This is wrong, install commands run together with other commands. Caching artifacts is indeed a separate step.

doc/source/using_building_elements.rst
```diff
@@ -0,0 +1,112 @@
...
+
+- Sources are fetched
+- The sandbox is prepared
+- Configure and Run commands
+- Install Commands and Caching Artifacts
+
+Sources are fetched
+-------------------
+
+The hidden first step is actually validating the ``yaml``. This includes
```
abderrahim, line 23: This "hidden" step needs to be "unhidden" in the documentation.

doc/source/using_building_elements.rst
```diff
@@ -0,0 +1,112 @@
...
+resolving includes, options, appends, which are denoted by ``(@)``,
+``(?)`` and ``(>)`` respectively.
```
abderrahim, line 25: I think it makes sense to link to the rest of the documentation, having a document like this feels very weird.

doc/source/using_execution_of_elements.rst
```diff
@@ -0,0 +1,112 @@
...
+how these are fetched. After the sources are fetched they may need to
+checked. For example a ``kind: tar`` would need to check the ``sha256``
+matches, and a ``git_repo`` source would need to switch to the specified
+commit sha.
+
+The Sandbox is prepared
+-----------------------
+
+The Sandbox is prepared as a temporary filesystem, where build
+dependencies (``build-depends:``) and general dependencies (``depends``)
```
abderrahim, line 38: build dependencies is enough, "general dependencies" as this document calls them are "both" build and runtime dependencies in buildstream terminology.

doc/source/using_execution_of_elements.rst
```diff
@@ -0,0 +1,112 @@
...
+are staged, along with their own runtime dependencies. This happens in
+an abstract state, which can quickly spot repeated files and overlaps.
+
+In general when a dependency is being staged, the produced artifact is
+added from the root (``/``). This can sometimes be changed using the
```
abderrahim: If we're talking about elements in general, they can be staged anywhere the plugin wants (and `location` isn't a generic configuration that all elements support). If we're talking about build elements, then "In general" isn't needed.

doc/source/using_execution_of_elements.rst
```diff
@@ -0,0 +1,112 @@
...
+``build-root`` location. The actual location of this differs slightly
+depending on the project file structure, which is why it is common to
+see elements use the variable ``%{build-root}`` which resolves to the
+correct location.
+
+Configure and Build commands
+----------------------------
+
+Now that all dependencies and sources are staged in a temporary
+filesystem, this filesystem is started up by BuildBox in a sandbox.
```
abderrahim: "this filesystem is started" <- what does that mean?

doc/source/using_execution_of_elements.rst
```diff
@@ -0,0 +1,112 @@
...
+
+The first commands to be run are configure commands and it is
+recommended to include things like moving the sources about, generating
+config files and other “configure” related actions into this section.
+These should be the commands that can only be run once (for example a
+folder can only be moved once), this is due to `BuildStream
+workspaces <https://docs.buildstream.build/2.4/developing/workspaces.html>`__.
```
abderrahim: You need to link to the appropriate document, not to the 2.4 version of the documentation.

But if you want to talk about workspaces, you should split configure commands from the rest of the commands above.

doc/source/using_execution_of_elements.rst
```diff
@@ -0,0 +1,112 @@
...
+
+::
+
+   When a [workspace](https://docs.buildstream.build/2.4/developing/workspaces.html)
+   is opened, it stages all the sources for the indicated element locally, then
+   when doing a build of that element it uses these local sources instead of
+   pulling in fresh sources. Builds using workspaces only run configure
+   commands once, and any subsequent builds using the same workspace will skip
+   the configure commands step, therefore steps of the build that aren't 
+   repeatable (without re-staging sources) should be added to configure
```
abderrahim: > repeatable

It seems this word is using in a different meaning here than usual.

doc/source/using_execution_of_elements.rst
```diff
@@ -0,0 +1,112 @@
...
+
+Install Commands and Caching Artifacts
+--------------------------------------
+
+Install commands are the final commands that are run before BuildStream
+collects the artifacts and closes the build sandbox. Install commands
+should mainly be comprised of moving the built artifacts from the
+``${build-root}`` to the ``${install-root}``.
+
+There is no need for Install commands to clean up any of the sources
```
abderrahim: I'd say it's the reverse. It's not that there is no need, it's more that you shouldn't. The sources can be stored as a build tree, and they need to stay around if you want to inspect the buildtree after the fact.

doc/source/using_execution_of_elements.rst
```diff
@@ -0,0 +1,112 @@
...
+that aren’t going to moved, as the BuildStream handles this when the
+sandbox is closed.
+
+Directories can be created under the install location, for example
+``%{install-root}/example/``, and these will be maintained when another
+element depends on this one, for example this will become
+``./example/``.
```
abderrahim: You meant `/example`

abderrahim (commented):
I left a few comments. This document needs a rewrite as it has a lot of factual inaccuracies and wrong terminology.

juergbi (commented):
I think this document contains useful information, however, the scope doesn't seem completely clear to me.

Looking at the current BuildStream documentation, I think we're indeed missing a good overview of the different stages for a given element during `bst build`, so it would be nice to have that.

However, this document includes a lot of details that I don't consider useful as part of an overview. We should link to other sections of the documentation for that (and expand other sections for details that aren't documented anywhere else right now).

I.e., I would define the scope of this page to be an overview of the build process of an element, as I think that's what we're missing most. Or did you have a different scope or target audience in mind? I'm open to other possibilities but I think we should be reasonably clear about what does and what does not belong on this new page.

doc/source/using_building_elements.rst
shymega-ct: What do you mean, specifically by this? Is there a specific page you could link to for context?

shymega-ct:
@juergbi I agree broadly with your thoughts here. I am rewriting this based on your comment, to make it a MVP first, and then iterate upon it.

Would it be best to open a new PR, or keep this one?

juergbi:
> Would it be best to open a new PR, or keep this one?

As this PR already has review comments, at least some of which may remain relevant, I'd probably keep this PR. Depending on how you approach this, a new PR could also be fine but please make sure to address or copy all existing review comments unless they don't apply at all anymore.

doc/source/using_building_elements.rst
shymega-ct: Any page in particular?

juergbi:
My attempt of getting this to a mergeable state: https://github.com/apache/buildstream/pull/2126

juergbi:
Closing this as I've merged #2126.

PR #2058, opened 2025-09-10
_frontend: allow showing the element kind in `bst show`

abderrahim:


src/buildstream/_frontend/widget.py
```diff
@@ -355,6 +355,7 @@ def show_pipeline(self, dependencies, format_):
             description = " ".join(element._description.splitlines())
 
             line = p.fmt_subst(line, "name", element._get_full_name(), fg="blue", bold=True)
+            line = p.fmt_subst(line, "kind", element.get_kind(), fg="blue", bold=True)
```
juergbi: Nit: I think I would drop the `bold` and/or the color.

juergbi (approved):
One nit but looks good to me overall.

src/buildstream/_frontend/widget.py
abderrahim: Dropped the bold (and changed the color).

PR #2060, opened 2025-09-15
docs/overlaps: Add 'Changing the location' section

shymega-ct:
- This adds a section about configuring some Element kinds for staging their dependencies in certain locations.

### DIFF
--- a/doc/source/handling-files/overlaps.rst
+++ b/doc/source/handling-files/overlaps.rst
@@ -144,6 +144,29 @@ command contain an ``rm`` statement, as such:
 This would cause later builds of ``runtime-only.bst`` to no longer
 conflict on the given file.
 
+Changing the location
+---------------------
+
+Some element kinds allow you to stage their dependencies in certain locations.
+This would allow for both versions to exist, separated in different
+directories. An important thing to note here is that this does also change the
+location of the runtime dependencies of an element. This means that it is
+important to check that the element in question can cope with this movement of
+its dependencies. Expanding off of this, if two dependencies have the same
+runtime dependency and then one is moved, there will now be two copies of that
+runtime dependency, one in the original location and one in the moved location.
+
+An example of changing the location of a dependency:
+
+.. code:: yaml
+
+   filename: components/example.bst
+   config:
+     location: "/test"
+
+The BuildStream documentation for changing the staging location can be found
+`here
+<https://docs.buildstream.build/master/buildstream.buildelement.html#location-for-staging-dependencies>`__
 
 Summary
 -------


doc/source/handling-files/overlaps.rst
```diff
@@ -144,6 +144,29 @@ command contain an ``rm`` statement, as such:
 This would cause later builds of ``runtime-only.bst`` to no longer
 conflict on the given file.
 
+Changing the location
```
abderrahim, line 147: This section doesn't fit with the rest of the document. The document discusses an example project, while this section discusses a theoretical example that isn't well defined.

I'm not sure what's the best way to deal with this. Extend the example project so that this can be applied? That could be too complicated for this tutorial. Move it to somewhere else in the documentation? I don't know where that could be.

shymega-ct: I agree that it doesn't /really/ fit with the rest of the document. I
looked through the docs to find an alternative place, but none felt
like a good place.

I guess we could just keep the section in `BuildElement` and refer to
that from other documentation?

shymega-ct: @juergbi @gtristan What are your thoughts on a good way to place this section in the docs?

juergbi: > I guess we could just keep the section in BuildElement and refer to
that from other documentation?

Yes, I think it would be reasonable to add the information from this section to the `BuildElement` documentation as that's where `location` is documented. That said, the `BuildElement` documentation is a bit hidden as it's part of the Plugin API reference even though `location` is more relevant to project developers than plugin developers. I.e., maybe we could find a better place for the whole `location` documentation. However, moving that around is not directly related to expanding the documentation to cover the handling of runtime dependencies, so I suggest to first land the expanded documentation in `BuildElement` and then separately consider moving that section to a more relevant/discoverable part of the documentation.

gtristan: I believe it is also, separately documented in `ScriptElement` right ?

There are a couple (slowly growing number) of features which cannot be directly implemented on the `Element` class but are common amongst abstract subclasses, it *can perhaps* make sense perhaps to add a page to consolidate documentation for these common features on a single page...

shymega-ct: I'm leaning towards consolidating documentation, but would like to reach a consensus amongst BuildStream developers.

shymega-ct: Just wanted to follow up on this. I agree broadly with @gtristan on consolidating the documentation with a new page, but would like to hear @abderrahim and @juergbi's opinions too.

abderrahim: It seems we have something about this topic in the [BuildElement API documentation](https://docs.buildstream.build/2.6/buildstream.buildelement.html#built-in-functionality) (the "Built-in functionality" section). It's hard to find given it's buried so deep in the API docs.

I suggest we take this section and put it in a new page in the "Using" part of the documentation, as it's more of a user-facing thing.

shymega-ct: Ack. Sounds good. I will merge that section, and the 'Changing the location' section in this PR together, in a suitable order.

For now, I will close this PR/discussion, and open a new PR next week.

shymega-ct:
Closing in favour of a new PR. Will link to the new PR when ready.

shymega-ct:
New PR here: https://github.com/apache/buildstream/pull/2103

PR #2061, opened 2025-09-15
docs/README: Add note to make OS requirement clear

shymega-ct:
This PR adds a note to the README to make the OS requirement clearer.

It might be that people think WSL2 can run Buildstream/Buildbox, and this has not yet been officially tested.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -110,6 +110,10 @@ To get started, first `install BuildStream by following the installation guide
 and then follow our tutorial in the
 `user guide <https://docs.buildstream.build/master/main_using.html>`_.
 
+  Running BuildStream on macOS or Windows may work under Docker Desktop, WSL,
+  or Podman Desktop, but it is not officially supported, and CI runs tests only on
+  Linux.
+
 We also recommend exploring some existing BuildStream projects:
 
 * https://gitlab.gnome.org/GNOME/gnome-build-meta/


README.rst
```diff
@@ -110,6 +110,8 @@ To get started, first `install BuildStream by following the installation guide
 and then follow our tutorial in the
 `user guide <https://docs.buildstream.build/master/main_using.html>`_.
 
+Currently, Buildstream (and Buildbox) are only supported on arm64/x86_64 Linux machines.
```
abderrahim: I don't think this is accurate.

AFAIK, we're not doing anything Linux specific in BuildStream. It should work the same on any POSIX system.

Same for architecture, there is nothing arm64/x86_64 specific in BuildStream. I know for a fact that it is being used on PowerPC and RISC V architectures just fine. There is a limitation in what target architectures it can build for, but this is more due to the fact it uses the bazel remote APIs [platform lexicon](https://github.com/bazelbuild/remote-apis/blob/main/build/bazel/remote/execution/v2/platform.md) rather than something in buildstream itself.

As for the limitations of buildbox, I don't think this is the right place to document them. Yes `buildbox-run-bubblewrap` is linux specific, but there are other `buildbox-run` implementations.

Another important thing is that we may support different things for local builds and for remote execution, so we shouldn't just put everything in one basket.

shymega-ct: Ah, I should have kept the architecture out of that addition.

Especially as we know Buildbox runs even on s390x architectures.

I'll remove the Buildbox mention, and the arm64/x86_64 mention for
Buildstream; although I think we should minimise any further depth of
explanations about RECC and target architectures. Perhaps it might be
worth linking to a page in the docs?

abderrahim: Nobody talked about RECC :wink: 

In that case, I don't see anything to add to this README. Unless you want to word it as "Windows is not supported" (and possibly mention WSL2)

shymega-ct: I've adjusted it in 95bda475f :smile:

README.rst
```diff
@@ -110,6 +110,9 @@ To get started, first `install BuildStream by following the installation guide
 and then follow our tutorial in the
 `user guide <https://docs.buildstream.build/master/main_using.html>`_.
 
+  Running Buildstream on Windows is not supported, although WSL2 may result in
```
juergbi:
```suggestion
  Running BuildStream on Windows is not supported, although WSL2 may result in
```

README.rst
```diff
@@ -110,6 +110,9 @@ To get started, first `install BuildStream by following the installation guide
...
+  some success; this workflow however, is not officially supporetd.
```
juergbi: "may result in some success" is too unclear to me. WSL2 is essentially just Linux in a VM (+ a lot of integration work) and I would expect BuildStream to work fine in WSL2 but haven't tested it myself.

If there are known issues with BuildStream in WSL2, I would be less vague. If there are no known issues, I might formulate it something like:
"Running BuildStream on Windows is not supported. WSL2 may be used but is also not officially supported."

jjardon:
I think this add a line about windows not being officially supported but It's not very clear about what OSes are actually supported (my understanding is only Linux but I might be wrong)

Can we be more clear about that, please?

shymega-ct:
> I think this add a line about windows not being officially supported but It's not very clear about what OSes are actually supported (my understanding is only Linux but I might be wrong)
> 
> Can we be more clear about that, please?

I believe it's Linux-only, but WSL2 is an instance of Linux in a VM, so I think @juergbi's suggestion works well.

cs-shadow:
> > I think this add a line about windows not being officially supported but It's not very clear about what OSes are actually supported ...

I concur, and would also suggest to phrase it in the positive direction of what's supported as opposed to enumerating what's not supported (which can be a never-ending list)

>> ... (my understanding is only Linux but I might be wrong)
> > Can we be more clear about that, please?
> 
> I believe it's Linux-only, but WSL2 is an instance of Linux in a VM, so I think @juergbi's suggestion works well.

I believe Linux is the only thing that we test on and where BuildStream is used in earnest. In the past, there were weak attempts to make it work on MacOS and Windows to varying degrees but I don't think either of those work today in a meaningful way.

Other than WSL, Docker Desktop (or Podman Desktop and such) can be a viable option in a similar "may be used but officially unsupported" category. The [contrib/bst-here](https://github.com/apache/buildstream/blob/master/contrib/bst-here) script was created to that end and can be potentially referenced here.

README.rst
```diff
@@ -110,6 +110,10 @@ To get started, first `install BuildStream by following the installation guide
 and then follow our tutorial in the
 `user guide <https://docs.buildstream.build/master/main_using.html>`_.
 
+  Running BuildStream on macOS or Windows may work under Docker Desktop, WSL,
+  or Podman Desktop, but we only test in CI under Linux. but is not officially
+  supported.
```
juergbi: Two 'but's and the end is not a complete sentence.

```suggestion
  or Podman Desktop, but it is not officially supported and CI runs tests only on Linux.
```

Or did you mean to remove the part after the second 'but' and it was left in by accident?

shymega-ct: Thanks, good catch! Fixed.

juergbi (approved):

PR #2063, opened 2025-09-18
nix: Add Nix tooling support

shymega-ct:
The key difference here is that the Nix package manager will use the Git rev of this repo to build itself, making it useful for developer iterations. It overrides the attribute set of the upstream Nixpkg derivation to use this Git repo.

- `default.nix`/`shell.nix` allow 'legacy Nix' to use the Flake without needing the Flake feature 'enabled'.
- `flake.lock` may need updating occasionally, but only if there's an update to the Buildstream package other than a version bump upstream on Nixpkgs.
- `flake.nix` is where the development shell and 'Flake-enabled Nix' tooling works from.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,42 @@
+{
+  "nodes": {
+    "flake-compat": {
+      "locked": {
+        "lastModified": 1733328505,
+        "narHash": "sha256-NeCCThCEP3eCl2l/+27kNNK7QrwZB1IJCrXfrbv5oqU=",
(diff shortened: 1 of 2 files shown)


shymega-ct:
Just to add to the above, but I _haven't_ added CI jobs for testing the Nix Flakes/derivations in this PR, as I felt it would best addressed in a new PR.

shymega-ct:
Closing as unneeded, and adds too much clutter to the root of the repository.

PR #2064, opened 2025-09-19
scheduler.py: Replace asyncio child watcher with our own watcher thread

### LINKED ISSUE #1990: Fix usage of deprecated get_child_watcher
There's a new warning in BuildStream output
```
  /usr/lib64/python3.12/asyncio/events.py:828: DeprecationWarning: 'get_child_watcher' is deprecated as of Python 3.12 and will be removed in Python 3.14.
```

juergbi:
The asyncio child watcher system was deprecated in Python 3.12 and will be fully removed in Python 3.14.

We use it only to watch a single process (buildbox-casd) and can use a simple watcher thread for that.

Fixes #1990.

### DIFF
--- a/src/buildstream/_scheduler/scheduler.py
+++ b/src/buildstream/_scheduler/scheduler.py
@@ -25,6 +25,7 @@ import signal
 import datetime
 import multiprocessing.forkserver
 import sys
+import threading
 from concurrent.futures import ThreadPoolExecutor
 
 # Local imports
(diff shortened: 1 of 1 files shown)


juergbi:
I've tested this also with Python 3.14 locally and that works but requires a fix for https://sourceforge.net/p/ruamel-yaml-clib/tickets/45/.

gtristan (approved):
Looks good to me.

PR #2066, opened 2025-09-19
Add support for Python 3.14

juergbi:
The test suite passes with this branch on Fedora 43: https://gitlab.com/BuildStream/buildstream-docker-images/-/jobs/11449217892

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,7 @@ concurrency:
 
 jobs:
   tests:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     continue-on-error: ${{ matrix.allow-failure || false }}
 
     strategy:
@@ -50,6 +50,9 @@ jobs:
             allow-failure: true
 
     steps:
+      - name: Disable AppArmor restriction for bubblewrap
+        run: sudo sysctl -w kernel.apparmor_restrict_unprivileged_userns=0
+
       - name: Check out repository
         uses: actions/checkout@v2
         # BuildStream requires tags to be able to find its version.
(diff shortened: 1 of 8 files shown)


.github/workflows/ci.yml
```diff
@@ -26,7 +26,7 @@ concurrency:
 
 jobs:
   tests:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
```
abderrahim, line 29: Updating to Ubuntu 24.04 isn't a good idea given we use bubblewrap. Is there anything else we can use?

https://etbe.coker.com.au/2024/04/24/ubuntu-24-04-bubblewrap/

Or maybe we can figure out a workaround

abderrahim: I see you've added `sudo aa-teardown || true`. I wonder what's the point of `|| true`, isn't it game over if we can't get apparmor out of the way?

juergbi: Unfortunately, the only standard Linux GitHub runner flavors are ubuntu-22.04 and ubuntu-24.04. I would also prefer staying away from the broken Ubuntu 24.04 but it seems we have no choice.

There is a bug in aa-teardown that results in a non-zero exit code despite unloading the profiles, as I understand it. https://bugs.launchpad.net/ubuntu/+source/apparmor/+bug/2093797. Should be fixed upstream https://gitlab.com/apparmor/apparmor/-/merge_requests/1484 but still seems to be an issue in Ubuntu 24.04.

Relevant GitHub issue: https://github.com/actions/runner-images/issues/10015

juergbi: It's still failing with these AppArmor disable commands, though. The error message changes from

`bwrap: loopback: Failed RTM_NEWADDR: Operation not permitted`

to

`bwrap: Creating new namespace failed: Permission denied`

but that's not helpful.

juergbi: This seems to work:

`sudo sysctl -w kernel.apparmor_restrict_unprivileged_userns=0`

abderrahim: Great

abderrahim (approved):

PR #2084, opened 2025-11-01
fix: use find_spec for pip plugins

kotborealis:
For pip plugins installed in editable mode, via `pip install -e ...`, buildstream failed to resolve the actual location of source files of a plugin.

`locate_file` resolved location to `site-packages/<module-name>/elements/<element.py>`, while `site-packages` contained only the `pth` file pointing to the another directory.

This patch updates `pluginoriginpip.py` to use `find_spec` to find files. It works properly with editable packages and `pth` files.

### DIFF
--- a/src/buildstream/_pluginfactory/pluginoriginpip.py
+++ b/src/buildstream/_pluginfactory/pluginoriginpip.py
@@ -11,7 +11,7 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 #
-import os
+from pathlib import Path
 
 from .._exceptions import PluginError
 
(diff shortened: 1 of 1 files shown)


kotborealis:
@gtristan Any chance this could be merged?

abderrahim:
Can you please rebase it?

kotborealis:
Rebased it cleanly, without any merge commits.

Also wanted to ask a question related to plugins: it's hard-coded that plugins search for a .yaml file for the default configuration. Should .yml files be considered valid as well?

kotborealis:
Accidentally closed this while messing with git rebases, reopening.

abderrahim (approved):
I moved the `importlib` import together with the others (inside the function). I don't know whether that's still needed or if it was just something specific to pkg_resources, but it's better to do that in a separate commit/PR.

Let's merge this.

PR #2085, opened 2025-11-02
Drop support for Python 3.9

juergbi:
Python 3.9 is no longer maintained upstream and at least one Python package (Click) has dropped support for it.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,7 +45,7 @@ build-frontend = "build"
 environment = { BST_BUNDLE_BUILDBOX = "1" }
 
 # The BuildBox binaries produced in buildbox-integration are linked against GLIBC 2.28
-# from Debian 10. See: https://gitlab.com/BuildGrid/buildbox/buildbox-integration.
+# from Rocky Linux 8.10. See: https://gitlab.com/BuildGrid/buildbox/buildbox-integration.
 #
 # The PyPA manylinux_2_28 platform tag identifies that our wheel will run on any x86_64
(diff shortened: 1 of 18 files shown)


pyproject.toml
```diff
@@ -53,10 +53,11 @@ environment = { BST_BUNDLE_BUILDBOX = "1" }
 manylinux-x86_64-image = "manylinux_2_28"
 
 skip = [
-  # BuildStream supports Python >= 3.9
+  # BuildStream supports Python >= 3.10
```
abderrahim: I wonder if we could set `requires-python` above and not have to update this list every time we bump the dependency

juergbi: The best way would be to migrate project metadata to `pyproject.toml`. However, if I remember correctly, it's not possible to add `[project]` with a single field such as `requires-python` and leave everything else as is. We'd have to migrate all the project metadata over from `setup.py`. While that would be nice, it may be a bit more work and should be a separate PR (I'm not sure whether we could keep the non-metadata parts of `setup.py` as is).

However, it does seem reasonable to replace part of the `skip` list with `project-requires-python`. I'll give this a try.

juergbi: `project-requires-python` can't be set in `pyproject.toml`, unfortunately. I had to set `python_requires` in `setup.cfg` instead.

abderrahim: I think I have a wip branch for migrating to pyproject.toml. We can probably land it and build on it, I'll send a pull request later.

PR #2088, opened 2025-11-09
Support artifact remote options in `bst artifact show`

gtristan:
This command was missed when applying the `--artifact-remote` and `--ignore-project-artifact-remotes` options uniformly to all relevant commands.


gtristan:
No reviews in a while... trivial and uncontroversial change (consistent CLI options) which also improves test coverage... lets merge this.

PR #2093, opened 2025-11-14
Store marked directories in Artifact proto

### LINKED ISSUE #2092: Filesystem in buildtree shell is read-only
Since BuildStream 2.6.0 (commit 7e884747f313bbe77aa7fc33444e9c3540af29bd), the filesystem in buildtree shells is read-only, which hinders investigation of build failures. This applies to both, interactive build shells as well as `bst shell --use-buildtree`.

juergbi:
Without marked directories, the filesystem in buildtree shells is read-only.

Fixes #2092.


gtristan (approved):
This is great !

It's so satisfying to see that this did not require any changes to `element.py` or `_artifact.py`, and voila, the CWD is also preserved, great :)

I'll happily approve this as is, after looking through the `bst shell` docs and noticing that nothing is mentioned about the behavior on an artifact without a locally supporting project... it's a subtle API break but it is a correction of something that was obviously broken...

It would probably be responsible to preemptively add a NEWS entry about the subtle API break, just for the sake of the quality of our release notes.

PR #2094, opened 2025-11-20
_elementsources: cache intermediate combined sources (in-memory)

abderrahim:
When tracking an element that has multiple sources that depend on previous sources, buildstream stages all previous sources each time, and since they aren't cached because we only cache sources that don't depend on previous sources or combined elements this ends up staging things from original sources again and again.

This change keeps the a digest of the intermediate staged sources in memory, and speeds up these cases a lot

### DIFF
--- a/src/buildstream/_elementsources.py
+++ b/src/buildstream/_elementsources.py
@@ -47,6 +47,8 @@ class ElementSources:
         self._cache_key = None  # Our cached cache key
         self._proto = None  # The cached Source proto
 
+        self._cache = {}
+
     # get_project():
     #
     # Return the project associated with this object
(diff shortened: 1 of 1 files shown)


aleixpol:
+1. the `bst source track` of my module went from 32h to 36 minutes.

src/buildstream/_elementsources.py
```diff
@@ -462,6 +464,10 @@ def _stage(self, *, stop=None):
             if source == stop:
                 break
 
+            if source in self._cache:
+                vdir = CasBasedDirectory(cas, digest=self._cache[source])
```
nathanwilliams-ct, line 468:
```suggestion
            if (digest := self._cache.get(source)) is not None:
                vdir = CasBasedDirectory(cas, digest)
```

PR #2097, opened 2025-11-21
Merge duplicate dependencies via link elements

### LINKED ISSUE #2002: BuildStream fails to build if there are multiple link dependencies with same target
It was noticed through https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/tree/nanonyme/repro?ref_type=heads that when any element has multiple dependencies that are links with same target, the element will silently not build and it will not be checkoutable since it has not been built.

Deduplication happens correctly with bst show for build-depends, runtime-depends and depends. Full cache key is also identical with one link element and three link elements with same target.

juergbi:
Duplicate dependencies are normally merged already in `extract_depends_from_node()`. However, that only works for dependencies that are declared with the same filename. If a link element is in a dependency list, merging may still be required but can be done only after the link element has been loaded.

The element state updates don't work correctly if there are duplicate dependencies that aren't merged.

Merging `Dependency` objects does not adjust the `name` or `path`. This means that if an element has duplicate dependencies, the name from the first dependency declaration is used. One could argue that we should always use the name of the link target, however, this would break cache keys (via `configure_dependencies()`) and I can't think of a real problem with the current approach.

Fixes #2002.


gtristan:
Will #2002 occur only if there are 2 link elements with the same target *declared in the same project* ?

I'm concerned that we are seeing issues with a toplevel project link to a sub-sub-project element, and a (differently named ?) link from a sub-project to the same sub-sub-project element.

EDIT: If we have the cross project issue, we **can** treat that as a separate issue, but it would be good to have clarity and an open issue.

juergbi:
> Will #2002 occur only if there are 2 link elements with the same target _declared in the same project_ ?

This is about an element having multiple direct dependencies that resolve to the same element. Either multiple links with the same target or a non-link dependency + a link that points to that non-link dependency.

Projects/junctions are not relevant for this issue, as far as I can tell. I can't completely rule out that there is another link element issue in combination with junctions, but I'm not aware of such an issue.

harrysarson:
I tested this branch on the real-world project that was hitting #2002 and this has fixed the "element never builds, cannot artifact checkout" issues.

I have had a read through the patches and they look sensible to me!

gtristan (approved):
LGTM :)

PR #2098, opened 2025-11-28
Support multi source source provenance info

### LINKED ISSUE #2073: Source provenance API for multiple-source source plugins
https://github.com/apache/buildstream/pull/2009 added the possibility to manually annotate sources with additional source provenance information that isn't part of the source configuration, such as the home page and the issue tracker. However, this applies to all SourceInfo produced by the source plugin.

This is problematic for source plugins such as `cargo` which download multiple archives/repositories based on a lock file from a previous source. Each of these archives/repositories needs to have its own provenance information. In addition it would be nice for the plugin to be able to pre-fill some information.

Adding some ideas from @gtristan:

> We could:
>
> * Add a type, e.g. `SourceProvenanceAttrs`, something specifically to describe the `provenance` block parsed by `Source` which represents *"Source provenance values which are explicitly declared by project authors"* 
> * Have a `@classmethod` to create it from a node `SourceProvenanceAttrs.new_from_node()` (similar to what we have for a few types), along with a normal constructor with all keyword arguments for the attributes.
> * Add a `SourceProvenanceAttrs` keyword argument to `Source.create_source_info()`, which *if specified*, is used to provide these attributes rather than the ones specified at the toplevel source.
>
> We could take this one step further and add a class level flag `Source.BST_CUSTOM_SOURCE_PROVENANCE`, which; if set to `True`, causes the *"provenance"* attribute on sources to be an invalid key (preventing project authors from writing toplevel *"provenance"* attributes on sources like `cargo` which support that in a custom way instead) - this could also trigger an assertion in `Source.create_source_info()` if the source fails to provide the attributes.

joshua-zivkovic:
Fixes https://github.com/apache/buildstream/issues/2073

This PR provides a way for multi-source plugins to provide source provenance info in their own way instead of only the top level source source info being used.

This is done by exposing the optional `source_provenance_attrs` argument to `create_source_info()` which when used overrides the usage of the default provenance for a source. Any plugin that needs to parse provenance in it's own way can then supply it's results using this. This has been trialled with the `cargo2` source plugin in https://gitlab.com/BuildStream/buildstream-plugins-community/-/merge_requests/408.

### DIFF
--- a/src/buildstream/source.py
+++ b/src/buildstream/source.py
@@ -759,6 +759,14 @@ class Source(Plugin):
     # The defaults from the project
     __defaults: Optional[Dict[str, Any]] = None
 
+    BST_CUSTOM_SOURCE_PROVENANCE = False
+    """Whether multiple sources' provenance information are provided
+
+    Used primarily to override the usage of top-level source
+    provenance of a source where individual sub-source's
+    provenance should instead be provided
+    """
+
     BST_REQUIRES_PREVIOUS_SOURCES_TRACK = False
     """Whether access to previous sources is required during track
 
@@ -827,6 +835,12 @@ class Source(Plugin):
         self.__provenance: Optional[
             _SourceProvenance
         ] = meta.provenance  # The _SourceProvenance for general user provided SourceInfo
+
+        if self.__provenance is not None and self.BST_CUSTOM_SOURCE_PROVENANCE:
+            raise SourceError(
+                f"{self._get_provenance()} Custom source provenance plugin: Refusing to use top level source provenance",
+                reason="top-level-provenance-on-custom-implementation",
+            )
 
         self.__key = None  # Cache key for source
 
@@ -1368,6 +1382,7 @@ class Source(Plugin):
         *,
         version_guess: Optional[str] = None,
         extra_data: Optional[Dict[str, str]] = None,
+        provenance_node: Optional[MappingNode] = None,
     ) -> SourceInfo:
         """Create a :class:`.SourceInfo` object
 
@@ -1387,14 +1402,22 @@ class Source(Plugin):
            version: A string which represents a unique version of this source input
            version_guess: An optional string representing the guessed human readable version
            extra_data: Additional plugin defined key/values
+           provenance_node: An optional :class:`Node <buildstream.node.Node>` with source provenance attributes,
+                            defaults to the provenance specified at the top level of the source.
 
         *Since: 2.5*
         """
         homepage = None
         issue_tracker = None
-        if self.__provenance is not None:
-            homepage = self.__provenance.homepage
-            issue_tracker = self.__provenance.issue_tracker
+
+        if provenance_node is not None:
+            source_provenance: Optional[_SourceProvenance] = _SourceProvenance.new_from_node(provenance_node)
+        else:
+            source_provenance = self.__provenance
+
+        if source_provenance is not None:
+            homepage = source_provenance.homepage
+            issue_tracker = source_provenance.issue_tracker
 
         return SourceInfo(
             self.get_kind(),
--- a/tests/frontend/show.py
+++ b/tests/frontend/show.py
@@ -814,3 +814,33 @@ def test_source_info_workspace(cli, datafiles, tmpdir):
 
     # There is no version guessing for a workspace
     assert source_info.get_str("version-guess", None) is None
+
+
+@pytest.mark.datafiles(os.path.join(DATA_DIR, "source-info"))
+def test_multi_source_info(cli, datafiles):
+    project = str(datafiles)
+    result = cli.run(
+        project=project, silent=True, args=["show", "--format", "%{name}:\n%{source-info}", "multisource.bst"]
+    )
+    result.assert_success()
+
+    loaded = _yaml.load_data(result.output)
+    sources = loaded.get_sequence("multisource.bst")
+
+    source_info = sources.mapping_at(0)
+    assert source_info.get_str("kind") == "multisource"
+    assert source_info.get_str("url") == "http://ponyfarm.com/ponies"
+    assert source_info.get_str("medium") == "pony-ride"
+    assert source_info.get_str("version-type") == "pony-age"
+    assert source_info.get_str("version") == "1234567"
+    assert source_info.get_str("version-guess", None) == "12"
+    assert "homepage" not in source_info
+
+    source_info = sources.mapping_at(1)
+    assert source_info.get_str("kind") == "multisource"
+    assert source_info.get_str("url") == "http://ponyfarm.com/happy"
+    assert source_info.get_str("medium") == "pony-ride"
+    assert source_info.get_str("version-type") == "pony-age"
+    assert source_info.get_str("version") == "1234567"
+    assert source_info.get_str("version-guess", None) == "12"
+    assert source_info.get_str("homepage") == "http://happy.ponyfarm.com"
--- a/tests/frontend/source-info/elements/multisource.bst
+++ b/tests/frontend/source-info/elements/multisource.bst
@@ -1,0 +1,4 @@
+kind: import
+
+sources:
+- kind: multisource
--- a/tests/sources/source_provenance_attributes/project.conf
+++ b/tests/sources/source_provenance_attributes/project.conf
@@ -1,0 +1,14 @@
+# Project with source provenance attributes
+name: foo
+min-version: 2.0
+
+element-path: elements
+
+aliases:
+  project_dir: file://{project_dir}
+
+plugins:
+- origin: local
+  path: plugins
+  sources:
+  - multisource-plugin
--- a/tests/frontend/source-info/plugins/multisource.py
+++ b/tests/frontend/source-info/plugins/multisource.py
@@ -1,0 +1,48 @@
+from buildstream import Node, Source
+
+
+class MultiSource(Source):
+    BST_MIN_VERSION = "2.0"
+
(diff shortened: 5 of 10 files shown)


src/buildstream/types.py
```diff
@@ -164,6 +164,65 @@ class OverlapAction(FastEnum):
...
+        )
+
+    def serialize(self) -> Dict[str, str]:
+        provenance = {}
+
+        if self.homepage is not None:
+            provenance["homepage"] = self.homepage
+
+        if self.issue_tracker is not None:
+            provenance["issue-tracker"] = self.homepage
```
juergbi:
```suggestion
            provenance["issue-tracker"] = self.issue_tracker
```

juergbi:
Do plugins need access to the actual `SourceProvenance` object or might it be sufficient to make the new `create_source_info()` parameter a `MappingNode` and let `create_source_info()` create a `_SourceProvenance` internally? This would expose less in the public API, but I'm not sure whether that would suffice or not.

joshua-zivkovic:
Yup, taking in to account https://github.com/apache/buildstream/pull/2095#issuecomment-3596947680, there's no need to expose SourceProvenance to achieve this.

juergbi:
I've updated this branch as per my previous comment, but this still requires a test.

src/buildstream/source.py
```diff
@@ -1367,6 +1367,7 @@ def create_source_info(
         version: str,
         *,
         version_guess: Optional[str] = None,
+        provenance_node: Optional[MappingNode] = None,
```
abderrahim: nit: I would add this after extra_data.

src/buildstream/source.py
```diff
@@ -1386,15 +1387,23 @@ def create_source_info(
                          choice depicting the type of version.
            version: A string which represents a unique version of this source input
            version_guess: An optional string representing the guessed human readable version
+           provenance_node: The optional YAML node with source provenance attributes,
```
abderrahim: It's probably better to refer to the Node class rather than saying a "YAML node"

src/buildstream/source.py
```diff
@@ -1386,15 +1387,23 @@ def create_source_info(
...
+                            defaults to the provenance specified at the top level of the source.
```
abderrahim: I'm not sure this silent fallback semantic is good. IIUC, if the plugin provides a provenance for every source info, this will be silently ignored. If the plugin provides multiple source infos, and only some of them set provenance info (e.g. your implementation in https://gitlab.com/BuildStream/buildstream-plugins-community/-/merge_requests/408 if the user doesn't manually modify the ref to include provenance), then source infos that don't set it get the fallback of a global source provenance which is the same for everything.

I'm not sure what would be the best option here, but I feel the current implementation might not be a good idea.

joshua-zivkovic: Ah, that is something that has been completely overlooked yes. I think a "multi-source" plugin should simply have a flag that it sets when used to tell the core to not use top-level source provenance data and actually disallow it (preventing the top level being silently ignored)

src/buildstream/source.py
```diff
@@ -759,6 +759,14 @@ class Source(Plugin):
     # The defaults from the project
     __defaults: Optional[Dict[str, Any]] = None
 
+    BST_MULTI_SOURCE_PROVENANCE = False
```
abderrahim: I would call this CUSTOM or something. While the main use is for sources that have multiple source fetchers that might want to have different source provenance, the *effect* of setting this flag is to allow the source to set its own source provenance programatically and prevent the buildstream core from parsing / using the provenance information from the top-level source configuration.

src/buildstream/source.py
```diff
@@ -1387,14 +1396,32 @@ def create_source_info(
            version: A string which represents a unique version of this source input
            version_guess: An optional string representing the guessed human readable version
            extra_data: Additional plugin defined key/values
+           provenance_node: The optional :class:`Node <buildstream.node.Node>` with source provenance attributes,
```
abderrahim: nit: The -> An

src/buildstream/source.py
```diff
@@ -1387,14 +1396,32 @@ def create_source_info(
...
+            provenance: Optional[_SourceProvenance] = _SourceProvenance.new_from_node(provenance_node)
+        # if the source uses multiple sources but the individual source_info has
+        # no specified provenance data, skip it. We don't want to use top level
+        # data as a fallback because it'll just be wrong
+        elif self.BST_MULTI_SOURCE_PROVENANCE:
+            if self.__provenance is not None:
+                # This warning is printed for each occurrence of missing provenance
+                # for each sub-source, maybe it should be done just the once per
+                # multi-source?
+                self.warn("Multi source plugin: Not using top level source provenance")
```
abderrahim: As a follow-up to my comment above, this shouldn't call them "Multi source plugin".

Also I'm not sure if this should be a warning or an error. I feel we should be strict with this, especially if this check runs early on in the pipeline (i.e. it's not only run if are looking for source info) 

And please use the provenance (as in the `ProvenanceInformation`, not the `SourceProvenance`) in the error/warning message.

tests/frontend/show.py
```diff
@@ -814,3 +814,33 @@ def test_source_info_workspace(cli, datafiles, tmpdir):
...
+    sources = loaded.get_sequence("multisource.bst")
+
+    source_info = sources.mapping_at(0)
+    assert source_info.get_str("kind") == "multisource"
+    assert source_info.get_str("url") == "http://ponyfarm.com/ponies"
+    assert source_info.get_str("medium") == "pony-ride"
+    assert source_info.get_str("version-type") == "pony-age"
+    assert source_info.get_str("version") == "1234567"
+    assert source_info.get_str("version-guess", None) == "12"
+    assert source_info.get_str("homepage") == "https://flying-ponies.com/index.html"
```
abderrahim: Is this test up to date? Shouldn't the homepage here be unset?

(also should we test for the warning/error condition?)

src/buildstream/source.py
juergbi: I think we should indeed make this an error and move it to the `Source` constructor. I don't see a backward compatibility concern as there are no `BST_CUSTOM_SOURCE_PROVENANCE` source plugins yet. If needed, we could potentially be less strict in the future, while the other way round would be impossible.

tests/frontend/show.py
joshua-zivkovic: Yes, I'm currently updating the tests to match the new behaviour

tests/frontend/source-info/elements/multisource.bst
```diff
@@ -0,0 +1,7 @@
+kind: import
+
+sources:
+- kind: multisource
+  ref:
+    - provenance:
+      homepage: https://flying-ponies.com/index.html
```
juergbi: I don't think this is used, is it? An additional test with a test plugin that uses provenance from `ref` could be useful, but specifying it here without being used is confusing. I consider such an additional test a blocker as nice-to-have but not a blocker. If you add such a test, please keep the current test using `Node.from_dict` as we want to make sure that a plugin can also programmatically create provenance information from non-ref metadata.

tests/sources/source_provenance_attributes/project.conf
```diff
@@ -0,0 +1,11 @@
+# Project with source provenance attributes
+name: foo
+min-version: 2.0
+
+element-path: elements
+
+# Don't allow any provenance
+source-provenance-fields: {}
```
juergbi: I think that test slipped in from the other PR.

src/buildstream/source.py
```diff
@@ -828,6 +836,12 @@ def __init__(
             _SourceProvenance
         ] = meta.provenance  # The _SourceProvenance for general user provided SourceInfo
 
+        if self.__provenance is not None and self.BST_CUSTOM_SOURCE_PROVENANCE:
+            raise SourceError(
+                f"{self._get_provenance()} Custom source plugin: Refusing to use top level source provenance",
```
juergbi: "Custom source plugin" is not clear to me. Did you mean "Custom source provenance" or maybe "Custom source provenance plugin"?

joshua-zivkovic: Ah, that is definitely not what I meant to put :D

tests/sources/source_provenance_attributes/project.conf
joshua-zivkovic: Yup, I'm about to push a fix

tests/frontend/source-info/elements/multisource.bst
joshua-zivkovic: I'm not sure I entirely see the benefit of having a specific test for provenance being provided from `ref`, actually. It isn't something that is explicitly supported.

juergbi: I'm fine with skipping that.

tests/frontend/show.py
juergbi: The incorrect homepage check has been removed. I'd suggest adding
```
assert "homepage" not in source_info
```
to check that the homepage is actually unset.

juergbi (approved):

src/buildstream/source.py
joshua-zivkovic: I believe this has been resolved with the `BST_CUSTOM_SOURCE_PROVENANCE` flag being introduced to block usage of top level provenance definitions.

tests/frontend/show.py
joshua-zivkovic: Added this in too, everything is happy.

PR #2099, opened 2026-01-07
Generic Source Provenance

joshua-zivkovic:
Towards https://github.com/apache/buildstream/issues/2069

This PR changes the approach to how buildstream handles source provenance information, moving from a hard coded implementation of attributes (which buildstream core needs to know nothing about) to a more generic, user defined approach where buildstream only needs to carry the values to the next step.

Users can now define their projects allowed set of source provenance attributes using the `source_provenance_fields` key in the project.conf. Where attributes are used in source provenance but not defined in the project.conf, buildstream will refuse attribute and throw an error informing the user that the attribute is not present in the project.conf

Currently there is no limitation on these values since neither plugins nor buildstream sbom are yet using this approach so there is no defined set of attributes plugins and users should refer to but this will soon be added to buildstream-sbom as it will depend on what buildstream-sbom supports for SPDX generation.

Note: This PR previously included updates to `source track` to allow plugins to automatically populate these source attributes where possible; this has been moved to a follow up PR.


src/buildstream/source.py
```diff
@@ -927,7 +929,16 @@ def set_ref(self, ref, node):
...
 
-    def track(self, *, previous_sources_dir: Optional[str] = None) -> SourceRef:
+    def load_source_provenance(self, node: MappingNode) -> None:
+        raise ImplError("Source plugin '{}' does not implement load_source_provenance()".format(self.get_kind()))
+
+    def get_source_provenance(self) -> SourceProvenance:
+        raise ImplError("Source plugin '{}' does not implement get_source_provenance()".format(self.get_kind()))
+
+    def set_source_provenance(self, source_provenance: SourceProvenance):
+        raise ImplError("Source plugin '{}' does not implement set_source_provenance()".format(self.get_kind()))
```
juergbi: To retain backward compatibility, we can't require source plugins to implement these methods. I.e., we need (trivial) default implementations for any method that is called unconditionally by the core. API documentation will be required as well, of course.

src/buildstream/source.py
```diff
@@ -927,7 +929,16 @@ def set_ref(self, ref, node):
...
+
+    def track(self, *, previous_sources_dir: Optional[str] = None) -> (SourceRef, SourceProvenance):
```
juergbi: We need to find a way to allow returning a `SourceProvenance` without breaking API.

src/buildstream/source.py
joshua-zivkovic: That's easy enough if we just go with a "do nothing" implementation

src/buildstream/source.py
joshua-zivkovic: Hmm, that might be more difficult to work out. I'll see if I can work something sensible out

joshua-zivkovic: For now I've gone with combining the old and new return signatures (https://github.com/apache/buildstream/pull/2099/files#diff-48200ef302e5d841cce81834e8879a65e59db922a22605b9d7fa16b4dd5c7067R941) and added some logic to unpackage it appropriately (https://github.com/apache/buildstream/pull/2099/files#diff-48200ef302e5d841cce81834e8879a65e59db922a22605b9d7fa16b4dd5c7067R1811). I believe that's all that's required to not break the API here

src/buildstream/element.py
```diff
@@ -2635,19 +2635,33 @@ def __load_sources(self, load_element):
...
                 # Provenance is optional
-                provenance_node = source.get_mapping(Symbol.PROVENANCE, default=None)
-                provenance = None
+                provenance_node: MappingNode = source.get_mapping(Symbol.PROVENANCE, default=None)
                 if provenance_node:
                     del source[Symbol.PROVENANCE]
-                    provenance = _SourceProvenance.new_from_node(provenance_node)
+                    defined_attrs = (
+                        project._project_conf.get_mapping("source-provenance-fields", None)
+                        or project.source_provenance_fields
```
juergbi: Is this the logic to handle the default on a project level? I'm not sure whether this works correctly, however, this logic should in any case be in the `Project` class such that `Element` and `Source` can simply use `project.source_provenance_fields` (`_project.conf` is considered private and should not be accessed from other classes). Or am I missing a reason why this can't be done in the `Project` class?

juergbi:
Have you already started looking into adding test(s) for this?

src/buildstream/source.py
```diff
@@ -1393,23 +1410,33 @@ def create_source_info(
...
+            try:
+                provenance_node.validate_keys(defined_provenance_fields.keys())
+            except LoadError as E:
+                raise LoadError(
+                    "Specified source attribute not defined in project config\n {}".format(E),
+                    LoadErrorReason.UNDEFINED_SOURCE_PROVENANCE_ATTRIBUTE,
+                )
+
+            # Make sure everything is a string
+            provenance = MappingNode.from_dict({key: value.as_str() for key, value in provenance_node.items()})
```
juergbi: Is there a reason we actually need to use a newly created `MappingNode` instead of directly using `provenance_node` (after verification)? I suspect that this recreation loses provenance of the provenance node (i.e., `.bst` file name and line number).

src/buildstream/source.py
```diff
@@ -572,14 +590,9 @@ def __init__(
         The url of the source input
         """
 
-        self.homepage: Optional[str] = homepage
-        """
-        The project homepage URL
-        """
-
-        self.issue_tracker: Optional[str] = issue_tracker
```
juergbi, line 580: These attributes are public, we can't drop them. We need to populate them from generic source provenance for backwards compatibility.

src/buildstream/element.py
joshua-zivkovic: Thanks, I've made the change to move the logic to `Project` instead

src/buildstream/source.py
joshua-zivkovic: I wasn't aware I could simply write back to a `MappingNode` like a dictionary, even though the docs clearly state that. Regardless, this is no longer being used since this logic has been changed to throwing an error in the event that the value could not be parsed as a string

src/buildstream/source.py
joshua-zivkovic: Sorry, I keep messing up the backwards compatibility. I've added these back in

src/buildstream/element.py
```diff
@@ -2635,19 +2635,29 @@ def __load_sources(self, load_element):
...
+                    except LoadError as E:
+                        raise LoadError(
+                            "Specified source attribute not defined in project config\n {}".format(E),
+                            LoadErrorReason.UNDEFINED_SOURCE_PROVENANCE_ATTRIBUTE,
+                        )
+
+                    # make sure everything is a string
+                    for value in provenance_node.values():
+                        if not isinstance(value, ScalarNode):
+                            raise LoadError(f"{value} is not parsable as a string", LoadErrorReason.INVALID_DATA)
```
juergbi: I'd suggest adding context to make the error easier to understand. Maybe something like the following?

```suggestion
                    for key, value in provenance_node.items():
                        if not isinstance(value, ScalarNode):
                            raise LoadError(f"{value}: Expected string for the value of provenance attribute '{key}'", LoadErrorReason.INVALID_DATA)
```

The same applies to the similar check in `source.py`.

src/buildstream/element.py
```diff
@@ -2635,19 +2635,29 @@ def __load_sources(self, load_element):
...
-                provenance = None
+                provenance_node: MappingNode = source.get_mapping(Symbol.PROVENANCE, default=None)
                 if provenance_node:
                     del source[Symbol.PROVENANCE]
-                    provenance = _SourceProvenance.new_from_node(provenance_node)
+                    try:
+                        provenance_node.validate_keys(project.source_provenance_fields.keys())
+                    except LoadError as E:
+                        raise LoadError(
+                            "Specified source attribute not defined in project config\n {}".format(E),
```
juergbi:
```suggestion
                            "Specified source provenance attribute not defined in project config\n {}".format(E),
```

juergbi (commented):
Somewhat of a nit but it seems we're mixing two terms for the same thing. I'm not sure whether "source provenance field" or "source provenance attribute" is a better fit. However, whatever we choose, we should be consistent to avoid confusion. Or is there a semantic distinction that I'm overlooking?

I think I used the term 'field' first because I thought the same term was used in the documentation already, but I may be mistaken. I'm not dead-set on the 'field' term. If 'attribute' is a better fit with the rest of BuildStream's API and documentation, I'm happy to settle on that as long as we're consistent.

joshua-zivkovic:
> Somewhat of a nit but it seems we're mixing two terms for the same thing. I'm not sure whether "source provenance field" or "source provenance attribute" is a better fit. However, whatever we choose, we should be consistent to avoid confusion. Or is there a semantic distinction that I'm overlooking?

I've noticed this too and haven't got a fixed opinion on which we use. I was just using "attributes" since https://github.com/apache/buildstream/issues/2069 uses the term. I'm not seeing any deciding usage in the docs.

In context of errors that already exist, it seems "attribute" reads better than "field", to me at least,  but that's just how I've been reading it all this time anyway.

I'll change up the error messages as per your suggestions and make the term used consistent based on what reads best from there

joshua-zivkovic:
I've seemingly messed something up with the source provenance attribute tests so they're currently failing. I'll look at this next week

src/buildstream/source.py
```diff
@@ -1387,14 +1402,22 @@ def create_source_info(
...
-            issue_tracker = self.__provenance.issue_tracker
+
+        if provenance_node is not None:
+            source_provenance: Optional[_SourceProvenance] = _SourceProvenance.new_from_node(provenance_node)
+        else:
+            source_provenance = self.__provenance
+
+        if source_provenance is not None:
+            homepage = source_provenance.homepage
+            issue_tracker = source_provenance.issue_tracker
```
franred: if all fields are optional in source_provenance, should we not check they exist before trying to assign them?

franred: I can see this has been changed in a future commit. https://github.com/apache/buildstream/pull/2099/commits/01e55e87c2f530337b81deffc4fb9544039b3445#diff-48200ef302e5d841cce81834e8879a65e59db922a22605b9d7fa16b4dd5c7067R1457

joshua-zivkovic: Yup, the PR (https://github.com/apache/buildstream/pull/2098) this is currently based on uses the old `_SourceProvenance` class, which defaults them to `None` if they're not given a value

src/buildstream/source.py
```diff
@@ -579,7 +588,12 @@ def __init__(
 
         self.issue_tracker: Optional[str] = issue_tracker
         """
-        The project issue tracking URL
+        The issue tracker for the project
```
juergbi: Was this an intentional change? I think having "URL" in the description was clearer.

src/buildstream/source.py
```diff
@@ -396,6 +402,8 @@
 
     # pylint: enable=cyclic-import
 
+SourceProvenance = MappingNode
```
juergbi: I'm not sure whether this alias is actually helpful or may be more confusing as it's not a distinct type. It doesn't seem like we're doing this for other mapping nodes. What advantage do you see with this?

src/buildstream/source.py
```diff
@@ -38,15 +38,11 @@
...
-  * Homepage
-
-    The ``homepage`` attribute can be used to specify the project homepage URL
-
-  * Issue Tracker
-
-    The ``issue-tracker`` attribute can be used to specify the project's issue tracking URL
+  Any attribute used in the ``provenance`` dictionary of a source must be
+  defined in the project.conf using the ``source-provenance-attributes`` dictionary
+  to define the attribute and its significance.
```
juergbi, line 45: I think we should point to 
```
:ref:`builtin defaults <project_builtin_defaults>`
```
in case project.conf doesn't define a `provenance` dictionary.

src/buildstream/source.py
joshua-zivkovic: I wanted to try and be clearer that something was for `SourceProvenance` rather than, somewhat confusingly, just any `MappingNode`. This obviously isn't entirely fool proof though and since there's no other cases of this I suppose it's sensible to just remove it

src/buildstream/source.py
joshua-zivkovic: I've added this. I'm not sure if we want mention that this behaviour is for backwards compatibility or not.

src/buildstream/source.py
joshua-zivkovic: I've changed this back. I don't remember any specifics about this change so it must have slipped through

PR #2102, opened 2026-01-29
_stream: Don't require runtime deps when building with --deps none

abderrahim:
Before this change, running a build with `--deps none` would also build / pull runtime dependencies. This commit changes the elements on the command line to not require their runtime dependencies if we're using `--deps none`.

In addition to the config file option dependencies `none` and `all`, this PR adds `run` and makes it the
default instead of `none`. The previous commit changed the meaning of
`--deps none` to be different from `--deps run`, so we change the default to be
`run`.

### DIFF
--- a/doc/source/using_config.rst
+++ b/doc/source/using_config.rst
@@ -381,6 +381,7 @@ Attributes
   values for this attribute are:
 
   * ``none``: Only build elements required to generate the expected target artifacts
+  * ``run``: Build required elements and their their runtime dependencies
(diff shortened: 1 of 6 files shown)


juergbi:
The proposed behavior seems reasonable. Looking at the documentation, the current `none` behavior might not have been intentional. It is technically still a breaking change from the CLI/config perspective but it may be acceptable for 2.7 given the default user config update which mitigates the change (and it won't break project config or element definitions).

I'd like to see at least minimal test coverage, though (could be a current test that is slightly expanded).

juergbi (approved):

PR #2104, opened 2026-02-06
Correctly track session elements in dynamic build plan

abderrahim:
When using a dynamic build plan, all elements are passed to the queue to be enqueued, but are only actually enqueued when they become required.

This moves the tracking of session elements to the first queue rather than just take everything that we pass to the queue (which would be all elements).

This is only used in the UI AFAICT, and doesn't actually affect the build.

### DIFF
--- a/src/buildstream/_scheduler/queues/queue.py
+++ b/src/buildstream/_scheduler/queues/queue.py
@@ -62,6 +62,7 @@ class Queue:
     # Resources this queues' jobs want
     resources = []  # type: List[int]
     log_to_file = True
+    session_elements = None
 
     def __init__(self, scheduler, *, imperative=False):
 
@@ -171,6 +172,15 @@ class Queue:
     #####################################################
     #          Scheduler / Pipeline facing APIs         #
     #####################################################
+
+    # set_session_elements()
+    #
+    # Track elements enqueued
+    #
+    # Args:
+    #    session_elements (list): a list to put session elements
+    def set_session_elements(self, session_elements):
+        self.session_elements = session_elements
 
     # enqueue()
     #
@@ -381,6 +391,9 @@ class Queue:
     #    element (Element): The Element to enqueue
     #
     def _enqueue_element(self, element):
+        if self.session_elements is not None:
+            self.session_elements.append(element)
+
         status = self.status(element)
 
         if status == QueueStatus.SKIP:
--- a/src/buildstream/_scheduler/queues/queue.py
+++ b/src/buildstream/_scheduler/queues/queue.py
@@ -62,6 +62,7 @@ class Queue:
     # Resources this queues' jobs want
     resources = []  # type: List[int]
     log_to_file = True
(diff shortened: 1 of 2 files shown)


src/buildstream/_stream.py
```diff
@@ -1828,7 +1828,6 @@ def _add_queue(self, queue, *, track=False):
     def _enqueue_plan(self, plan, *, queue=None):
         queue = queue or self.queues[0]
         queue.enqueue(plan)
-        self.session_elements += plan
```
juergbi, line 1833: The frontend still uses `Stream.session_elements`, which is now always empty.

I don't see how this branch can work as it is, but I could be missing something.

abderrahim: We pass the list to the queue a bit earlier

```py
            queue.set_session_elements(self.session_elements)
```

which then fills it as elements are getting enqueued.

Yeah, it's a bit confusing but I couldn't find a better way to do it.

abderrahim: I guess the alternative is to make Element support more than one callback when it becomes required? This way both the Stream and the Queue can pass their own callbacks?

juergbi: Ah, it's just a reference to the list created in `Stream`. I think `set_session_elements()` threw me off as that doesn't actually set any elements, it just passes a reference to the list. With clearer naming and/or comment, it might be ok.

abderrahim: Is this better?

src/buildstream/_stream.py
```diff
@@ -1816,6 +1816,7 @@ def _add_queue(self, queue, *, track=False):
         if not track and not self.queues:
             # First non-track queue
             queue.set_required_element_check()
+            queue.set_session_elements(self.session_elements)
```
juergbi, line 1821: This is never executed for `bst source track`, resulting in the frontend indicating 0 session elements.

`query_cache` marks the `CacheQueryQueue` (with `storage-service`) also as `track=True`. Need to make sure that fixing `bst source track` won't make the counter wrong in sessions with a `CacheQueryQueue`.

src/buildstream/_scheduler/queues/queue.py
```diff
@@ -381,6 +393,9 @@ def _element_log_path(self, element):
     #    element (Element): The Element to enqueue
     #
     def _enqueue_element(self, element):
+        if self.session_elements is not None:
+            self.session_elements.append(element)
```
juergbi: `_enqueue_element` seems to be called also for retried jobs. I haven't tested it but this might increase the counter too much. Maybe `session_elements` should be a set instead? (May need to use `OrderedDict` as the order may be significant).

src/buildstream/_stream.py
abderrahim: Fixed by using a different condition here.

juergbi (approved):

PR #2112, opened 2026-04-17
source.py: Fix mirror usage in subprojects

### LINKED ISSUE #2110: Buildstream Mirroring Bug
At Codethink we have a buildstream container project which junctions from freedesktop-sdk and freedesktop-sdk-extra, each using buildstream-plugins-community and all mirrored per project. What we saw is that, depending on the buildstream-plugins-community version, this may or may not work. What I mean here by "not work" is that, despite mirroring being in place, it suddenly tried to download the fdsdk-extra buildstream-plugins-community version being 2.3.0 from pypi rather than the mirror.

Unfortunately, for now I do not have a minimal reproducer ready for this, but wanted to report this upstream before leaving for a two weeks vacation. Internally at Codethink, I will also file a container bug referencing all information available at this time. Thanks!

juergbi:
Don't skip source cloning if there are no mirrors configured in the element's project. With alias mapping, the mirror configuration of the parent project matters.

For `fetch()`, this bug only affected source plugins that don't use `SourceFetcher` objects. For `track()`, all source plugins were affected.

Fixes #2110.

### DIFF
--- a/src/buildstream/source.py
+++ b/src/buildstream/source.py
@@ -2036,11 +2036,7 @@ class Source(Plugin):
         #
         else:
             alias = self._get_alias()
-            if self.__first_pass:
-                mirrors = project.first_pass_config.mirrors
-            else:
-                mirrors = project.config.mirrors
-            if not mirrors or not alias:
+            if not alias:
                 self.fetch(**kwargs)
                 return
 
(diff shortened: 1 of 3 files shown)


gtristan (approved):
Good spot !

abderrahim (approved):

PR #2119, opened 2026-05-01
Defer source provenance check until the project is fully loaded

### LINKED ISSUE #2116: Junction element with provenance node triggers crash
With BuildStream 2.7.0 (#2099), a junction element with a provenance node such as

```
kind: junction

sources:
- kind: local
  path: .
  provenance:
    originator: bar
```
results in the following crash for a simple `bst show`:
```
08:41:23 [--:--:--][        ][    main:core activity                 ] START   Loading elements
08:41:23 [00:00:00][        ][    main:core activity                 ] SUCCESS Loading elements
08:41:23 [--:--:--][        ][    main:core activity                 ] START   Resolving elements
08:41:23 [--:--:--][        ][    main:core activity                 ] BUG     'NoneType' object has no attribute 'keys'

      File "buildstream/.venv/bin/bst", line 10, in <module>
        sys.exit(cli())
                 ~~~^^
      File "buildstream/.venv/lib/python3.14/site-packages/click/core.py", line 1485, in __call__
        return self.main(*args, **kwargs)
               ~~~~~~~~~^^^^^^^^^^^^^^^^^
      File "buildstream/src/buildstream/_frontend/cli.py", line 282, in override_main
        original_main(self, args=args, prog_name=prog_name, complete_var=None, standalone_mode=standalone_mode, **extra)
        ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "buildstream/.venv/lib/python3.14/site-packages/click/core.py", line 1406, in main
        rv = self.invoke(ctx)
      File "buildstream/.venv/lib/python3.14/site-packages/click/core.py", line 1873, in invoke
        return _process_result(sub_ctx.command.invoke(sub_ctx))
                               ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
      File "buildstream/.venv/lib/python3.14/site-packages/click/core.py", line 1269, in invoke
        return ctx.invoke(self.callback, **ctx.params)
               ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "buildstream/.venv/lib/python3.14/site-packages/click/core.py", line 824, in invoke
        return callback(*args, **kwargs)
      File "buildstream/.venv/lib/python3.14/site-packages/click/decorators.py", line 46, in new_func
        return f(get_current_context().obj, *args, **kwargs)
      File "buildstream/src/buildstream/_frontend/cli.py", line 661, in show
        dependencies = app.stream.load_selection(
            elements, selection=deps, except_targets=except_, need_state=need_state
        )
      File "buildstream/src/buildstream/_stream.py", line 168, in load_selection
        target_objects = self._load(
            targets,
        ...<9 lines>...
            need_state=need_state,
        )
      File "buildstream/src/buildstream/_stream.py", line 1684, in _load
        elements, except_elements, artifacts = self._load_elements_from_targets(
                                               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
            targets, except_targets, rewritable=False, valid_artifact_names=load_artifacts
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        )
        ^
      File "buildstream/src/buildstream/_stream.py", line 1497, in _load_elements_from_targets
        elements, except_elements = self._load_elements([element_names, except_targets])
                                    ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "buildstream/src/buildstream/_stream.py", line 1447, in _load_elements
        elements = self._project.load_elements(targets)
      File "buildstream/src/buildstream/_project.py", line 520, in load_elements
        elements = [Element._new_from_load_element(load_element, task) for load_element in load_elements]
                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^
      File "buildstream/src/buildstream/element.py", line 1075, in _new_from_load_element
        element.__load_sources(load_element)
        ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^
      File "buildstream/src/buildstream/element.py", line 2642, in __load_sources
        provenance_node.validate_keys(project.source_provenance_attributes.keys())
                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    AttributeError: 'NoneType' object has no attribute 'keys'
```

As far as I can tell, the issue is that `project.source_provenance_attributes` is set only as part of the second project loading pass while junction elements may already be loaded before the second project loading pass.

juergbi:
Junctions may be loaded before a project is fully loaded, which means
that the source provenance attribute project configuration may not be
available yet.

This adds a callback mechanism to defer the source provenance attribute
check if the project is not fully loaded when a junction element with a
provenance node is loaded.

      File "src/buildstream/element.py", line 1075, in _new_from_load_element
        element.__load_sources(load_element)
        ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^
      File "src/buildstream/element.py", line 2642, in __load_sources
        provenance_node.validate_keys(project.source_provenance_attributes.keys())
                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    AttributeError: 'NoneType' object has no attribute 'keys'

Fixes #2116.

### DIFF
--- a/src/buildstream/_context.py
+++ b/src/buildstream/_context.py
@@ -565,6 +565,13 @@ class Context:
         #
         return self._projects[0]
 
+    # ensure_fully_loaded():
+    #
+    # Ensure all projects are fully loaded.
+    def ensure_fully_loaded(self) -> None:
+        for project in self._projects:
+            project.ensure_fully_loaded()
+
     # initialize_remotes()
     #
     # This will resolve what remotes each loaded project will interact
@@ -592,9 +599,7 @@ class Context:
         ignore_project_source_remotes: bool = False,
     ) -> None:
 
-        # Ensure all projects are fully loaded.
-        for project in self._projects:
-            project.ensure_fully_loaded()
+        self.ensure_fully_loaded()
 
         #
         # If the global remote execution specs have been overridden by the
--- a/src/buildstream/_project.py
+++ b/src/buildstream/_project.py
@@ -15,7 +15,7 @@
 #        Tristan Van Berkom <tristan.vanberkom@codethink.co.uk>
 #        Tiago Gomes <tiago.gomes@codethink.co.uk>
 
-from typing import TYPE_CHECKING, Optional, Dict, Union, List, Sequence
+from typing import TYPE_CHECKING, Optional, Dict, Union, List, Sequence, Callable
 
 import os
 import urllib.parse
@@ -157,6 +157,8 @@ class Project:
         self._partially_loaded: bool = False
         self._fully_loaded: bool = False
         self._project_includes: Optional[Includes] = None
+
+        self._fully_loaded_callbacks: List[Callable[[], None]] = []
 
         #
         # Initialization body
@@ -692,6 +694,17 @@ class Project:
     def loaded_projects(self):
         yield from self.load_context.loaded_projects()
 
+    # register_fully_loaded_callback()
+    #
+    # Call the specified function once the project is fully loaded.
+    #
+    # Args:
+    #    callback (Callable[[], None]): A function to call once fully loaded
+    #
+    def register_fully_loaded_callback(self, callback: Callable[[], None]):
+        assert not self._fully_loaded
+        self._fully_loaded_callbacks.append(callback)
+
     ########################################################
     #                    Private Methods                   #
     ########################################################
@@ -1016,6 +1029,10 @@ class Project:
         self.source_provenance_attributes = project_conf_second_pass.get_mapping(
             "source-provenance-attributes", None
         ) or config.get_mapping("source-provenance-attributes")
+
+        for callback in self._fully_loaded_callbacks:
+            callback()
+        self._fully_loaded_callbacks = None
 
     # _load_pass():
     #
--- a/src/buildstream/_stream.py
+++ b/src/buildstream/_stream.py
@@ -1684,6 +1684,7 @@ class Stream:
         elements, except_elements, artifacts = self._load_elements_from_targets(
             targets, except_targets, rewritable=False, valid_artifact_names=load_artifacts
         )
+        self._context.ensure_fully_loaded()
 
         if artifacts:
             if selection in (_PipelineSelection.ALL, _PipelineSelection.RUN):
--- a/src/buildstream/element.py
+++ b/src/buildstream/element.py
@@ -2638,13 +2638,21 @@ class Element(Plugin):
                 provenance_node: MappingNode = source.get_mapping(Symbol.PROVENANCE, default=None)
                 if provenance_node:
                     del source[Symbol.PROVENANCE]
-                    try:
-                        provenance_node.validate_keys(project.source_provenance_attributes.keys())
-                    except LoadError as E:
-                        raise LoadError(
-                            f"Specified source provenance attribute not defined in project config\n {E}",
-                            LoadErrorReason.UNDEFINED_SOURCE_PROVENANCE_ATTRIBUTE,
-                        )
+
+                    def source_provenance_attribute_check(provenance_node=provenance_node):
+                        try:
+                            provenance_node.validate_keys(project.source_provenance_attributes.keys())
+                        except LoadError as E:
+                            raise LoadError(
+                                f"Specified source provenance attribute not defined in project config\n {E}",
+                                LoadErrorReason.UNDEFINED_SOURCE_PROVENANCE_ATTRIBUTE,
+                            )
+
+                    # Source provenance attributes are available only after the project is fully loaded
+                    if project.source_provenance_attributes is None:
+                        project.register_fully_loaded_callback(source_provenance_attribute_check)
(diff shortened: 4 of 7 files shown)


abderrahim:
I'm not sure what to think of this. The implementation and the reasoning behind it look correct, but I'm not a fan of adding a callback like this.

I'm not very knowledgeable about the loading process, so correct me if my understanding is wrong. The usual solution for this kind of things that we want to use at load time is to load them early, so that we can have them already when loading junctions. But for this particular case, we want to avoid that to allow these source provenance attributes to be loaded using an include?

I wonder if it wouldn't be nicer to have some regular Plugin method that gets called on second pass, rather than having to register a callback.

Conceptually, I think this check needs to happen at plugin configure time rather than at plugin load time. Maybe we can do that instead by having `Source` override `Plugin._configure()` to make this check?

Sorry if I don't sound coherent, I wrote this as I'm trying to understand and didn't edit it afterwards because I wanted to document my chain of thoughts.

juergbi:
> I'm not very knowledgeable about the loading process, so correct me if my understanding is wrong. The usual solution for this kind of things that we want to use at load time is to load them early, so that we can have them already when loading junctions. But for this particular case, we want to avoid that to allow these source provenance attributes to be loaded using an include?

Yes. Unfortunately, the combination of includes from project.conf and junctions makes the whole loading process very tricky.

I would also have preferred to do it without a callback. However, the junction/include combination already now relies on not loading pass 2 of projects earlier than necessary. The callback is the simplest approach I could think of that works with the current architecture/constraints (and hopefully not rendering a currently valid project invalid). Assuming we do need to support include files for source provenance.

> Conceptually, I think this check needs to happen at plugin configure time rather than at plugin load time. Maybe we can do that instead by having Source override Plugin._configure() to make this check?

`Source._configure()` is called by the constructor, so I don't see a meaningful difference there. Or am I misunderstanding your suggestion?

The basic issue is that sources of junctions need to be fully functional (fetch) even if the second loading pass of the project is not done yet (the source provenance attributes might even be defined in an include of the very subproject this junction points to). Based on that, I don't see a possibility to move the check to some other already existing method.

An alternative to the callback could be to add e.g. `Element._second_pass_done()`, which is called unconditionally by the project, instead of invoking a registered callback. However, that currently sounds worse to me.

It's certainly possible that I'm overlooking a better option, of course. Can you think of anything else?

juergbi:
Another option could be to hand over the provenance nodes to the `Project` object and move the check to the `Project` class, invoked at the end of the second loading pass. Should functionally be equivalent, just implemented without callback. Not sure whether it's really better, though.

abderrahim:
Right, I was confused. The way it is actually handled is that we have two different codepaths for first and second pass. E.g. 

https://github.com/apache/buildstream/blob/5349c6c84b23c31fcaee98a2e6101a0aa2668258/src/buildstream/source.py#L2089-L2092

https://github.com/apache/buildstream/blob/5349c6c84b23c31fcaee98a2e6101a0aa2668258/src/buildstream/_project.py#L220-L223

However I'm not sure how this works. Is it the case that junctions only ever use the first pass config, and other elements use the second pass? If this is the case, I guess we don't have another option.

juergbi:
> Is it the case that junctions only ever use the first pass config, and other elements use the second pass?

Yes (+ link elements):

https://github.com/apache/buildstream/blob/5349c6c84b23c31fcaee98a2e6101a0aa2668258/src/buildstream/_loader/loadelement.pyx#L277

juergbi:
@gtristan Any thoughts on this?

juergbi:
@abderrahim Do you have any functional concerns or is your only concern that the callback might not be the the optimal implementation choice?

As the callback mechanism is an implementation detail, not part of the public API (and also not a lot of code), I think we should get this merged as is, if there are no functional concerns. We can always tweak the implementation later if we can come up with a better approach.

If you think the following alternative I suggested in an earlier comment would be significantly better, I'm happy to give this a try. But could also be done as potential follow-up instead of blocking this bug fix.

> Another option could be to hand over the provenance nodes to the `Project` object and move the check to the `Project` class, invoked at the end of the second loading pass. Should functionally be equivalent, just implemented without callback. Not sure whether it's really better, though.

abderrahim:
Yeah, I'm mostly concerned that this might make the implementation even harder to follow. Feel free to merge this, and tweak it later.

abderrahim (approved):

PR #2103, opened 2026-02-05
docs: Move BuildElement's "Built-in functionality" to new doc page

shymega-ct:
This is a draft PR that relates to https://github.com/apache/buildstream/pull/2060#discussion_r2709087741.

PR #2122, opened 2026-05-08
docs: Move BuildElement's "Built-in functionality" to a new page

juergbi:
The documentation of how to use the common features provided by the `BuildElement` base class is currently part of the API reference even though it's more relevant to project authors than plugin developers.

This moves that section to a new page in the "Reference" section, next to other plugin-specific documentation.

The main content of the new page is unchanged but I've tweaked the initial paragraphs to make sense with the structure change.

This is a replacement for #2103 with proper links and TOC entry (a bit quirky with Sphinx but I think it looks and works reasonably well now).


juergbi:
From https://github.com/apache/buildstream/pull/2103#pullrequestreview-3897891844 (@tmewett):
> My 2c: the main issue I have is that BuildElement is an API name, and if we're trying to document this for users it might make sense to frame this in a different way.

I've linked the new page as "Build Elements" in the TOC and tried to make the intro paragraph a bit more understandable for project authors. If you have any suggestions how to further improve that, please let me know.

doc/source/core_plugins.rst
```diff
@@ -27,8 +27,21 @@ given element specify their plugin specific configuration directly
...
 --------
+
+ * :doc:`elements/stack`
+ * :doc:`elements/import`
+ * :doc:`elements/compose`
+ * :doc:`elements/script`
+ * :doc:`elements/link`
+ * :doc:`elements/junction`
+ * :doc:`elements/filter`
+ * :doc:`Build Elements <core_buildelement>`
```
abderrahim, line 38: This feels a bit hidden here, but I can't think of a better place to put it.

abderrahim (approved):

doc/source/core_plugins.rst
juergbi: Yes, it doesn't seem ideal, but it's the best I could come up with and should definitely still be more discoverable than the status quo. If we can think of a better solution in the future, we can improve it further.

PR #2125, opened 2026-05-20
fix: use full element names in workspace reset

kotborealis:
Fixes a bug where `workspace reset` did not work for junctions.

For example, we have a project with struct like this:
```
nested.bst:hello.bst
```

So the root project has one element `nested.bst`, which is a junction to another project, and the junction has one element `hello.bst`.

If we try to open this element as workspace, it works:
```
bst workspace open nested.bst:hello.bst
```

If we try to close this workspace, it works:
```
bst workspace close nested.bst:hello.bst
```

However, if we try `bst workspace reset` command on it, it fails with error `Workspace does not exist`.
The error is that `workspace_reset` method called `workspace_exists` with `element.name`, which was resolved to just `hello.bst`. In comparsion, a few lines down, it uses `get_workspace` with `element._get_full_name()`, which is `nested.bst:hello.bst`.

This patch updates `workspace_reset` function to check existence using element full name to support junctions.

### DIFF
--- a/src/buildstream/_stream.py
+++ b/src/buildstream/_stream.py
@@ -1152,8 +1152,9 @@ class Stream:
 
         nonexisting = []
         for element in elements:
-            if not self.workspace_exists(element.name):
-                nonexisting.append(element.name)
+            name = element._get_full_name()
+            if not self.workspace_exists(name):
+                nonexisting.append(name)
         if nonexisting:
             raise StreamError("Workspace does not exist", detail="\n".join(nonexisting))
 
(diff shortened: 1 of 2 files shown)


abderrahim:
Thanks for the patch, I'll try to make some time to review it.

Can you please add a tests for this?

kotborealis:
@abderrahim added one test for this exact case

kotborealis:
@abderrahim bump :)

juergbi (approved):

Issue #1822, opened 2023-02-04
Support checkout for failure buildtrees


nanonyme:
Sometimes it's not possible to enter a shell in buildtree. BuildStream 2 should support checking out buildtree so that it can be analyzed manually.

nanonyme:
How I would suggest this to work is that when you try to artifact checkout failed build, any potential cached buildtree would be considered included and checked out.

nanonyme:
Currently instead if you checkout failed build, you will silently get empty checkout

nanonyme:
@juergbi turns out I already had feature request for this. So linking https://github.com/apache/buildstream/commit/83bfd0478bbf30311877f8feaf7715529def9c94 to this so it will not get lost

PR #2129, opened 2026-06-01
cli: add buildtree checkout command

ownik:
This PR adds a "--use-buildtrees" option to "bst artifacts checkout", similar to the existing option in "bst shell".

The goal is to make it possible to check out build artifacts from the sandbox build tree when it is available.

Would be okay with adding this functionality, or are there reasons not to expose it through "bst artifacts checkout"? Or is there another way to get build artifacts from sandbox now?

### DIFF
--- a/src/buildstream/_stream.py
+++ b/src/buildstream/_stream.py
@@ -333,23 +333,8 @@ class Stream:
                     reason="shell-missing-deps",
                 )
 
-        # Check if we require a pull queue attempt, with given artifact state and context
         if usebuildtree:
-            if not element._cached_buildroot():
-                if not element._cached():
-                    message = "Artifact not cached locally or in available remotes"
-                    reason = "missing-buildtree-artifact-not-cached"
-                elif element._buildroot_exists():
-                    message = "Buildtree is not cached locally or in available remotes"
-                    reason = "missing-buildtree-artifact-buildtree-not-cached"
-                else:
-                    message = "Artifact was created without buildtree"
-                    reason = "missing-buildtree-artifact-created-without-buildtree"
-                raise StreamError(message, reason=reason)
-
-            # Raise warning if the element is cached in a failed state
-            if element._cached_failure():
-                self._context.messenger.warn("using a buildtree from a failed build.")
+            self._check_buildtree(element)
 
         # Ensure we have our sources if we are launching a build shell
         if scope == _Scope.BUILD and not usebuildtree:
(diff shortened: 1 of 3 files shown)


ownik:
@BenjaminSchubert @abderrahim @cs-shadow @gtristan @juergbi I've made changes to verify that buildtree exists in the cache, similar to the shell command. Could you please review, please?

abderrahim:
Looks like this is fixing https://github.com/apache/buildstream/issues/1822, please link the issue in your PR. Also please add a test for this behaviour.

I'll try to review it today.

abderrahim:
A few things need to be considered in the design of this feature:
* Someone recently suggested to implement this as a flag to `source checkout` rather than `artifact checkout`. It kinda makes sense that the build tree is related to the source, but it's also stored as part of the artifact. Maybe it makes sense to move it to a separate toplevel command? something like `bst buildtree checkout`?
* This is currently checking out the buildroot rather than what is traditionally considered a "build tree". This means it can work with both build elements and non-build-elements, but it also means we get a whole sandbox with the dependencies, the build tree, and even the artifact (if the build succeeded) which is probably not what people asking for the build tree are looking for.
* if we go with a new toplevel command, we might want to expose options to checkout either the build tree or the build root.

ownik:
@abderrahim Thanks for the feedback! I've reworked the implementation into a new `bst buildtree checkout` command. By default it checks out the buildtree, and a `--buildroot` option can be used to export the entire buildroot instead.

I've also added an integration test covering the new command and link the issue in PR.

ownik:
@abderrahim Any chance to review this PR?

src/buildstream/_stream.py
```diff
@@ -916,6 +901,74 @@ def artifact_delete(self, targets, *, selection=_PipelineSelection.NONE):
...
+            connect_artifact_cache=True,
+            artifact_remotes=artifact_remotes,
+            ignore_project_artifact_remotes=ignore_project_artifact_remotes,
+        )
+
+        # self.targets contains a list of the loaded target objects
+        # if we specify --deps build, Stream._load() will return a list
+        # of build dependency objects, however, we need to prepare a sandbox
+        # with the target (which has had its appropriate dependencies loaded)
+        element: Element = self.targets[0]
```
abderrahim: Does this make sense for the build tree? Either we're checking out just the buildtree, and then dependencies don't make sense, or we check out the whole buildroot and it already contains the build dependencies.

Am I missing something?

src/buildstream/_stream.py
```diff
@@ -916,6 +901,74 @@ def artifact_delete(self, targets, *, selection=_PipelineSelection.NONE):
...
+
+        self._check_location_writable(location, force=force, tar=tar)
+
+        # Check whether the required elements are cached, and then
+        # try to pull them if they are not already cached.
+        #
+        self.query_cache(elements)
+        self._pull_missing_artifacts(elements)
```
abderrahim: I think we only need to pull the artifact for the element in question. No need for dependencies here.

as an aside, I wonder if we should try to override the `pull-buildtrees` configuration in this case: if we're pulling to extract a build tree, we should probably try to pull the buildtree even if we usually don't.

src/buildstream/_stream.py
```diff
@@ -916,6 +901,74 @@ def artifact_delete(self, targets, *, selection=_PipelineSelection.NONE):
...
+
+        self._check_buildtree(element)
+
+        try:
+            artifact = element._get_artifact()
+            virdir = artifact.get_buildroot() if buildroot else artifact.get_buildtree()
```
abderrahim, line 990: Another thing we need to be aware of (and also give a suggestion about to the user) is that not all elements have a build tree: only build elements do.

So it's not enough to check for the build root, we also need to check for the build tree. Then depending on the availability and the user request, we can suggest an alternative to the user.

| buildtree available | buildroot available | user requested | result |
| --- | --- | --- | --- |
| Y | Y | build tree | extract build tree |
| Y | Y | build root | extract build root |
| N | Y | build tree | error + suggest extract build root |
| N | Y | build root | extract build root |
| N | N | build tree | error |
| N | N | build root | error |

tests/integration/buildtree.py
```diff
@@ -0,0 +1,133 @@
...
+#
+# Verify fail cases when checkout buildtree
+#
+@pytest.mark.datafiles(DATA_DIR)
+@pytest.mark.skipif(not HAVE_SANDBOX, reason="Only available with a functioning sandbox")
+def test_buildtree_checkout_fail(cli, datafiles):
+    project = str(datafiles)
+    element_name = "build-shell/buildtree.bst"
+    checkout = os.path.join(cli.directory, "checkout")
+    tar = os.path.join(cli.directory, "source-checkout.tar")
```
abderrahim:
```suggestion
    tar = os.path.join(cli.directory, "buildtree.tar")
```

tests/integration/buildtree.py
```diff
@@ -0,0 +1,133 @@
...
+DATA_DIR = os.path.join(os.path.dirname(os.path.realpath(__file__)), "project")
+
+#
+# Verify fail cases when checkout buildtree
+#
+@pytest.mark.datafiles(DATA_DIR)
+@pytest.mark.skipif(not HAVE_SANDBOX, reason="Only available with a functioning sandbox")
+def test_buildtree_checkout_fail(cli, datafiles):
+    project = str(datafiles)
+    element_name = "build-shell/buildtree.bst"
```
abderrahim, line 41: We should also test the buildtree of a script element (using `build-shell/script.bst`). This is the corner case I mentioned above where we have a build root but not a build tree.

abderrahim (commented):
Thanks for the contributions. See my comments below.

src/buildstream/_stream.py
ownik: Sorry for the delayed response, and thanks for pointing this out.

Yes, you're right. I've changed this to `element: Element = elements[0]`, is it ok?

src/buildstream/_stream.py
ownik: Done 👍

src/buildstream/_stream.py
ownik: Thanks for pointing this out. I've accounted for all of the cases you described and added tests covering them.

Could you please take another look when you have a chance?

tests/integration/buildtree.py
ownik: Fixed

tests/integration/buildtree.py
ownik: i've added a test case which uses `build-shell/script.bst`

PR #2133, opened 2026-06-11
loader: normalize ./-prefixed element names (#2132)

### LINKED ISSUE #2132: `./foo.bst` and `foo.bst` produce different element hashes
Buildstream appears to treat an element path with a leading `./` differently from the same path without it.

Reproducer, using [tests/elements/filter/basic](https://github.com/apache/buildstream/tree/master/tests/elements/filter/basic):

```bash
$ bst show input.bst
   ...
   buildable 110d3798d21e046c9aebbc51a1a3465655f0be1d0f44646b8c0557d724d253cb./input.bst

$ bst show ./input.bst
...
   buildable 55aa30d77816a4056a0b9a2b4f8265d70675aa7afd9de5f71f01d02d76832f22input.bst
```

It looks wrong, because semantically they are the same element.

It gets worse with combinations of other commands. If I build `input.bst`, `./input.bst` is not considered built.
```bash
$ bst build input.bst

$ bst show ./input.bst
...
   buildable 55aa30d77816a4056a0b9a2b4f8265d70675aa7afd9de5f71f01d02d76832f22 ./input.bst

$ bst show input.bst
      cached 110d3798d21e046c9aebbc51a1a3465655f0be1d0f44646b8c0557d724d253cb input.bst
```

This leads to confusing moments, like `bst build input.bst && bst artifact checkout ./input.bst`. It looks the same to the user, but buildstream complains that element is not built.

Found it while messing with junctions:
```
$ bst show --deps none --format '%{name}||%{state}' my/junction.bst
my/junction.bst||junction
./my/junction.bst||no reference
```

BuildStream could not resolve `./my/junction.bst` as junction and complains about missing ref.

kotborealis:
Fixes #2132 by normalizing `./`-prefixed names in loader, so `./a.bst` and `a.bst` are treated exactly the same.

### DIFF
--- a/src/buildstream/_loader/loader.py
+++ b/src/buildstream/_loader/loader.py
@@ -168,6 +168,11 @@ class Loader:
 
         return target_elements
 
+    def _normalize_element_name(self, filename):
+        # Keep equivalent relative spellings on the same loader identity path,
+        # so ref lookup, cache keys and workspace handling stay consistent.
+        return os.path.normpath(filename)
+
     # get_loader():
     #
     # Obtains the appropriate loader for the specified junction
@@ -399,6 +404,8 @@ class Loader:
     #                   or None, if loading of the subproject is disabled.
     #
     def _load_one_file(self, filename, provenance_node, *, load_subprojects=True):
+
+        filename = self._normalize_element_name(filename)
 
         element = None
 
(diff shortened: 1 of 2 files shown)


juergbi (changes requested):
Thanks for the PR. This issue also affects targets such as `foo/./bar.bst`, not just a `./` prefix. I think we can use `os.path.normpath()` to cover all these cases.

juergbi (approved):

PR #2135, opened 2026-06-22
_frontend: Disable status printer in non-interactive mode

### LINKED ISSUE #2111: Debugging plugins
I want to debug plugins in a stupid way, by spraying `breakpoint()` into sources and using `pdb` in the terminal.
However, while building, buildstream keeps spamming status updates, which overwrite debugger's prompt.

Is there any way to stop status while debugger is active? Non-interactive options do not help.

Currently I'm using a small patch to disable status updates if env variable is set.

juergbi:
It was already disabled if the output streams weren't connected to a terminal. However, the CLI option `--no-interactive` accidentally enabled the status printer.

Fixes #2111.

### DIFF
--- a/src/buildstream/_frontend/app.py
+++ b/src/buildstream/_frontend/app.py
@@ -259,15 +259,16 @@ class App:
             self.stream.init()
 
             # Create our status printer, only available in interactive
-            self._status = Status(
-                self.context,
-                self._state,
-                self._content_profile,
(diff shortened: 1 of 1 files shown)


abderrahim:
Undrafting as per https://github.com/apache/buildstream/issues/2111#issuecomment-4780581592

abderrahim (approved):

PR #2136, opened 2026-06-23
fix(docs): update `workspace reset --soft` to avoid smart-quotes

kotborealis:
In docs for incremental builds and buildelement `--soft` turned into `–soft`, which is ignored by CLI parser.
This lead to running `bst workspace reset –-soft --all` (with en-dash) by copypasting it, which turned out to hard-reset of workspaces.

See:
* https://docs.buildstream.build/2.5/developing/workspaces.html#incremental-builds:~:text=bst%20workspace%20reset-,%E2%80%93soft%2C,-like%20so%3A
* https://docs.buildstream.build/2.5/buildstream.buildelement.html#:~:text=bst%20workspace%20reset-,%E2%80%93soft,-is%20called%20on

This patch fixes docs so sphinx would not render it as `en-dash`: moved `--soft` outside of `:ref:`, as I could not get it to work otherwise.

### DIFF
--- a/doc/source/developing/workspaces.rst
+++ b/doc/source/developing/workspaces.rst
@@ -105,7 +105,8 @@ where build scripts automatically detect sources in it's configuration phase,
 so newly added sources you add might be ignored.
 
 In order to force the configuration step to be called again on the next build,
-you can use :ref:`bst workspace reset --soft <invoking_workspace_reset>`, like so:
+you can use :ref:`bst workspace reset <invoking_workspace_reset>` with the
+``--soft`` option, like so:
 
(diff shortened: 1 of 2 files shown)


jjardon:
I think this looks ok to go: @abderrahim @juergbi ok to merge?

abderrahim (approved):

juergbi:
I've found an upstream issue for this: https://github.com/sphinx-doc/sphinx/issues/11492

With enough backslashes, escaping seems to work. I've opened #2155 as alternative to this PR, also covering more instances of this issue.

PR #2137, opened 2026-06-23
fix: make `workspace reset` accept either `--all` or elements list

kotborealis:
When using `bst workspace reset --all a.bst b.bst c.bst`, list of elements is ignored and it resets all workspaces, which could be confusing.

This patch adds a check that either `--all` is specified or list of elementsis provided.

This also fixes the following strange problem:

Running `bst workspace reset --all --soft` (with en-dash in `--soft`) causes buildstream to parse it as: reset with `all` flag and with list of elements `--soft`, as it could not parse it as a flag due to en-dash. This leads to bst hard-resetting workspace when for user it seems like he requested the `--soft`-one.

Now bst would throw error that either `--all` or element list needs to be provided, and would not reset user changes to workspace.

Ref. https://github.com/apache/buildstream/pull/2136

### DIFF
--- a/src/buildstream/_frontend/cli.py
+++ b/src/buildstream/_frontend/cli.py
@@ -1211,14 +1211,21 @@ def workspace_close(app, remove_dir, all_, elements):
     is_flag=True,
     help="Mark workspace to re-execute configuration steps (if any) on next build. Does not alter workspace contents.",
 )
-@click.option("--all", "-a", "all_", is_flag=True, help="Reset all open workspaces")
+@click.option("--all", "-a", "all_", is_flag=True, help="Reset all open workspaces. Cannot be combined with ELEMENTS.")
 @click.argument("elements", nargs=-1, type=click.Path(readable=False))
 @click.pass_obj
 def workspace_reset(app, soft, all_, elements):
-    """Reset a workspace to its original state"""
+    """Reset a workspace to its original state.
+
+    Specify either one or more ELEMENTS, or use ``--all`` to target all
+    workspaces at once. These options are mutually exclusive.
+    """
 
     # Check that the workspaces in question exist
     with app.initialized():
+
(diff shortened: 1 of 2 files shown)


tests/frontend/workspace.py
```diff
@@ -634,6 +634,49 @@ def test_reset_all(cli, tmpdir, datafiles):
     assert not os.path.exists(os.path.join(workspace_beta, "etc", "pony.conf"))
 
 
+@pytest.mark.datafiles(DATA_DIR)
+def test_reset_all_with_elements_error(cli, tmpdir, datafiles):
```
abderrahim: The two tests you added here are essentially the same, please merge them into one and use `@pytest.mark.parametrize` to pass the element name (which would be either the actual element or the bogus --soft option)

src/buildstream/_frontend/cli.py
```diff
@@ -1211,15 +1211,22 @@ def workspace_close(app, remove_dir, all_, elements):
...
 )
-@click.option("--all", "-a", "all_", is_flag=True, help="Reset all open workspaces")
+@click.option("--all", "-a", "all_", is_flag=True, help="Reset all open workspaces. Cannot be combined with ELEMENTS.")
 @click.argument("elements", nargs=-1, type=click.Path(readable=False))
 @click.pass_obj
 def workspace_reset(app, soft, all_, elements):
-    """Reset a workspace to its original state"""
+    """Reset a workspace to its original state.
+
+    Specify either one or more ELEMENTS, or use -a to target all
```
abderrahim: I think I prefer to say `--all` rather than `-a` when documenting things.

kotborealis: Wanted to avoid surprises with smart quotes and en-dash. Fixed.

juergbi (approved):

PR #2147, opened 2026-07-06
Shell API and CLI: Add option for staging additional runtime targets 

nathanwilliams-ct:
This enables users to add additional functionality such as debug tooling in the shell sandbox,
without needing to modify the target element. This is achived through introducing a new option to
shell.

All existing API and UX is maintained, to not break existing scripts. An alternative design was
considered, to have the additional elements as positional arguments similar to the existing element,
but this would need manual parsing to handle the cases where `--` is present and not
present, splitting based on a `.bst` suffix. This UX could be re-visited in future.

Example usage:
  bst shell --with base.bst example.bst -- cat example.txt
Where:
- example.bst is a simple import element with no dependencies that imports a file called example.txt
- base.bst provides a basic alpine sysroot with a standard set of unix tooling (sh, df, cat etc).

Changes:
- Introduces `test_with_other_targets` integration test to the shell test suite.
- Adds `--with` cli option to the shell subcommand and updates it's documentation.
  - option can be used multiple times by caller, providing a list of targets.
- Extends the shell top level calling interface in Buildstream core to accept a list of
other_targets
  - This is where the targets are loaded into elements and checked to make sure they are present
- Extends the shell element implementation to accept a list of other targets
  - This is where the other elements are staged and integrated into the sandbox
  
towards: #422

### DIFF
--- a/src/buildstream/_frontend/cli.py
+++ b/src/buildstream/_frontend/cli.py
@@ -20,6 +20,7 @@ from typing import TYPE_CHECKING
 import shutil
 import click
 from .. import _yaml
+from .._frontend.app import App
 from .._exceptions import BstError, LoadError, AppError, RemoteError
 from .complete import main_bashcomplete, complete_path, CompleteUnhandled
 from ..types import _CacheBuildTrees, _SchedulerErrorAction, _PipelineSelection, _HostMount, _Scope
@@ -723,7 +724,7 @@ def show(app, elements, deps, except_, order, format_):
 @click.argument("command", type=click.STRING, nargs=-1)
 @click.pass_obj
 def shell(
-    app,
+    app: App,
     target,
     command,
     mount,
--- a/src/buildstream/_frontend/cli.py
+++ b/src/buildstream/_frontend/cli.py
@@ -378,8 +378,6 @@ def cli(context, **kwargs):
     user preferences configuration file.
     """
 
-    from .app import App
-
     # Create the App, giving it the main arguments
     context.obj = App.create(dict(kwargs))
     context.call_on_close(context.obj.cleanup)
@@ -687,6 +685,13 @@ def show(app, elements, deps, except_, order, format_):
     metavar="HOSTPATH PATH",
     help="Mount a file or directory into the sandbox",
 )
+@click.option(
+    "--with",
+    "other_targets",
+    type=click.Path(readable=False),
+    multiple=True,
+    help="A additional target to stage into an element's sandbox environment",
+)
 @click.option("--isolate", is_flag=True, help="Create an isolated build sandbox")
 @click.option(
     "--use-buildtree",
@@ -726,6 +731,7 @@ def show(app, elements, deps, except_, order, format_):
 def shell(
     app: App,
     target,
+    other_targets,
     command,
     mount,
     isolate,
@@ -749,13 +755,38 @@ def shell(
     otherwise bst may respond to them instead. e.g.
 
     \b
-        bst shell example.bst -- df -h
+        bst shell base.bst -- df -h
 
     Use the --build option to create a temporary sysroot for
     building the element instead.
 
+    Use the --with option to stage the artifacts of other elements
+    into the temporary sysroot to make them available to run e.g.
+
+    \b
+        bst shell --with base.bst example.bst -- cat example.txt
+
     If no COMMAND is specified, the default is to attempt
     to run an interactive shell.
+
+    # Examples:
+
+    \b
+        # Attempt to run an interactive shell with example.bst
+        bst shell example.bst
+        # Attempt to run an df -h with example.bst
+        bst shell example.bst -- df h
+        # In a workspace directory, attempt to shell into the workspace element
+        bst shell
+        # Attempt to run cat from base.bst to read example.txt from example.bst
+        bst shell --with base.bst example.bst -- cat example.txt
+        # Attempt to run an interactive shell with the sources and all dependencies of example.bst
+        bst shell --build example.bst
+
+    For all examples on this page:
+    - example.bst is a simple import element with no dependencies
+       that imports a file called example.txt
+    - base.bst provides a basic alpine sysroot with a standard set of unix tooling (sh, df, cat etc).
     """
 
     # Buildtree can only be used with build shells
@@ -765,6 +796,7 @@ def shell(
     scope = _Scope.BUILD if build_ else _Scope.RUN
 
     with app.initialized():
+        assert app.stream, "Must have Stream initialised"
         if not target:
             target = app.stream.get_default_target()
             if not target:
@@ -773,19 +805,35 @@ def shell(
         mounts = [_HostMount(path, host_path) for host_path, path in mount]
 
         try:
-            exitcode = app.stream.shell(
-                target,
-                scope,
-                app.shell_prompt,
-                mounts=mounts,
-                isolate=isolate,
-                command=command,
-                usebuildtree=cli_buildtree,
-                artifact_remotes=artifact_remotes,
-                source_remotes=source_remotes,
-                ignore_project_artifact_remotes=ignore_project_artifact_remotes,
-                ignore_project_source_remotes=ignore_project_source_remotes,
-            )
+            if other_targets:
+                exitcode = app.stream.shell_with(
+                    target,
+                    other_targets,
+                    scope,
+                    app.shell_prompt,
+                    mounts=mounts,
+                    isolate=isolate,
+                    command=command,
+                    usebuildtree=cli_buildtree,
+                    artifact_remotes=artifact_remotes,
+                    source_remotes=source_remotes,
+                    ignore_project_artifact_remotes=ignore_project_artifact_remotes,
+                    ignore_project_source_remotes=ignore_project_source_remotes,
+                )
+            else:
+                exitcode = app.stream.shell(
+                    target,
+                    scope,
+                    app.shell_prompt,
+                    mounts=mounts,
+                    isolate=isolate,
+                    command=command,
+                    usebuildtree=cli_buildtree,
+                    artifact_remotes=artifact_remotes,
+                    source_remotes=source_remotes,
+                    ignore_project_artifact_remotes=ignore_project_artifact_remotes,
+                    ignore_project_source_remotes=ignore_project_source_remotes,
+                )
         except BstError as e:
             raise AppError("Error launching shell: {}".format(e), detail=e.detail, reason=e.reason) from e
 
--- a/src/buildstream/_loader/loader.py
+++ b/src/buildstream/_loader/loader.py
@@ -21,7 +21,7 @@ from .._exceptions import LoadError
 from ..exceptions import LoadErrorReason
 from .. import _yaml
 from ..element import Element
-from ..node import Node
+from ..node import Node, MappingNode
 from .._profile import Topics, PROFILER
 from .._includes import Includes
 from .._utils import valid_chars_name
@@ -1014,7 +1014,9 @@ class Loader:
     #            - (str): name of the element
     #            - (Loader): loader for sub-project
     #
-    def _parse_name(self, name, provenance_node, *, load_subprojects=True):
+    def _parse_name(
+        self, name: str, provenance_node: MappingNode, *, load_subprojects: bool = True
+    ) -> tuple[str | None, str, "Loader"]:
         # We allow to split only once since deep junctions names are forbidden.
         # Users who want to refer to elements in sub-sub-projects are required
         # to create junctions on the top level project.
--- a/src/buildstream/_loader/loader.py
+++ b/src/buildstream/_loader/loader.py
@@ -21,7 +21,7 @@ from .._exceptions import LoadError
 from ..exceptions import LoadErrorReason
 from .. import _yaml
 from ..element import Element
-from ..node import Node
+from ..node import Node, MappingNode
 from .._profile import Topics, PROFILER
 from .._includes import Includes
 from .._utils import valid_chars_name
@@ -73,7 +73,7 @@ class Loader:
         self._links = {}  # Dict of link target target paths indexed by link element paths
         self._loaders = {}  # Dict of junction loaders
         self._loader_search_provenances = {}  # Dictionary of provenance nodes of ongoing child loader searches
-
+        self._fullpath_overrides: dict[str, str] = {}  # Dictionary: Original Path, Replacement Path
         self._includes = Includes(self, copy_tree=True)
 
         assert project.name is not None
@@ -255,6 +255,19 @@ class Loader:
     #            Private Methods              #
     ###########################################
 
+    # _set_fullpath_overrides()
+    #
+    # Set an fullpath override for a element-path relative bst file
+    #
+    # This enables runtime modified elements to be pulled from a temporary directory
+    #
+    # Args:
+    #    filename (str): The element-path relative bst file
+    #    fullpath (str): A fullpath to the bst file
+    #
+    def _set_fullpath_overrides(self, filename: str, fullpath: str):
+        self._fullpath_overrides[filename] = fullpath
+
     # _load_file_no_deps():
     #
     # Load a bst file as a LoadElement
@@ -275,8 +288,8 @@ class Loader:
 
         self._assert_element_name(filename, provenance_node)
 
-        # Load the data and process any conditional statements therein
-        fullpath = os.path.join(self._basedir, filename)
+        fullpath = self._fullpath_overrides.get(filename, os.path.join(self._basedir, filename))
+
         try:
             node = _yaml.load(
                 fullpath, shortname=filename, copy_tree=self.load_context.rewritable, project=self.project
@@ -1014,7 +1027,9 @@ class Loader:
     #            - (str): name of the element
     #            - (Loader): loader for sub-project
     #
-    def _parse_name(self, name, provenance_node, *, load_subprojects=True):
+    def _parse_name(
+        self, name: str, provenance_node: MappingNode, *, load_subprojects: bool = True
+    ) -> tuple[str | None, str, "Loader"]:
         # We allow to split only once since deep junctions names are forbidden.
         # Users who want to refer to elements in sub-sub-projects are required
         # to create junctions on the top level project.
--- a/src/buildstream/_stream.py
+++ b/src/buildstream/_stream.py
@@ -16,6 +16,7 @@
 #        Jürg Billeter <juerg.billeter@codethink.co.uk>
 #        Tristan Maat <tristan.maat@codethink.co.uk>
 
+
 import itertools
 import os
 import sys
@@ -27,7 +28,11 @@ import tempfile
 from contextlib import contextmanager, suppress
 from collections import deque
 from typing import List, Tuple, Optional, Iterable, Callable
+from ruamel.yaml import CommentedMap
 
+
+from ._context import Context
+from .node import MappingNode
 from ._artifactelement import verify_artifact_ref, ArtifactElement
 from ._artifactproject import ArtifactProject
 from ._exceptions import StreamError, ImplError, BstError, ArtifactElementError, ArtifactError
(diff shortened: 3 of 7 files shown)


nathanwilliams-ct:
Thanks, whoever triggered the CI... looks like it needs a little work, despite the tests working locally in the tox environment and I missed the linters.

nathanwilliams-ct:
I think this should pass CI now, fixed docs, linters, static type checker, formatter and 3.10 issues

nathanwilliams-ct:
pre-commit hooks would be nice. I keep running the commands locally, but It seems I am doing them in the wrong order.. fixing one thing breaks the others XD

nathanwilliams-ct:
```bash
tox
tox -- --integration
tox -e docs
tox -e lint
tox -e mypy
tox -e format
```

And the tests locally only run on 313 and 314 for me, while CI does also 310 311 312

Is complete list right?

juergbi:
> And the tests locally only run on 313 and 314 for me, while CI does also 310 311 312

tox runs the tests on all Python versions from 3.10 to 3.14 by default, but skipping versions where the interpreter is not installed. It's normally fine to test against just one Python version locally, letting CI take care of the others, unless you're actively debugging a version-specific issue.

> Is complete list right?

CI also runs buildgrid and buildbarn, tests against master of buildstream-plugins-community, and builds wheels. Unless you're working in these specific areas, it shouldn't be necessary to run those locally.

nathanwilliams-ct:
Ah there's a subtle conflict between format and the docs generation.

nathanwilliams-ct:
Static type checking moved out into it's own PR: https://github.com/apache/buildstream/pull/2153

src/buildstream/element.py
```diff
@@ -2076,6 +2091,17 @@ def _shell(self, scope=None, *, mounts=None, isolate=False, prompt=None, command
...
+            with self.timed_activity("Staging other_targets", silent_nested=True), self.__collect_overlaps(sandbox):
+                self.stage_dependency_artifacts(sandbox, other_elements)
+
+            if other_elements:
+                # Stage artifacts from other_elements into the sandbox.
+                for element in other_elements:
+                    # Stage deps in the sandbox root
+                    with element.timed_activity("Integrating sandbox"), sandbox.batch():
+                        for dep in element._dependencies(_Scope.RUN):
+                            dep.integrate(sandbox)
```
juergbi, line 2103: The separate staging and integration into the same sandbox could be problematic. The main target element and the other targets may share some dependencies, in which case an element gets staged twice. This may cause confusing overlap warnings or errors (and in some constellations maybe also some real overlap conflicts).

Separate integration means that integration commands of dependencies of the main target can't cover integration with other targets. And integration commands of shared dependencies will be executed twice (or even more with multiple `--with`). Additional sandbox batch execution for other target integration may also not be the most efficient approach, but performance is not my main concern here.

I'm not saying that this approach is definitely unacceptable, but at the very least it needs to have documented and tested behavior for mentioned aspects such as overlaps and integration commands. Also build shells might not be tested at all right now, if I haven't missed anything.

Regarding overlaps, a possible alternative would be to stage the other elements into a separate sandbox / virtual directory (with the usual overlap processing) and then merge it into the shell where the overlap handling may be different (e.g., the `--with` tree allowed to always replace files). This was also suggested in https://mail.gnome.org/archives/buildstream-list/2019-February/msg00001.html. Integration commands will still be problematic but maybe some limitations there are acceptable (but should also be clarified). I would definitely at least use a single integration sandbox for all 'other' elements and don't duplicate integration within that part.

### Virtual stack element

If it was only for runtime shells, I think the behavior should rather be equivalent to creating a stack element that has the main target and all other targets as dependencies, which would likely not even require any changes in `element.py`. One caveat is that runtime shells use the environment variables from the target element, so that would break with the (virtual) stack element approach.

However, build shells make things more complicated as there the main target has essentially full control over the sandbox.

### Inject other targets as dependencies

One possible alternative that comes to mind is that we may be able to inject the `--with` elements as dependencies of the main target (runtime dependency for runtime shells and build dependency for build shells). There could be element plugins where this is problematic for build shells but normal build elements should be fine and build shells anyway can't work with all element plugins.

It's possible that I'm missing something why this would be a bad idea, but it might be worth exploring if nobody can think of a clear blocker right away.

One issue I can think of is that it might not work with buildtrees where we get the full sandbox root from CAS and don't stage anything. It may be possible to support an alternative buildtree support (only used with `--with`) where we first construct a sandbox like for a normal build shell and then only replace the source/build directory with the corresponding directory from the buildtree. If we want to go down this route, this should likely wait for a follow-up PR.

nathanwilliams-ct: I will look to explore these alternative routes, thanks for the feedback.

nathanwilliams-ct:
```
$ tox -e venv -- bst --directory tests/integration/project shell manual/import-file.bst                            
[--:--:--][        ][    main:core activity                 ] START   Loading elements
[00:00:00][        ][    main:core activity                 ] SUCCESS Loading elements
[--:--:--][        ][    main:core activity                 ] START   Resolving elements
[00:00:00][        ][    main:core activity                 ] SUCCESS Resolving elements
[--:--:--][        ][    main:core activity                 ] START   Initializing remote caches
[00:00:00][        ][    main:core activity                 ] SUCCESS Initializing remote caches
[--:--:--][        ][    main:core activity                 ] START   Query cache
[00:00:00][        ][    main:core activity                 ] SUCCESS Query cache
[--:--:--][dc2422ef][    main:manual/import-file.bst        ] START   Staging dependencies
[00:00:00][dc2422ef][    main:manual/import-file.bst        ] SUCCESS Staging dependencies
[--:--:--][dc2422ef][    main:manual/import-file.bst        ] START   Integrating sandbox
[00:00:00][dc2422ef][    main:manual/import-file.bst        ] SUCCESS Integrating sandbox
[--:--:--][dc2422ef][    main:manual/import-file.bst        ] STATUS  Running command

    sh -i

Error launching shell: Staged artifacts do not provide command 'sh'

$ tox -e venv -- bst --directory tests/integration/project shell manual/import-file.bst --with base/base-alpine.bst       
[--:--:--][        ][    main:core activity                 ] START   Loading elements
[00:00:00][        ][    main:core activity                 ] SUCCESS Loading elements
[--:--:--][        ][    main:core activity                 ] START   Resolving elements
[00:00:00][        ][    main:core activity                 ] SUCCESS Resolving elements
[--:--:--][        ][    main:core activity                 ] START   Initializing remote caches
[00:00:00][        ][    main:core activity                 ] SUCCESS Initializing remote caches
[--:--:--][        ][    main:core activity                 ] START   Query cache
[00:00:00][        ][    main:core activity                 ] SUCCESS Query cache
[--:--:--][e8d846b7][   build:manual/import-file.bst_tempshwrznrg.bst] START   test/manual-import-file.bst_tempshwrznrg/e8d846b7-build.20260818-111411.log
[--:--:--][e8d846b7][   build:manual/import-file.bst_tempshwrznrg.bst] START   Staging sources
[00:00:00][e8d846b7][   build:manual/import-file.bst_tempshwrznrg.bst] SUCCESS Staging sources
[--:--:--][e8d846b7][   build:manual/import-file.bst_tempshwrznrg.bst] START   Caching artifact
[00:00:00][e8d846b7][   build:manual/import-file.bst_tempshwrznrg.bst] SUCCESS Caching artifact
[00:00:00][e8d846b7][   build:manual/import-file.bst_tempshwrznrg.bst] SUCCESS test/manual-import-file.bst_tempshwrznrg/e8d846b7-build.20260818-111411.log
[--:--:--][        ][    main:core activity                 ] START   Loading elements
[00:00:00][        ][    main:core activity                 ] SUCCESS Loading elements
[--:--:--][        ][    main:core activity                 ] START   Resolving elements
[00:00:00][        ][    main:core activity                 ] SUCCESS Resolving elements
[--:--:--][        ][    main:core activity                 ] START   Initializing remote caches
[00:00:00][        ][    main:core activity                 ] SUCCESS Initializing remote caches
[--:--:--][        ][    main:core activity                 ] START   Query cache
[00:00:00][        ][    main:core activity                 ] SUCCESS Query cache
[--:--:--][e8d846b7][    main:manual/import-file.bst_tempshwrznrg.bst] START   Staging dependencies
[00:00:00][e8d846b7][    main:manual/import-file.bst_tempshwrznrg.bst] SUCCESS Staging dependencies
[--:--:--][e8d846b7][    main:manual/import-file.bst_tempshwrznrg.bst] START   Integrating sandbox
[00:00:00][e8d846b7][    main:manual/import-file.bst_tempshwrznrg.bst] SUCCESS Integrating sandbox
[--:--:--][e8d846b7][    main:manual/import-file.bst_tempshwrznrg.bst] STATUS  Running command

    sh -i
[e8d846b7@manual/import-file.bst_tempshwrznrg.bst:/]$ cat test.txt
This is a test
[e8d846b7@manual/import-file.bst_tempshwrznrg.bst:/]$ exit
```

hmm, I went down the route of creating a temporary element.

Due to how the element loading works, it makes it almost impossible to inject additional dependencies in at runtime. Temporary stack element would have made buildtree ones unhelpful. Solving the overlap problem and dealing with overlapping sandboxes and integration commands was too complicated and messy.

The downside of using a temporary element is the shell command needs to attempt to run a build on the temporary element, before it can shell, to cache it's build result.

I'll tidy up my prototype with some tests and push it a bit later on.

juergbi: > Temporary stack element would have made buildtree ones unhelpful.

> The downside of using a temporary element is the shell command needs to attempt to run a build on the temporary element, before it can shell, to cache it's build result.

Not being able to use a cached build seems like a major downside to me. I was originally thinking of injecting it in the in-memory representation, not creating a temporary file. But maybe that's not feasible.

A possible mitigation with the temporary file approach could be using different approaches for build and runtime shells. A temporary stack element for runtime shells and what your branch is doing now for build shells. Or have you already considered this as not tenable for some reason?

juergbi: While injecting deps in the in-memory representation should theoretically be possible, it would probably be too invasive just for this feature.

A possible tweak to the temp file approach could be to add some kind of substitution / path override dict to the `Loader`. This could be set in `Stream.shell_with()`. `_load_file_no_deps()` would then use the path override, instead of constructing the regular `fullpath`. However, it would still use the regular `filename` as `shortname`.

I haven't prototyped this, but the possible advantages are:
* No cache key difference for runtime shells, being able to use a previously cached build
* Logging wouldn't expose temp filenames
* Temp file could be placed in a temporary directory instead of polluting the elements directory with a temporary file

Any thoughts?

nathanwilliams-ct: Temporary stack element for runtime shell would be possible I did try it out, but because build shells don't work with that approach and with the aim to keep maintenance cost down: I didn't want two separate implementations if we can help it.

I tried a few ways to inject dependencies in-memory, but part the problem I found was a lot of dependency loading and resolving work is done at the early stages immediately after loading the yaml and parsing, where it's basically impossible to inject the dependencies in a sensible way. Doing it at a later stage e.g. in the Stream.shell or the Element.execute_shell method and or trying to add a inject_extra_deps method to the Element class, I couldn't get it to work. Partly due to the complexity of the shell construction, especially where it involves the cached buildtrees being used which means the whole shell construction process is skipped. The lack of type hints to actually understand the control flow also doesn’t help(https://github.com/apache/buildstream/pull/2167).

I will have a play with adjusting the tempfile approach to hide it a bit better like you suggest.

nathanwilliams-ct: It might be nice to have a 'reload' method on Element that can cleanly reload the element if there are changes at runtime, but it would involve some heavy refactoring that I currently don't have the context to do. Element is heavily implemented around the idea of everything is immutable.

We could almost do with a replacing the whole Element class with an explicit state machine to represent elements at different stages of it's lifecycle. There are so many Optional fields and 'state' booleans, it's difficult to untangle and work out which state the element is currently in, and what order to call things in etc.

image

```plantuml
[*] --> LoadElement: Load from YAML
LoadElement --> WeakElement: Calculate weak cache key
WeakElement --> StrongElement: Resolve dependencies and calculate strong key
StrongElement --> CachedElement: Element has cached artifacts
StrongElement --> PreparedElement: Prepare sources and dependencies for a build
PreparedElement --> CachedElement: Build the element
PreparedElement --> FailedBuildElement: cache buildtree
```

nathanwilliams-ct: I've made the changes to how the tempfile is handled to make it more transparent. Original element name now appears in the logs.

src/buildstream/_stream.py
```diff
@@ -235,6 +240,65 @@ def query_cache(self, elements, *, sources_of_cached_elements=False, only_source
...
+            target_node["build-depends"] = r_depends
+        else:
+            raise StreamError(
+                "Only BUILD and RUN scopes are supported",
+                detail="Use the --build and --use-buildtree options to shell into a build tree",
+                reason="only-build-run-supported",
+            )
+
+        with tempfile.NamedTemporaryFile(
+            delete_on_close=False, prefix=f"{target_name.replace("/","_")}_temp", suffix=".bst"
```
nathanwilliams-ct: This works for whichever version of python I am using *sigh*

```suggestion
            delete_on_close=False, prefix=f"{target_name.replace('/','_')}_temp", suffix=".bst"
```

I am not very happy with the replace.

tests/integration/shell.py
```diff
@@ -46,18 +49,35 @@
 #    mount (tuple): A (host, target) tuple for the `--mount` option
 #    element (str): The element to build and run a shell with
 #    isolate (bool): Whether to pass --isolate to `bst shell`
-#
-def execute_shell(cli, project, command, *, config=None, mount=None, element="base.bst", isolate=False):
+#    other_elements (list(str)): Other elements to stage in the sandbox
```
nathanwilliams-ct, line 52:
```suggestion
#    other_elements (list(str)): Other elements to stage in the sandbox
#
```

src/buildstream/_loader/loader.py
```diff
@@ -255,6 +255,18 @@ def foreach_parent(parent):
...
 
+    # _set_fullpath_overrides()
+    #
+    # Set an fullpath override for a element-path relative bst file
+    #
+    # This enables runtime modified elements to be pulled from a temporary directory
+    #
+    # Args:
+    #    filename (str): The element-path relative bst file
+    #    fullpath (str): A fullpath to the bst file
```
nathanwilliams-ct, line 266:
```suggestion
    #    fullpath (str): A fullpath to the bst file
    #
```

src/buildstream/_stream.py
```diff
@@ -235,6 +240,65 @@ def query_cache(self, elements, *, sources_of_cached_elements=False, only_source
...
+
+        with tempfile.NamedTemporaryFile(
+            delete_on_close=False, prefix=f"{target_name.replace('/','_')}_temp", suffix=".bst"
+        ) as temp_target_file:
+            _yaml.roundtrip_dump(target_node, temp_target_file)
+            temp_target_file.close()  # delete_on_close is false so this doesn't remove the file, but delete is True(default) so we delete the file when we leave the context manager.
+
+            target_loader._set_fullpath_overrides(target_name, temp_target_file.name)
+
+            self.build([target])
```
juergbi, line 299: This should no longer be necessary with the fullpath override, right?

src/buildstream/_loader/loader.py
```diff
@@ -255,6 +255,19 @@ def foreach_parent(parent):
     #            Private Methods              #
     ###########################################
 
+    # _set_fullpath_overrides()
```
juergbi: This shouldn't be a private method as it's used by `Stream`. `Loader` is not part of the public (plugin) API, so 'public' methods in `Loader` will still only be part of the internal API.

Also, `_set_fullpath_overrides()` sounds to me like it sets the whole dict as it's plural. I might call it `add_fullpath_override()`.

Also see a related comment: https://github.com/apache/buildstream/pull/2147#discussion_r3989808035

src/buildstream/_frontend/cli.py
```diff
@@ -748,13 +755,38 @@ def shell(
...
+        bst shell
+        # Attempt to run cat from base.bst to read example.txt from example.bst
+        bst shell --with base.bst example.bst -- cat example.txt
+        # Attempt to run an interactive shell with the sources and all dependencies of example.bst
+        bst shell --build example.bst
+
+    For all examples on this page:
+    - example.bst is a simple import element with no dependencies
+       that imports a file called example.txt
+    - base.bst provides a basic alpine sysroot with a standard set of unix tooling (sh, df, cat etc).
```
juergbi, line 789: build_docs CI fails with
```
/home/testuser/buildstream/doc/source/using_commands.rst:42: WARNING: Block quote ends without a blank line; unexpected unindent. [docutils]
```

juergbi:
Nit: We prefer limiting normal text lines in git commit messages to 72 characters, matching usual git conventions.

juergbi:
The test seems to be failing on Python 3.10 (Ubuntu 22.04) and Python 3.11 (Debian 12).

src/buildstream/_stream.py
```diff
@@ -235,6 +240,65 @@ def query_cache(self, elements, *, sources_of_cached_elements=False, only_source
...
+    #    (int): The exit code of the launched shell
+    #
+    def shell_with(self, target: str, other_targets: Iterable[str], scope: _Scope, *args, **kwargs):
+
+        assert self._project, "Must have a project"
+        assert self._project.loader, "Project must have loader"
+
+        _, target_name, target_loader = self._project.loader._parse_name(target, MappingNode.from_dict({}))
+
+        target_path = os.path.join(target_loader._basedir, target_name)
```
juergbi, line 267: We shouldn't be using the private members `_parse_name()` and `_basedir`. They should either be public/internal, or we need some new public/internal Loader API as alternative.

A possible option could be to add a `Loader` context manager method that takes care of `_parse_name()`, `os.path.join()`, ` _yaml.roundtrip_load()`, creating the temporary file, `_yaml.roundtrip_dump()` and `_set_fullpath_overrides()`. Leaving only the actual YAML manipulation logic in this method.

src/buildstream/_stream.py
nathanwilliams-ct: Changing the dependencies messes with the cache key so a build is necessary still.

juergbi: Direct runtime dependencies shouldn't affect the cache key. And for build shells (where we add build dependencies), we shouldn't invoke a build before entering the shell anyway. Or am I misremembering or missing something? Have you confirmed a cache key issue with the latest version of your branch (with the fullpath override)?

nathanwilliams-ct: hmmm, No, I will do some further testing.

src/buildstream/_frontend/cli.py
nathanwilliams-ct: hmm does CI do something different to calling tox locally? doing something like WARNINGS_FATAL=TRUE ? I don't see this error when running it locally. 


It's also a very unhelpful error message, I haven't touched `using_commands.rst`, it should point at `cli.py`....

src/buildstream/_loader/loader.py
```diff
@@ -251,10 +255,61 @@ def foreach_parent(parent):
...
+
+        _, target_name, target_loader = self._parse_name(target, MappingNode.from_dict({}))
+
+        target_path = os.path.join(target_loader._basedir, target_name)
+        target_node: CommentedMap = _yaml.roundtrip_load(target_path)
+
+        modify_element_function(target_node)
+
+        # FIXME When 3.12 hits EOL, replace this with tempfile.NamedTemporaryFile itself.
+        with _legacy_named_temporary_file_delete_on_close(
```
juergbi, line 280: We generally use `utils._tempnamedfile()`, which adds cleanup on SIGTERM. And we just `.flush()` the temp file without closing it, which works also without `delete_on_close=False` support.

While moving the utility function to `delete_on_close=False` sounds reasonable once we require Python 3.12, I don't think it makes sense to add a helper to the `Loader` here. Let's be consistent and use the same approach as elsewhere in the codebase.

PR #2153, opened 2026-07-17
Static Type Checking: Enums

nathanwilliams-ct:
Buildstream implements a custom `FastEnum`[1].
mypy only supports `enum.Enum` (and it's official variations) when it comes to doing static type checks on enums [2].
We can't subclass Enum in FastEnum to make mypy happy, because Enum doesn't allow subclassing [3].
So we end up with a bunch of `str`, `int` etc around the codebase,
instead of the appropriate Enum classes like `OverlapAction` or `_Scope` which are usually recorded separately in the doc strings.
This means we don't get proper static type checking on our Enums :( .
With stub files [4][5] we can lie to mypy about what type the Enum classes are[6],
so the type hints around the codebase are now correct be corrected.
Now we can have proper static type checking on the custom Enums, Yay!

So

```
class OverlapAction(FastEnum):
    ERROR: str
    WARNING: str
    IGNORE: str
```

gets stubbed as

```
from enum import Enum
class OverlapAction(Enum):
    ERROR: str
    WARNING: str
    IGNORE: str
```

[1] src/buildstream/types.py#L32
[2] https://mypy.readthedocs.io/en/stable/literal_types.html#enums
[3] https://docs.python.org/3/howto/enum.html#restricted-enum-subclassing
[4] https://typing.python.org/en/latest/guides/writing_stubs.html
[5] https://mypy.readthedocs.io/en/stable/stubgen.html
[6] https://github.com/python/mypy/issues/3217


juergbi (approved):
LGTM. Ideally, we'd have a small script to automate the generation of `type.pyi` and then we could run that in CI to verify that it's up-to-date. However, I think regeneration is reasonably simple as is and there shouldn't be a functional risk with an outdated `type.pyi`, so I think it's fine to merge this as is.

src/buildstream/types.py
```diff
@@ -61,7 +64,7 @@ def __new__(cls, value):
         try:
             return cls._value_to_entry[value]
         except KeyError:
-            if type(value) is cls:  # pylint: disable=unidiomatic-typecheck
+            if isinstance(value, cls):  # pylint: disable=unidiomatic-typecheck
```
wsalmonct, line 67: AFAICT your are now doing this right.

https://pylint.pycqa.org/en/latest/user_guide/messages/convention/unidiomatic-typecheck.html

```suggestion
            if isinstance(value, cls):
```

So you dont need to disable the lint.

src/buildstream/_pipeline.py
```diff
@@ -44,7 +44,7 @@
 # Yields:
 #    Elements in the scope of the specified target elements
 #
-def dependencies(targets: List[Element], scope: int, *, recurse: bool = True) -> Iterator[Element]:
+def dependencies(targets: List[Element], scope: _Scope, *, recurse: bool = True) -> Iterator[Element]:
```
wsalmonct, line 47: It strikes me as odd to have private API (enum) in a public API (function), technically this is in a private module, eg _pipeline but the class might be placed into public API else were, eg at a higher level module.

juergbi: Placing `_pipeline` into public API would violate our public API rules. I.e., this is an internal function and `_Scope` is an internal type. It may look odd but it follows our rules and shouldn't be exposed anywhere in the public API, so I don't see an issue.

Marking this as resolved but if you see a real issue or have a suggestion for improvement, feel free to reopen it.

PR #2154, opened 2026-07-17
Element: Remove unused & deprecated node_subst methods

nathanwilliams-ct:
These deprecation notices were introduced into the codebase 6 years ago (April 16th, 2020 5:56 PM) e5ed02da0b99c16481eac58963f4ebbdf0481590

There have been many Buildstream versions since then, a search in the Buildstream codebase shows they are unused, so it's about time they are removed.

### DIFF
--- a/src/buildstream/_elementproxy.py
+++ b/src/buildstream/_elementproxy.py
@@ -76,12 +76,6 @@ class ElementProxy(PluginProxy):
         #
         return cast("Element", self._plugin)._search(_Scope.RUN, name)
 
-    def node_subst_vars(self, node: "ScalarNode") -> str:
-        return cast("Element", self._plugin).node_subst_vars(node)
-
-    def node_subst_sequence_vars(self, node: "SequenceNode[ScalarNode]") -> List[str]:
-        return cast("Element", self._plugin).node_subst_sequence_vars(node)
-
     def compute_manifest(
         self, *, include: Optional[List[str]] = None, exclude: Optional[List[str]] = None, orphans: bool = True
     ) -> str:
--- a/src/buildstream/element.py
+++ b/src/buildstream/element.py
@@ -505,55 +505,6 @@ class Element(Plugin):
 
         return None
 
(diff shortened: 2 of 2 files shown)


juergbi:
Unfortunately, this is part of the public plugin API where we guarantee stability across stable releases. We failed to remove it before 2.0. And some plugins in buildstream-plugins-community are even still using it, should probably fix that.

IdrisTheDragon:
So this PR is a candidate for 3.0 assuming we sort the buildstream community plugins for it. Do you know if there's an issue to track this?

juergbi:
I don't think we have an issue in https://gitlab.com/BuildStream/buildstream-plugins-community/ for this, probably makes sense to open one.

We also don't have a milestone or label for BuildStream 3.0, not sure whether we should create one without more specific plans.

IdrisTheDragon:
Might be worth creating a label at minimum, so it's easy to find these sorts of issues/ PRs for if/when a 3.0 is planned.

nanonyme:
Please do port the community plugins. It's a good concrete example why the API is not needed. Presuming it's not.

nanonyme:
Necessary API could obviously be undeprecated and porting plugins is good case study.

jjardon:
> I don't think we have an issue in https://gitlab.com/BuildStream/buildstream-plugins-community/ for this, probably makes sense to open one.

I have opened https://gitlab.com/BuildStream/buildstream-plugins-community/-/work_items/94

nathanwilliams-ct:
https://gitlab.com/BuildStream/buildstream-plugins-community/-/merge_requests/454

nathanwilliams-ct:
> https://gitlab.com/BuildStream/buildstream-plugins-community/-/merge_requests/454

This has now been merged into community plugins. So release 2.4 of community plugins won't use this API any more.

Unless there are any other users climb out of the woodwork, this PR can be included in a theoretical 3.0 release of Buildstream.

PR #2160, opened 2026-08-03
_cas: avoid blocking calls to long grpc methods

### LINKED ISSUE #2157: Can't stop artifact pulls sometimes
After trying to stop a build with ctrl-c and choosing terminate, buildstream will fail to terminate any running pulls forever. I get a bunch of logs with `STATUS  Pull terminating` but nothing happens. Nothing more is logged with default logging options. After this I have to kill the bst process.

To my memory, this should always happen if #2156 has also occurred in the same bst invocation. Similar to that one, the machine can be turned offline and the pulls never fail, timeout or stop. Also similarly to that issue, reproduction steps are unclear other than these 2 issues almost always happening together. Although, I believe I have seen this issue happen on its own occasionally as well.

abderrahim:
Long-running synchronous grpc calls block the thread where they are
called, which can cause the scheduler to be unable to terminate the
job.

This commit replaces them with usage of the "asynchronous" future
API. The result() method of a grpc future will not block the thread
while it's waiting, and thus allows terminating the job without
waiting for the blocking call to return. This covers the LocalCAS
methods {Fetch,Upload}{Tree,MissingBlobs}.

As an added benefit, we try to cancel the grpc calls when
terminated. This depends on the server side (buildbox-casd) promptly
cancelling the remote calls, which it doesn't consistently do at the
moment. But that can be fixed independently.

Fixes https://github.com/apache/buildstream/issues/2157

### DIFF
--- a/src/buildstream/_cas/casremote.py
+++ b/src/buildstream/_cas/casremote.py
@@ -92,7 +92,13 @@ class _CASBatchRead:
         local_cas = self._remote.casd.get_local_cas()
 
         for request in self._requests:
-            batch_response = local_cas.FetchMissingBlobs(request)
+            batch_response_future = local_cas.FetchMissingBlobs.future(request)
+
+            try:
+                batch_response = batch_response_future.result()
+            except:
+                batch_response_future.cancel()
+                raise
 
             for response in batch_response.responses:
                 if response.status.code == code_pb2.NOT_FOUND:
@@ -146,7 +152,13 @@ class _CASBatchUpdate:
         local_cas = self._remote.casd.get_local_cas()
 
         for request in self._requests:
-            batch_response = local_cas.UploadMissingBlobs(request)
+            batch_response_future = local_cas.UploadMissingBlobs.future(request)
+
+            try:
+                batch_response = batch_response_future.result()
+            except:
+                batch_response_future.cancel()
+                raise
 
             for response in batch_response.responses:
                 if response.status.code != code_pb2.OK:
--- a/src/buildstream/_cas/cascache.py
+++ b/src/buildstream/_cas/cascache.py
@@ -127,7 +127,13 @@ class CASCache:
         request.fetch_file_blobs = not self._remote_cache
 
         try:
-            local_cas.FetchTree(request)
+            response_future = local_cas.FetchTree.future(request)
+            try:
+                response_future.result()
+            except:
+                response_future.cancel()
+                raise
+
             if not self._remote_cache:
                 return True
         except grpc.RpcError as e:
@@ -141,7 +147,13 @@ class CASCache:
         request = local_cas_pb2.UploadTreeRequest()
         request.root_digest.CopyFrom(digest)
         try:
-            local_cas.UploadTree(request)
+            response_future = local_cas.UploadTree.future(request)
+            try:
+                response_future.result()
+            except:
+                response_future.cancel()
+                raise
+
             return True
         except grpc.RpcError as e:
             if e.code() == grpc.StatusCode.NOT_FOUND:
@@ -231,7 +243,12 @@ class CASCache:
             request.root_digest.CopyFrom(tree)
             request.fetch_file_blobs = True
 
-            local_cas.FetchTree(request)
+            response_future = local_cas.FetchTree.future(request)
+            try:
+                response_future.result()
+            except:
+                response_future.cancel()
+                raise
 
     # fetch_directory():
     #
@@ -252,7 +269,13 @@ class CASCache:
         request.fetch_file_blobs = False
 
         try:
-            local_cas.FetchTree(request)
+            response_future = local_cas.FetchTree.future(request)
+            try:
+                response_future.result()
+            except:
+                response_future.cancel()
+                raise
+
         except grpc.RpcError as e:
             if e.code() == grpc.StatusCode.NOT_FOUND:
                 raise BlobNotFound(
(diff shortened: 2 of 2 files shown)


src/buildstream/_cas/casremote.py
```diff
@@ -92,7 +94,16 @@ def send(self, *, missing_blobs=None):
...
 
         for request in self._requests:
-            batch_response = local_cas.FetchMissingBlobs(request)
+            batch_response_future = local_cas.FetchMissingBlobs.future(request)
+
+            try:
+                while batch_response_future.running():
+                    time.sleep(1)
+            except:
+                batch_response_future.cancel()
```
juergbi, line 100: As I understand it, this will typically not accomplish anything as cancellation will only be possible before the function has started execution. It might be as good as it gets without switching to gRPC AsyncIO, but I think we should at least make this clear in a code comment.

src/buildstream/_cas/casremote.py
```diff
@@ -92,7 +94,16 @@ def send(self, *, missing_blobs=None):
         local_cas = self._remote.casd.get_local_cas()
 
         for request in self._requests:
-            batch_response = local_cas.FetchMissingBlobs(request)
+            batch_response_future = local_cas.FetchMissingBlobs.future(request)
+
+            try:
+                while batch_response_future.running():
+                    time.sleep(1)
```
juergbi: Can't we use `batch_response_future.result(timeout=1)` to avoid an extra delay of up to 1 second once the call is done?

src/buildstream/_cas/casremote.py
abderrahim: I am actually not sure how this works in grpc. Does the grpc asyncio API do things differently?

I took this from the example linked from https://grpc.io/docs/guides/cancellation/. It noticed that it doesn't actually cancel the call from the buildbox-casd side, but I assumed that it would need some work from the buildbox-casd side.

Anyway, this is something I did some time ago and wasn't very happy about it. I only pushed it because the discussion came up in #2157

src/buildstream/_cas/casremote.py
abderrahim: I'll take a look.

src/buildstream/_cas/casremote.py
juergbi: Ah, this is actually a `grpc.Future`, not a generic Python stdlib `concurrent.futures.Future`. That's better than what I first thought and should actually perform proper gRPC cancellation.

Yes, the server (buildbox-casd) side should react to this as well, but that's an improvement that can be done separately and it shouldn't be needed to fix the #2157 issue (although it may still cause some delay on shutdown). buildbox-casd already handles cancellation for local execution requests and as part of the retry mechanism. Further cancellation points can (and should) be added but may need some work as there may be a blocking gRPC client call in some nested function call.

src/buildstream/_cas/casremote.py
abderrahim: pushed a change to use this. I can't say I find it more elegant. Feel free to suggest improvements.

(I'm wondering whether it makes sense to have this loop as a utility function)

juergbi: The updated version may not be more elegant but it should be slightly more efficient.

I think a helper function would make sense. If we don't need it beyond these two gRPC calls, I wouldn't consider it essential but if it's straight forward, why not?

juergbi: Looking at the gRPC Python code, we might not even need a timeout. The internal wait function seems to make sure that the thread is not blocked for longer than 0.1s. See https://github.com/grpc/grpc/blob/f167d6a879e5b3223c99170663b2cc3e791ac6ad/src/python/grpcio/grpc/_common.py#L119

I haven't verified this but we may be able to call `.result()` without a loop or timeout.

abderrahim: We could also do the same to the FetchTree calls in cascache.py, but these ones looked like the more urgent.

abderrahim: > I haven't verified this but we may be able to call `.result()` without a loop or timeout.

Seems to be the case indeed. I like this much better :-)

juergbi: > We could also do the same to the FetchTree calls in cascache.py

Yes, I think we should apply this change also to `FetchTree` in `fetch_directory()` and `FindMissingBlobs` in `missing_blobs()`. This should cover all gRPC calls involving ordinary remotes (i.e., everything except `storage-service`). To also cover `storage-service`, we might have to apply it to practically all gRPC calls.

abderrahim: Updated to cover all instances of `{Fetch,Upload}{Tree,MissingBlobs}`

juergbi: The `FindMissingBlobs` call in `CASCache.missing_blobs()` should be covered as well as that's also used when talking to a remote (and may involve uploading blobs from the local cache to the remote, so can take a while).

The asset Fetch/Push calls are not covered either but they should finish quickly, so are less likely to be affected by this issue.

abderrahim: I didn't know `FindMissingBlobs` could upload blobs to the remote, that's why I skipped it along with the remote asset calls (which should be indeed pretty fast). I changed `FindMissingBlobs`.

juergbi (approved):
LGTM

PR #2161, opened 2026-08-05
examples: use ftp.gnu.org instead of ftpmirror.gnu.org

abderrahim:
The latter has been unreliable recently

### DIFF
--- a/doc/examples/autotools/project.conf
+++ b/doc/examples/autotools/project.conf
@@ -10,10 +10,17 @@ element-path: elements
 # Define some aliases for the tarballs we download
 aliases:
   alpine: https://bst-integration-test-images.ams3.cdn.digitaloceanspaces.com/
-  gnu: http://ftpmirror.gnu.org/gnu/automake/
+  gnu: https://ftp.gnu.org/gnu/automake/
 
 plugins:
 - origin: pip
   package-name: buildstream-plugins
   elements:
   - autotools
+
+# Define mirrors
+mirrors:
+- name: gnu_mirror
+  aliases:
+    gnu:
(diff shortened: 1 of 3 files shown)


doc/examples/autotools/project.conf
```diff
@@ -10,7 +10,7 @@ element-path: elements
 # Define some aliases for the tarballs we download
 aliases:
   alpine: https://bst-integration-test-images.ams3.cdn.digitaloceanspaces.com/
-  gnu: http://ftpmirror.gnu.org/gnu/automake/
+  gnu: http://ftp.gnu.org/gnu/automake/
```
juergbi: If we change this, I think we should configure `ftpmirror.gnu.org` as mirror in `project.conf` to still prefer `ftpmirror.gnu.org` but allow fallback. (In the past `ftp.gnu.org` was unreliable and I think GNU wants people to use `ftpmirror.gnu.org`).

And we should probably switch to https.

I'm ok with deferring the mirror addition if that's not trivial, though.

(We could also consider mirroring it on our infrastructure).

abderrahim: Added a mirror and moved to https for all the links.

I'm not sure about "our infrastructure". What do we have? (the DO bucket used for alpine image?) and who has access to it?

juergbi (approved):

doc/examples/autotools/project.conf
juergbi: I don't have a clear proposal. The actual example project (without automake itself) would be tiny if packaged separately, so, it might not be too difficult to find a place for it.

However, if the ftpmirror + ftp combination is reliable, it may be ok to leave it like this.

PR #2163, opened 2026-08-06
Move static metadata from setup.py/setup.cfg into pyproject.toml

jjardon:
Move the static project metadata that can be expressed with PEP 621 into the [project] table of pyproject.toml and remove the corresponding duplicated arguments from setup() and the python_requires line from setup.cfg.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,6 +23,40 @@ requires = [
 ]
 build-backend = "setuptools.build_meta"
 
+[project]
+name = "BuildStream"
+dynamic = ["version", "dependencies"]
+description = "A framework for modelling build pipelines in YAML"
+readme = { file = "README.rst", content-type = "text/x-rst" }
+license = "Apache-2.0"
+requires-python = ">=3.10"
+authors = [
+    { name = "The Apache Software Foundation", email = "dev@buildstream.apache.org" },
+]
+classifiers = [
+    "Environment :: Console",
+    "Intended Audience :: Developers",
+    "License :: OSI Approved :: Apache Software License",
+    "Operating System :: POSIX",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
+    "Programming Language :: Python :: 3.14",
+    "Topic :: Software Development :: Build Tools",
+]
+
+[project.urls]
+Homepage = "https://buildstream.build"
+Documentation = "https://docs.buildstream.build"
+Source = "https://github.com/apache/buildstream"
+Tracker = "https://github.com/apache/buildstream/issues"
+"Mailing List" = "https://lists.apache.org/list.html?dev@buildstream.apache.org"
+
+[project.scripts]
+bst = "buildstream._frontend:cli"
+
 [tool.black]
 line-length = 119
 exclude = '''
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,8 +13,8 @@
 #
 [build-system]
 requires = [
-    # We need at least version 41.2.0 to support python 3.9
-    "setuptools>=41.2.0",
+    # We need at least version 77 to support PEP 639
+    "setuptools>=77",
     # In order to build wheels, and as required by PEP 517
     "wheel",
     # Require a new enough cython for the python versions we support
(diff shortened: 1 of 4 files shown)


abderrahim:
Looks like this somehow changed the version of `black` used

pyproject.toml
```diff
@@ -23,6 +23,40 @@ requires = [
 ]
 build-backend = "setuptools.build_meta"
 
+[project]
+name = "BuildStream"
+dynamic = ["version", "dependencies"]
+description = "A framework for modelling build pipelines in YAML"
+readme = { file = "README.rst", content-type = "text/x-rst" }
+license = { text = "Apache License Version 2.0" }
```
juergbi: Can we increase our minimum setuptools version to 77.0 and then switch to PEP 639? https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license-and-license-files

```suggestion
license = "Apache-2.0"
```

And then we could drop the deprecated license classifier, fixing this warning, which has a deadline:

```
        Please use a simple string containing a SPDX expression for `project.license`. You can also use `project.license-files`. (Both options available on setuptools>=77.0.0).

        By 2027-Feb-18, you need to update your project and remove deprecated calls
        or your builds will no longer be supported.

        See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.
```

The latest maintenance releases of Python 3.10 and 3.11 already bundle setuptools 79.0 and more recent Python versions don't bundle setuptools at all anymore. Debian 12 (oldstable) and Ubuntu 22.04 and 24.04 still have older setuptools versions, but I don't know whether that would actually be a problem when installing BuildStream via pip.

jjardon: As discussed in matrix, I will send this in other PR (I have applied the license suggestion though)

juergbi:
> Looks like this somehow changed the version of `black` used

It's not a different version of `black`. The behavior of `black` is affected by `requires-python` in `pyproject.toml` (but apparently not by the equivalent in `setup.cfg`, which we had before).

juergbi:
I've opened https://github.com/apache/buildstream/pull/2164 to separate the partial reformatting from this branch. The `black` issue should disappear when rebasing this branch on top of https://github.com/apache/buildstream/pull/2164.

pyproject.toml
```diff
@@ -23,6 +23,40 @@ requires = [
 ]
 build-backend = "setuptools.build_meta"
 
+[project]
+name = "BuildStream"
+dynamic = ["version", "dependencies"]
+description = "A framework for modelling build pipelines in YAML"
+readme = { file = "README.rst", content-type = "text/x-rst" }
```
abderrahim: I think content-type isn't needed in this case.

pyproject.toml
```diff
@@ -23,6 +23,40 @@ requires = [
...
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
+    "Programming Language :: Python :: 3.14",
+    "Topic :: Software Development :: Build Tools",
+]
+
+[project.urls]
+Homepage = "https://buildstream.build"
```
abderrahim, line 50: https://buildstream.apache.org/

pyproject.toml
nathanwilliams-ct: +1

I even have the docs for this still open in a tab: https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#readme from when I looked it up the other day.

setup.cfg
```diff
@@ -11,9 +11,6 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 #
-[options]
-python_requires = >=3.10
```
nathanwilliams-ct, line 15: If you are moving stuff to pyproject.toml.. is it worth moving everything else in the setup.cfg? 

for examples:
- https://docs.pytest.org/en/stable/reference/customize.html#pyproject-toml
- https://github.com/python-versioneer/python-versioneer/blob/master/INSTALL.md#common-steps
- etc

nathanwilliams-ct: (or move the pytest and related config into the tox.ini e.g. https://gitlab.com/BuildStream/buildstream-plugins-community/-/blob/master/tox.ini?ref_type=heads#L108  although I am debating moving the pytest config out of tox.ini over in buildstream-plugins-community, as INI files are deprecated with tox: https://tox.wiki/en/4.60.1/reference/config.html#discovery-and-file-types)

pyproject.toml
jjardon: That is a change in content; prefer to do it as part of other PR

setup.cfg
jjardon: All that can be done after merging this; I wanted to keep things to a minimum for now

pyproject.toml
jjardon: Fixed

tox.ini
```diff
@@ -140,7 +140,7 @@ setenv =
 [testenv:format]
 skip_install = True
 deps =
-    black~=26.3.0
+    black==26.3.0
```
juergbi: I think this change was an earlier attempt to fix the issue with `black`. #2164 has been merged instead, and I think there is no need for this change anymore.

jjardon: @juergbi sure, removed

juergbi (approved):

pyproject.toml
juergbi: You've included the setuptools version bump as otherwise CI failed with the new license field. I've verified that `pip install` from the source directory still works in a venv on Debian 12 despite Debian 12 installing setuptools 66.1.

PR #2166, opened 2026-08-11
_stream: Persist soft workspace resets

kotborealis:
The soft reset path cleared `workspace.last_build` only in memory. Since `workspace_reset()` returned without `Workspaces.save_config()`, the old hash remained in `.bst2/workspaces.yml` and was loaded by the next BuildStream invocation. That retained the previous buildtree and could skip configure-commands.

This regressed in 1eedab4fd, when the final save after the reset loop was removed as hard reset moved to `workspace_close()`/`workspace_open()`; those methods save the config themselves, but the soft path does not. The later `last_successful` -> `last_build` rename preserved the omission.

Save the config once after all soft-reset targets are processed and cover the on-disk state before and after a new CLI invocation.

### DIFF
--- a/src/buildstream/_stream.py
+++ b/src/buildstream/_stream.py
@@ -1174,6 +1174,9 @@ class Stream:
             workspaces.save_config()
             self.workspace_open([element._get_full_name()], no_checkout=False, force=True, custom_dir=workspace_path)
 
+        if soft:
+            workspaces.save_config()
+
     # workspace_exists
     #
     # Check if a workspace exists
--- a/src/buildstream/_stream.py
+++ b/src/buildstream/_stream.py
@@ -1168,6 +1168,7 @@ class Stream:
                 self._context.messenger.info(
                     "Reset workspace state for {} at: {}".format(element.name, workspace_path)
                 )
+                workspaces.save_config()
                 continue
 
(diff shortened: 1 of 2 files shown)


kotborealis:
@abderrahim please take a look at this, it would be great to merge this into buildstream 2.8, as it breaks local workflows a bit.

nathanwilliams-ct:
LGTM

kotborealis:
@nathanwilliams-ct @abderrahim ping, it would be really nice to merge this into 2.8, as this patch allows to work properly with --no-strict and workspaces :)

src/buildstream/_stream.py
```diff
@@ -1174,6 +1174,9 @@ def workspace_reset(self, targets, *, soft):
             workspaces.save_config()
             self.workspace_open([element._get_full_name()], no_checkout=False, force=True, custom_dir=workspace_path)
 
+        if soft:
+            workspaces.save_config()
```
abderrahim: I think it makes more sense (given the current code) to move this inside the loop (under the existing `if soft:`)

abderrahim (approved):
Small comment. Looks good otherwise.

src/buildstream/_stream.py
kotborealis: Done

PR #2167, opened 2026-08-11
Type checking focused around element.py and shell testing

nathanwilliams-ct:
Note: This adds a number of assertions where nullable values were assumed to be populated.

There should not be any functional changes.


src/buildstream/_frontend/app.py
```diff
@@ -87,9 +93,9 @@ def __init__(self, main_options):
         self._detail_profile = Profile(dim=True)
 
         # Cached messages
-        self._cached_message_lock = threading.Lock()
-        self._cached_message_text = ""
-        self._cache_messages = None
+        self._cached_message_lock: Lock = threading.Lock()
```
richardmaw-codethink: Since Lock has been imported directly I'd be tempted to switch over to using it.


```suggestion
        self._cached_message_lock: Lock = Lock()
```

src/buildstream/_frontend/app.py
```diff
@@ -587,9 +596,9 @@ def _render_status(self):
     # Handle ^C SIGINT interruptions in the scheduling main loop
     #
     def _interrupt_handler(self):
-
+        assert self.stream, "Must have a Stream in App for this"
         # Only handle ^C interactively in interactive mode
-        if not self.interactive:
+        if not self.interactive:  #
```
richardmaw-codethink: Leftover unfinished comment, presumably replaced by the assertion.


```suggestion
        if not self.interactive:
```

src/buildstream/_loader/loadcontext.py
```diff
@@ -14,23 +14,31 @@
...
 class ProjectLoaders:
-    def __init__(self, project_name):
+    def __init__(self, project_name: str):
 
         # The project name
         self._name = project_name
 
         # A list of all loaded loaders for this project
-        self._collect = []
+        self._collect: list["Loader"] = []
```
richardmaw-codethink, line 41: https://docs.python.org/3/library/__future__.html#future__.annotations is the typical solution recommended for circular definitions

but honestly it's due to become the default in the next release and the import deprecated to eventually become a syntax error so switching to it is going to cause more work down the line

src/buildstream/_loader/loader.py
```diff
@@ -622,13 +649,13 @@ def _check_circular_deps(top_element):
...
-    def _search_for_local_override(self, override_path):
+    def _search_for_local_override(self, override_path: str) -> Optional[ScalarNode]:
         junction = self.project.junction
         if junction is None:
             return None
 
         # Try the override without any link substitutions first
-        with suppress(KeyError):
+        if override_path in junction.overrides:
             return junction.overrides[override_path]
```
richardmaw-codethink: I have a style preference for walrusing this kind of thing because otherwise you hash the key twice, but it's not universally recommended.


```suggestion
        if (override_node := junction.overrides.get(override_path)) is not None:
            return override_node
```

src/buildstream/_loader/loader.py
```diff
@@ -841,6 +869,10 @@ def provenance_str():
 
         element = Element._new_from_load_element(load_element)
 
+        from ..plugins.elements.junction import JunctionElement
```
richardmaw-codethink, line 872: Is this deferred import intentional? It's not going to have the problem of deferring finding import problems since `Element._new_from_load_element` will have needed to import it, but it does seem out of place.

src/buildstream/_loader/loader.py
```diff
@@ -875,9 +908,10 @@ def provenance_str():
             # we haven't yet for this element),
             # element._get_cache_key() can fail if used with the
             # default _KeyStrength.STRONG.
-            basedir = os.path.join(
-                self.project.directory, ".bst", "staged-junctions", filename, element._get_cache_key(_KeyStrength.WEAK)
-            )
+            assert self.project.directory, "The loaders project must have a project directory"
+            key = element._get_cache_key(_KeyStrength.WEAK)
+            assert key, "We expect a weak key is always available"
```
richardmaw-codethink, line 913: When you get to it, it's possible to overload the definition of _get_cache_key to make the assert unnecessary.

```python3
@overload
def _get_cache_key(self, strength: Literal[_KeyStrength.WEAK]) -> str:
    ...
@overload
def _get_cache_key(self, strength: Literal[_KeyStrength.STRONG) -> Optional[str]:
   ...
def _get_cache_key(self, strength: _KeyStrength = KeyStrength.STRONG) -> Optional[str]:
    if strength == _KeyStrength.STRONG:
        return self.__cache_key
    else:
        return self.__weak_cache_key
```


```suggestion
            key = element._get_cache_key(_KeyStrength.WEAK)
```

src/buildstream/_testing/runcli.py
```diff
@@ -333,7 +372,7 @@ def run(self, project=None, silent=False, env=None, cwd=None, options=None, args
             if project:
                 bst_args += ["--directory", str(project)]
 
-            for option, value in options:
+            for option, value in batched(options, n=2):
```
richardmaw-codethink, line 375: Nice catch!

src/buildstream/_testing/runcli.py
```diff
@@ -394,11 +433,12 @@ def _invoke(self, cli_object, args=None, binary_capture=False):
                 exc_info = sys.exc_info()
             finally:
                 sys.stdout.flush()
+                # sys.stderr.flush()
```
richardmaw-codethink: I think this is accidentally left over.

src/buildstream/element.py
```diff
@@ -1051,25 +1074,26 @@ def _stage_dependency_artifacts(self, sandbox, scope, *, path=None, include=None
...
     @classmethod
-    def _new_from_load_element(cls, load_element, task=None):
+    def _new_from_load_element(cls, load_element: LoadElement, task: Optional[Task] = None) -> Element:
 
         if not load_element.first_pass:
             load_element.project.ensure_fully_loaded()
 
-        with suppress(KeyError):
+        if load_element in cls.__instantiated_elements:
             return cls.__instantiated_elements[load_element]
```
richardmaw-codethink: See previous comment.


```suggestion
        if (instantiated_element := cls.__instantiated_elements.get(load_element)) is not None:
            return instantiated_element
```

src/buildstream/element.py
```diff
@@ -1284,7 +1311,7 @@ def _buildable(self):
     #
     # None is returned if information for the cache key is missing.
     #
-    def _get_cache_key(self, strength=_KeyStrength.STRONG):
+    def _get_cache_key(self, strength=_KeyStrength.STRONG) -> str | None:
```
richardmaw-codethink, line 1324: See previous comment about _get_cache_key

```suggestion
    @overload
    def _get_cache_key(self, strength: Literal[_KeyStrength.WEAK]) -> str:
        ...
    @overload
    def _get_cache_key(self, strength: Literal[_KeyStrength.STRONG) -> Optional[str]:
       ...
    def _get_cache_key(self, strength: _KeyStrength = _KeyStrength.STRONG) -> Optional[str]:
        if strength == _KeyStrength.STRONG:
            return self.__cache_key
        else:
            return self.__weak_cache_key
```

src/buildstream/element.py
```diff
@@ -1779,12 +1815,13 @@ def _cache_artifact(self, sandbox, collect):
             except DirectoryError:
                 pass
 
-        # We should always have cache keys already set when caching an artifact
-        assert self.__cache_key is not None
-        assert self.__artifact._cache_key is not None
+        #
```
richardmaw-codethink: I suspect this comment is left-over.


```suggestion
```

src/buildstream/element.py
```diff
@@ -2310,7 +2346,10 @@ def _fetch(self, fetch_original=False):
     #
     # None is returned if information for the cache key is missing.
     #
-    def _calculate_cache_key(self, dependencies, weak_cache_key=None):
+    def _calculate_cache_key(self, dependencies: list[list[str]], weak_cache_key: Optional[str] = None) -> str | None:
+        assert self.__sandbox_config, "Element should have a sadbox config to calculate cache key"
```
richardmaw-codethink:
```suggestion
        assert self.__sandbox_config, "Element should have a sandbox config to calculate cache key"
```

src/buildstream/element.py
```diff
@@ -2592,7 +2636,7 @@ def _get_logs(self) -> List[str]:
     #    (ElementProxy): An ElementProxy to self, for owner.
     #
     def __get_proxy(self, owner: "Element") -> ElementProxy:
-        with suppress(KeyError):
+        if owner in self.__proxies:
             return self.__proxies[owner]
```
richardmaw-codethink: See previous comment about walrus


```suggestion
        if (cached_proxy := self.__proxies.get(owner)) is not None:
            return cached_proxy
```

richardmaw-codethink (commented):
I was curious and had some comments while reading through, hope you don't mind.

src/buildstream/_testing/runcli.py
nathanwilliams-ct:
```suggestion
```

src/buildstream/_loader/loader.py
nathanwilliams-ct: It's a cyclic import as JunctionElement imports Element already.

I am not a big fan of the cast here, but I am not sure on better way to do it with python's type checking. The cast is mainly to keep the type checker happy, and I am unsure if it's 'free' at runtime and makes the import unnecessary. I did debate wrapping it behind a `if TYPE_CHECKING` gate.

An `assert isinstance(element,JunctionElement)` also didn't play nice, as Element is it's base class.

I was trying to work out how I might get _new_from_load_element to just return a JunctionElement.

src/buildstream/_loader/loader.py
nathanwilliams-ct: With how Element is currently implemented, the WEAK cache key is also Optional within element. Unless we change `_get_cache_key` to assert that a WEAK key is available at minimum, which I don't want to do in this PR.

src/buildstream/_loader/loadcontext.py
nathanwilliams-ct: I went with the Manual String annotations as that is what is used elsewhere in the codebase at the moment.

Rabbit hole of documentation:
- https://docs.python.org/3/library/__future__.html#future__.annotations
- https://docs.python.org/3/reference/compound_stmts.html#annotations
- https://peps.python.org/pep-0749/
- https://docs.python.org/3/library/annotationlib.html#module-annotationlib

Which summarises to:
- For 3.14 annotations will be lazily evaluated.
- `from __future__ import annotations` sticks around until 3.13 reaches end of life.
- `from __future__ import annotations` has been around since 3.7

Buildstream currently supports the following python versions: 310,311,312,313,314

Using `from __future__ import annotations` would mean a simple find and replace could remove it in future. Using manual quotes like this would be a lot of manual re-work..

I think I will switch to `from __future__ import annotations` for this PR to reduce future work...?

src/buildstream/element.py
```diff
@@ -2700,11 +2778,18 @@ def source_provenance_attribute_check(provenance_node=provenance_node):
...
+    def __get_dependency_artifact_names(self) -> list[str]:
+        refs = []
+        for dep in self._dependencies(_Scope.BUILD):
+            key = dep._get_cache_key()
+            assert key, f"Must have cache key for {_get_normal_name(dep.name)}"
+            refs.append(
+                os.path.join(
+                    dep.project_name,
+                    _get_normal_name(dep.name),
+                )
```
nathanwilliams-ct, line 2792:
```suggestion
            key = dep._get_cache_key()
            if key is None:
               key = ""
            refs.append(
                os.path.join(
                    dep.project_name,
                    _get_normal_name(dep.name),
                    key,
                )
```

src/buildstream/_loader/loadcontext.py
abderrahim: We don't want to rely on `from __future__ import annotations` since it's going away, but if the code can be made to work with it for python < 3.14, and with the native lazy annotations support in python 3.14, then it should be fine. We can import it conditionally.

But let's please try to keep changes to a minimum. If you want to do this, please do it separately.

PR #2173, opened 2026-08-12
Element dependencies: Improve inline documentation to make it clear that deps of  _Scope.BUILD don't return self

nathanwilliams-ct:
All other cases yield an element and a subset of it's dependencies recursively with visit...

I suspect the _Scope.BUILD case should do this too.

### DIFF
--- a/src/buildstream/element.py
+++ b/src/buildstream/element.py
@@ -864,7 +864,7 @@ class Element(Plugin):
     # Yields:
     #    (Element): The dependencies in `scope`, in deterministic staging order
     #
-    def _dependencies(self, scope: _Scope, *, recurse=True, visited=None):
+    def _dependencies(self, scope: _Scope, *, recurse=True, visited=None) -> Iterator["Element"]:
 
         # The format of visited is (BitMap(), BitMap()), with the first BitMap
         # containing element that have been visited for the `_Scope.BUILD` case
@@ -883,8 +883,8 @@ class Element(Plugin):
                         yield dep
         else:
 
-            def visit(element, scope, visited):
-                if scope == _Scope.ALL:
+            def visit(element: Element, scope: _Scope, visited: tuple[BitMap, BitMap]) -> Iterator[Element]:
+                if scope == _Scope.ALL:  # The element, its runtime and build dependencies
                     visited[0].add(element._unique_id)
                     visited[1].add(element._unique_id)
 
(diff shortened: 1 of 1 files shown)


abderrahim:
No,  `--deps build` generally implies the element's build dependencies but not the element itself.

juergbi (commented):
No, this is an intentional difference.

https://docs.buildstream.build/master/arch_dependency_model.html

src/buildstream/element.py
```diff
@@ -893,22 +893,22 @@ def visit(element, scope, visited):
...
                     visited[0].add(element._unique_id)
 
                     for dep in element.__build_dependencies:
                         if dep._unique_id not in visited[1]:
                             yield from visit(dep, _Scope.RUN, visited)
-
-                elif scope == _Scope.RUN:
+                    yield element
+                elif scope == _Scope.RUN: # The element and it's runtime dependencies
                     visited[1].add(element._unique_id)
```
nathanwilliams-ct, line 904:
```suggestion
                elif scope == _Scope.BUILD: # The element's build dependencies only
visited[0].add(element._unique_id)

for dep in element.__build_dependencies:
if dep._unique_id not in visited[1]:
yield from visit(dep, _Scope.RUN, visited)
                    # _Scope.BUILD intentionally excludes yield element here.
                elif scope == _Scope.RUN: # The element and it's runtime dependencies
```

nathanwilliams-ct:
ta both.

I think it might still be valuable to add this documentation inline.

src/buildstream/element.py
```diff
@@ -883,8 +883,8 @@ def _dependencies(self, scope: _Scope, *, recurse=True, visited=None):
                         yield dep
         else:
 
-            def visit(element, scope, visited):
-                if scope == _Scope.ALL:
+            def visit(element: Element, scope: _Scope, visited: tuple[BitMap,BitMap]) -> Generator[Element]:
+                if scope == _Scope.ALL: # The element, it's runtime and build dependencies
```
juergbi:
```suggestion
                if scope == _Scope.ALL: # The element, its runtime and build dependencies
```

src/buildstream/element.py
```diff
@@ -893,22 +893,22 @@ def visit(element, scope, visited):
...
+                elif scope == _Scope.BUILD: # The element's build dependencies only
                     visited[0].add(element._unique_id)
 
                     for dep in element.__build_dependencies:
                         if dep._unique_id not in visited[1]:
                             yield from visit(dep, _Scope.RUN, visited)
-
-                elif scope == _Scope.RUN:
+                    # _Scope.BUILD intentionally excludes yielding the element itself.
+                elif scope == _Scope.RUN: # The element and it's runtime dependencies
```
juergbi:
```suggestion
                elif scope == _Scope.RUN: # The element and its runtime dependencies
```

juergbi (approved):
LGTM, just a couple of nits.

nathanwilliams-ct:
CI effected by this: #2177 I just need to rebase.

PR #2174, opened 2026-08-15
_stream: use full element name in `artifact log`/`list-contents`

c1assik:
`artifact_log()` and `artifact_list_contents()` were keying their result dicts by `obj.name`, which is just the bare, project-relative element name. For elements loaded across a junction, this drops the owning junction prefix, so e.g. `bst artifact list-contents element.bst nested.bst:element.bst` would only show one entry, since both targets resolved to the same key.

Use `obj._get_full_name()` instead, which includes the junction prefix, falling back to the artifact ref for `ArtifactElement` targets (i.e. when an artifact is referred to by its ref rather than its element name) as before.

Also drops a dead `obj.name = {ref: "No artifact cached"}` assignment in `artifact_list_contents()` that was immediately followed by `continue` and never used.

Add tests covering `artifact log` and `artifact list-contents` with a cross-junction element alongside a same-named top-level element.

### DIFF
--- a/src/buildstream/_stream.py
+++ b/src/buildstream/_stream.py
@@ -844,7 +844,7 @@ class Stream:
                 self._context.messenger.warn("{} is cached without log files".format(ref))
                 continue
 
-            artifact_logs[obj.name] = obj._get_logs()
+            artifact_logs[obj._get_full_name()] = obj._get_logs()
 
         return artifact_logs
 
@@ -869,15 +869,12 @@ class Stream:
             ref = obj.get_artifact_name()
             if not obj._cached():
                 self._context.messenger.warn("{} is not cached".format(ref))
-                obj.name = {ref: "No artifact cached"}
                 continue
-            if isinstance(obj, ArtifactElement):
-                obj.name = ref
 
             # Just hand over a Directory here
             artifact = obj._get_artifact()
             files = artifact.get_files()
-            elements_to_files[obj.name] = files
+            elements_to_files[obj._get_full_name()] = files
 
         return elements_to_files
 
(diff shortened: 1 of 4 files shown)


src/buildstream/_stream.py
```diff
@@ -844,7 +844,8 @@ def artifact_log(self, targets):
                 self._context.messenger.warn("{} is cached without log files".format(ref))
                 continue
 
-            artifact_logs[obj.name] = obj._get_logs()
+            name = ref if isinstance(obj, ArtifactElement) else obj._get_full_name()
```
nathanwilliams-ct: Perhaps it would be better if `ArtifactElement` overrode the `_get_full_name` method or the `__full_name` variable, instead of having this awkward condition here?

```suggestion
            name = obj._get_full_name()
```

In `ArtifactElement`
```python
def _get_full_name(self)
   return self.get_artifact_name()
```

c1assik: Added a method _get_full_name() to ArtifactElement.

 return self.get_artifact_name(), unconditionally  fails with` AssertionError: assert key is not None` in get_artifact_name() so if the cache key isn't set yet, fall back to super()._get_full_name() (the base Plugin/Element name, which is always safe since it's precomputed earlier in Plugin.__init__). Once construction finishes and the key is set, it returns the full artifact ref as intended

nathanwilliams-ct:
LGTM

src/buildstream/_stream.py
abderrahim: I think it makes more sense to set the `name` in ArtifactElement (you can set it before chaining up to the super class `__init__()` so that it gets picked up). I feel it would be more elegant than trying to override `_get_full_name()`.

PR #2175, opened 2026-08-17
 _frontend/cli.py: Update default min-version for `bst init`

juergbi:
This is implicitly required by tests after creating the 2.8.0.dev0 tag.

### DIFF
--- a/src/buildstream/_frontend/cli.py
+++ b/src/buildstream/_frontend/cli.py
@@ -420,7 +420,7 @@ def help_command(ctx, command):
 @click.option(
     "--min-version",
     type=click.STRING,
-    default="2.7",
+    default="2.8",
     show_default=True,
     help="The required format version",
 )


nathanwilliams-ct:
LGTM

src/buildstream/_frontend/cli.py
```diff
@@ -420,7 +420,7 @@ def help_command(ctx, command):
 @click.option(
     "--min-version",
     type=click.STRING,
-    default="2.7",
```
nathanwilliams-ct, line 423: In the test, is it worth adding a memo, along the lines of:

`assert project_conf.get_str("min-version") == get_default_min_version(), 'project min version does not equal default min version. If this is failing and repository was recently tagged, a version update PR is required for this test to pass again.'`

abderrahim (approved):

src/buildstream/_frontend/cli.py
juergbi: Yes, or maybe we can find a way to streamline the process. But let's handle this outside of this minimal update, after the final 2.8.0 release.

nathanwilliams-ct: agreed, getting the CI working again is priority.

nathanwilliams-ct: https://github.com/apache/buildstream/issues/2177 created this as a follow-up.

PR #2185, opened 2026-08-25
Migrate from ujson to orjson

### LINKED ISSUE #2184: Find replacement for ujson (ujson is in maintenance-only mode)
See https://github.com/ultrajson/ultrajson#project-status

ujson authors recommend migrate to https://pypi.org/project/orjson/

jjardon:
ujson is in maintenance-only mode and its own maintainers recommend migrating to orjson. This change replaces the dependency and updates the single call site in _cachekey.py.

Key differences handled:
- orjson.dumps() returns bytes (so the extra .encode("utf-8") is removed)
- sort_keys=True becomes option=orjson.OPT_SORT_KEYS
- escape_forward_slashes=False is the default behaviour of orjson

Tested successfully at https://github.com/jjardon/buildstream-1/actions/runs/32796227859?pr=4

Fixes #2184

### DIFF
--- a/src/buildstream/node.pyx
+++ b/src/buildstream/node.pyx
@@ -81,7 +81,7 @@ cdef class Node:
         self.column = column
 
     # This is in order to ensure we never add a `Node` to a cache key
-    # as ujson will try to convert objects if they have a `__json__`
+    # as orjson will try to convert objects if they have a `__json__`
     # attribute.
     def __json__(self):
         raise ValueError("Nodes should not be allowed when jsonify-ing data", self)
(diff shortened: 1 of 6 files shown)


src/buildstream/node.pyx
```diff
@@ -81,7 +81,7 @@ cdef class Node:
         self.column = column
 
     # This is in order to ensure we never add a `Node` to a cache key
-    # as ujson will try to convert objects if they have a `__json__`
+    # as orjson will try to convert objects if they have a `__json__`
```
abderrahim, line 84: I can't see anything to this effect in orjson repository.

PR #2187, opened 2026-09-02
_sourcetests: Do not assume Source.ref is a string

aleixpol:
types.py specifies as:
`SourceRef = Union[None, int, str, List[Any], Dict[str, Any]]`


abderrahim (approved):
Please add a _sourcetests: prefix to your commit message. It looks weird without it.

aleixpol:
> Please add a _sourcetests: prefix to your commit message. It looks weird without it.

Added

